PR #11, opened 2018-11-25
Fixed Search bug

sinomsinom:
When searching the page count would be determined by a case insensitive search and the Quests displayed would be determined from a case sensitive search. I made both case insensitive and changed the search from "beginsWith" to "contains"


HeyZeer0:
I literally fixed this the time you pushed this pull request, lmfao xD
Changing the search to .contains will include wrong quests to the search array so I'll keep it as startsWith

Thanks for your contribution, I'm waiting more from you =)

PR #13, opened 2018-12-01
Fixed small things

sinomsinom:
Basically tried to remove as many eclipse-warnings as possible without breaking anything. 
Also changed some inconsistencies and typos i found


HeyZeer0:
Denied, there is no reason to generify things like that, since doesn't make a lot of difference cuz the compiler by itself ignores it, you also changed things that could break the Music System and make the file replacer slower.
And yes, the music player runs as a loop, and some times the #stop method doesn't works so that's why there is a precaution stopper.

PR #15, opened 2018-12-11
Prevent closing loot chests that contain mythic items

Bedo9041:
Not actually tested for mythics (for obvious reasons) - but worked when I tested it using rares/uniques. Couldn't find any events in forge to handle closing an inventory, so this looked like the best way of doing it.


HeyZeer0:
So I made some quick changes into it, thanks for helping us developing this project!

PR #19, opened 2018-12-14
Implement Quest Giver Icons

Bedo9041:
At the moment, this is by no means a perfect implementation. It's issues include:

- Only working when a NPC has the grey 'NPC' label
- Detecting quest-relevant NPCs when their name appears in the current step description of a quest, meaning:
  - NPCs are only given icons if their name appears in the quest text, certain NPCs could therefore be ignored (e.g. in the quest "King's Recruit", the final step says: "Talk to the King.", however the NPC is called 'Ragni's King', which means no icon is drawn)
  - NPCs could be given icons when you don't actually need to talk to them (e.g. in the quest "Enzan's Brother", the final step says: "Find Enzan's brother", because the name 'Enzan' is in this text, the NPC 'Enzan' has a question mark over their head even though you don't need to talk to them again in that quest)
- I've reused the nameplate drawing code instead of drawing an image, this means that the quest NPCs don't look as good as they could:
  - [Quest can be started](https://i.imgur.com/zQTcvBA.png)
  - [Quest in progress](https://i.imgur.com/KyoRr9k.png)
- There's some weirdness when moving the camera angle around, this doesn't ruin it, but probably could be improved:
  - https://i.imgur.com/kMV5qem.png
  - https://i.imgur.com/ipZ9rK4.png
I've recorded an example of me going through the first 3 quests: King's Recruit, Enzan's brother and Poisoning the Pest: https://youtu.be/fwT9VW7Y_5o

### DIFF
--- a/src/main/java/cf/wynntils/modules/questbook/configs/QuestBookConfig.java
+++ b/src/main/java/cf/wynntils/modules/questbook/configs/QuestBookConfig.java
@@ -19,4 +19,7 @@ public class QuestBookConfig extends SettingsClass {
     @Setting(displayName = "Set quest location to compass", description = "Should the compass follow the quest coordinates")
     public boolean compassFollowQuests = true;
 
+    @Setting(displayName = "Quest NPC icons", description = "Should quest NPCs have icons above their heads")
+    public boolean questGiverIcons = true;
+
 }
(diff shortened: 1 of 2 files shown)


HeyZeer0:
Seems to be fantastic, I'll not merge this since it's not on it best performance, and I'm on vacation so I can't improve it, I kinda have some ideas on how we can detect the NPCs on a better way, I'll just finish my idea and I'll let you know if you want.

Bedo9041:
Sounds good - I'm almost certain there are a million and one ways to do it better, but with my limited knowledge this was the best I could come up with.

HeyZeer0:
[Nameplate for already identified items](https://trello.com/c/KmuEeY6C/89-nameplate-for-already-identified-items)

PR #7, opened 2018-10-28
Fixed grammar errors.

talkingkittycat:
Fixed grammar errors in the quest book GUI.

PR #27, opened 2018-12-28
Fixed typos and clarified vagueness

CringeAura:


CringeAura:
Can someone explain how a **two-word change in a display text** causes a build to fail? (referring to commit #7)

PR #41, opened 2019-01-12
QoL - Party join

HighCrit:
I've made it so that you can now click the /party join ... command to join the party.


HighCrit:
I've made two version of this, one which can only do party invites and the other that does both. They might not be optimized very well. I'm a beginner and this is the first opensource project I contributed to.

PR #43, opened 2019-01-14
New ticker message: Inventory full

HighCrit:
I've implemented a new ticker message that will tell the player when his/her inventory is full.


Bedo9041 (approved):
Inventory full messages work well - there was a problem with the messages appearing 2x more often than they should - but that was my fault for incorrectly implementing the tick event (already fixed on master).
Chat tab ordering works well - existing installations default all tab's order number to 0, but that doesn't appear to cause any issues.
In the future, I advise using different git branches, which will allow you to split changes into different PRs more easily.

HighCrit:
Yeah, I'll use different branches next time around. And, for default to 0 - I probably should've told that, as during testing I couldn't find an instance where the old settings (and order of them) were altered.

PR #45, opened 2019-01-17
Fixed an exception due to the possibility to rename items

jpresent:


CringeAura:
[Exception caused by ability to rename items on 1.18 beta](https://trello.com/c/bRN0po9J/112-exception-caused-by-ability-to-rename-items-on-118-beta)

PR #46, opened 2019-01-18
Added tier colours for ingredients

HighCrit:
https://imgur.com/a/zGEXeCg


HighCrit:
Don't merge yet! There seems to be a bug with dropped items.

HeyZeer0:
As you wish ^-^

HighCrit:
[BUG] The highlights stay when quickly dropping the items from one's inventory using Q. Does not occur in chest or wiith drag & drop. Only affects ingredients.

PR #50, opened 2019-01-23
Bug fix - Opening questbook with discoveries completed

HighCrit:
Fixes a bug reported by Saynkas, https://pastebin.com/vDL63Qr8, when Completed is given as a minimum level for a discovery


HighCrit:
It'll likely not fix the problem only prevent crashes

PR #51, opened 2019-01-23
Added Party overlay & healthbars

HighCrit:
[What]
Party overlay and party member health bar (above character and in overlay) [imgur](https://imgur.com/a/eBO3EY2)
[How]
It searches the TabGui for yellow names (since the API breaks when joining parties), creates a list of those names in PlayerInfo. Then it tries to call the EntityPlayer by name - if successful it will map the player name to it's current health. For the overlay to display.
[Performance impact]
Unknown - though I didn't notice any with three instances of Minecraft open.
[Known issues]
Players with names longer than 14 characters won't be seen by this method.
[Possible fixes]
Keep bothering Colin to fix API.


DevScyu:
For Lazy People, Here is the imgur image.

HeyZeer0:
This thing is super old and not optimized at all, probably the best option is to close this xD

PR #64, opened 2019-01-30
Added player health bars on war

HighCrit:
I've tested the health bar outside of war and I've tested the if statement during the war. Though I'm fairly confident it will work, I'm unable to test it in the actual situation.


HighCrit:
It has been confirmed that it works

PR #82, opened 2019-03-07
Fixed Alchemy being unselectable in profession filter

InSimpleTermsJordan:
If I understand this correctly, those "16"s that were in there before were likely put there because there are 8 professions, times two for the profession name and the corresponding unicode character. Problem is, the professionArray actually contains 18 items, because there's the "no filter" setting at the beginning, So (I believe) the "% 16" was causing it to loop back to "no profession" when left-clicking on the filter button while the Woodworking profession is selected.

That or I could be completely wrong. Guess I'll see tomorrow.


DevScyu:
Will test later tonight unless someone else can test first.

HighCrit:
Everything was fine, except the going back part - but that was easily fixed.

PR #89, opened 2019-04-11
Updated french

DevChromium:


src/main/resources/assets/wynntils/lang/fr_fr.lang
```diff
@@ -462,12 +462,6 @@ wynntils.config.other.enum.text_shadow.none=Aucun
 wynntils.config.other.enum.text_shadow.normal=Normaux
 wynntils.config.other.enum.text_shadow.outline=Contour
 
-## OTHER CONFIG ##
```
HighCrit, line 465: What happened here?

DevChromium: no idea

PR #101, opened 2019-06-28
Map updates

MitalAshok:
Added a new setting called "Minimap Coordinates" in MapConfig. If true, displays the current coordinates under the map. Defaults to false.

The minimap zoom in / out keys will now also zoom the main map when it is active (4x faster than default scroll zoom)

Added coordinates of where mouse is hovering in main map (Like map.wynncraft.com has in the corner)

Middle clicking on the main map will create a compass beacon at the location middle clicked

Fixed a bug where the compass beacon icon didn't refresh until map was closed.

Also changed default zoom range (Can zoom in and out further, controlled by [`com.wynntils.modules.map.configs.MapConfig.WorldMap.minZoom` / `maxZoom`](https://github.com/Wynntils/Wynntils/pull/101/files#diff-d594ef30017a8115a28bafd940850741R72))

---

I also added another constructor for WaypointCreationMenu so that this can be added somewhere to `com.wynntils.modules.map.overlays.ui.WorldMapUI`:

```java
// Prompt to create a waypoint where the mouse currently is
MapProfile map = MapModule.getModule().getMainMap();
int worldX = getMouseWorldX(mouseX, map);
int worldZ = getMouseWorldZ(mouseY, map);

Minecraft.getMinecraft().displayGuiScreen(new WaypointCreationMenu(null, worldX, worldZ));
```

But I couldn't think of a button combination to make a waypoint.

### DIFF
--- a/src/main/java/com/wynntils/modules/map/configs/MapConfig.java
+++ b/src/main/java/com/wynntils/modules/map/configs/MapConfig.java
@@ -37,7 +37,10 @@ public class MapConfig extends SettingsClass {
     @Setting.Limitations.IntLimit(min = 75, max = 200)
     public int mapSize = 100;
 
-    @Setting(displayName = "Display Only North", description = "Should only north be displayed on the minimap?", order = 6)
+    @Setting(displayName = "Minimap Coordinates", description = "Should your coordinates be displayed below the minimap?", order = 6)
+    public boolean showCoords = false;
+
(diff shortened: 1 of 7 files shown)


HeyZeer0:
Probably the only thing that'll change is the position and the design!

Also let me know your Discord IGN so I can give you the contributor role!
Amazing work, thank you =)

PR #102, opened 2019-06-29
Fixed chat spam filter

MitalAshok:
Instead of making the text box longer, this will put the repeat count on a new line if it is too long (reverts c37c073c1f6d0acacba3dae079f6f6e17bff2576)

Also if the repeated message had disappeared because of time, it will reappear as if it was posted again.


HeyZeer0:
The problem with this PR is that it actually haves a inefficient way of doing that, the smartest way of doing it would be to generate a random id for every chat line, and keep the splitted ones with the same ID, recalculating the line amount after updating it. Since Wynntils already rewrites 80% of the chat interface, I don't think that would be hard to do xD

MitalAshok:
@HeyZeer0 How does this look? A subclass for messages that has a `groupId` for every line grouping them together, and just remove all of the last groupId and replace it with a new message with a new counter at the bottom, and if a line is added from somewhere else, it isn't that subclass so is ignored. (And just incrementing a static groupId instead of a random id, if that's ok?)

PR #103, opened 2019-07-02
Added minimap icons

MitalAshok:
Also an option to turn it off and change the size in the settings, and the compass is replaced with a red arrow if it is too far.


HeyZeer0:
I'm just wondering how long in ns it actually takes per tick to calculate all of the math, could you check that? @MitalAshok

MitalAshok:
Hi @HeyZeer0 I did some timings (At [MitalAshok/Wynntils:profiled_minimap_icons](https://github.com/MitalAshok/Wynntils/compare/minimap_icons...MitalAshok:profiled_minimap_icons?expand=1) if you want to run them yourself, I could only run it on my one pc) and I got:

```
Minimap icon render info:
Ran 591 times in 10.02 s (59.01 Hz)
Min:    25512 ns
Max:    105570 ns
Q1:     27565 ns
Q3:     32843 ns
Median: 28739 ns
Mean:   32702 ns
Std. Dev: +/- 9498 ns
```

Most of the variance would come from rendering, I would think, which is why a few take 10 times longer. I've not worked with mods too much, so I'm not sure.

For the maths that actually happens, it is only run on a few icons because most of them are filtered out [here](https://github.com/Wynntils/Wynntils/pull/103/files#diff-d149eb9aab6e830013f3a448a8f9c6baR137) after doing a few (usually only 1) float comparison.

PR #109, opened 2019-07-03
Waypoint colours and disable waypoints

MitalAshok:
Added a colour selector in the waypoint creation menu. Currently just calls `GlStateManager.color` and renders the icon as it normally would, and the actual selector itself is not the prettiest (Just 4 rgba sliders).

Also fixed a bug where the enabledMapIcons option wasn't honoured (I just forgot to do it in the minimap icon commit). And I also refactored a bunch of code (`MapIconInfo` -> `MapIcon`, which is now an abstract class, and replaced `drawRectF` drawing the icon with `MapIcon.renderAt`)


HeyZeer0:
If you take a look at UI elements, you'll see that there's a UIEColorWheel, that's what the implementation should actually use. But the rest seems fine xD

Also you can use CustomColor#applyColor to apply the color

PR #120, opened 2019-07-11
Added quick cast gui

MitalAshok:
Used by pressing the keybind, holding left click, moving to the spell, and releasing the mouse. This also doubles as the only in-game place where the damage of spells is listed (Taken from https://wynncraft.com/help).

image

Current issues:
* Can't move while in menu (Because it's a GUIScreen, but if it was just a GUI, you couldn't see the mouse cursor)
* Text is unreadably small if window is smaller than 720x720.
* Not explained properly in the keybind name

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
+++ b/src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
@@ -18,6 +18,7 @@ import com.wynntils.modules.utilities.managers.DailyReminderManager;
 import com.wynntils.modules.utilities.managers.KeyManager;
 import com.wynntils.modules.utilities.managers.MountHorseManager;
 import com.wynntils.modules.utilities.managers.NametagManager;
+import com.wynntils.modules.utilities.overlays.ui.GuiQuickCast;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.audio.PositionedSoundRecord;
 import net.minecraft.entity.Entity;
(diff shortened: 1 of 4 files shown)


CringeAura:
Based on what I can see here, I believe this would be ineffective for users to quickly cast spells. If someone was fighting Cybel in LI and trying to cast a spell through this GUI, they would probably die or take severe damage because of the amount of time casting spells through this GUI would take.

MitalAshok:
Hi I [made a demo](https://i.imgur.com/yIWLqYZ.mp4) of the time taken. It's slower than the existing keybinds for each spell, but it's much faster than clicking 3 times (Because it's only 1 click)

CringeAura:
> 
> 
> Hi I [made a demo](https://i.imgur.com/yIWLqYZ.mp4) of the time taken. It's slower than the existing keybinds for each spell, but it's much faster than clicking 3 times (Because it's only 1 click)

It seems like this would be beneficial to users who are familiar with this GUI. Otherwise, the current casting systems in place would better suit other users.

HeyZeer0:
Hey! now you can use GuiMovementScreen to allow the user to move while inside a GuiScreen, just be sure to call the default constructor =P

HeyZeer0 (changes requested):
You can remove this onClick event and add it to the mouseClick, just be sure to call super.mouseClick() at the end of handleMouseInput()

Also, replace GuiScreen with GuiMovementScreen, that will allow the user to move inside the GUI. By the way be sure to remove the handleInput() or do a super at it too, otherwise it'll conflict with GuiMovementScreen

A good idea would also be able to see how much damage you would deal with that spell, that would be a bit complicated since you would have to analyze the compass, if you are going to, store the contents that you find at PlayerInfo, you can use a sub-instance if you want. You can open the GUI in the background using a class named FakeInventory, just check how I do that for handling the QuestBook (Check QuestManager) **OR** maybe you could calculate the stats from raw using the player inventory, would actually be faster I guess

PR #126, opened 2019-08-01
Questbook changes (closes #105)

### LINKED ISSUE #105: The QuestBook GUI is a mess
The QuestBook GUI is actually a mess and needs to be rewritten for more modularity

HighCrit:
Dissected the QuestbookGUI into separate QuestBookPages which can be registered using the new QuestbookHandler. I did my best to create a base class which can be used to create future pages.

-All pages with an icon are now accessible on the main page (See MainPage).
-There's is a reload button available at the the QuestsPage to reload all quest info.
-Removed some code wasn't being used at all

Did some testing on both a low and high level account and didn't experience any crashes or bugs. (Note there will be bugs)

closes #105


HighCrit:
closes #105

HeyZeer0 (changes requested):
Fantastic job, I love to see your improvement over all of this time :)
I would just require you to replace the page registration system with an enum, which would make things a bit faster and easier, like `QuestBookHandler.openPage(true, QuestBookPage.QUESTS)`, you can store the instances inside avoiding the sorting things.
There are some redundancies in some methods names, for example at QuestBookHandler you have a method named openQuestBookPage, you don't need the QuestBook part cuz it's already mentioned on the class name.

Good Job 👍 ❤️

HeyZeer0 (approved):
Approved

PR #131, opened 2019-08-07
Added path waypoints

MitalAshok:
A path waypoint lets you draw on the map. They are a list of connected points.

Also split up `WorldMapUI` so that UIs that use the world map can be made, and decoupled `MapIcon` from textures (to allow the new icon type, `MapPathWaypointIcon`).


HeyZeer0:
@MitalAshok Could you provide some images or gifs? ^^

MitalAshok:
Sure @HeyZeer0!

PathWaypointCreationUI

Minimap with path

It's literally just "click to add a point", and then it renders (a black for outline then coloured) line between them.

Also I hadn't thought about this but the path goes over all the other icons. Should it go under?

HighCrit:
Under would've been preferable imo

PR #137, opened 2019-08-14
Added ability to keybind the auto update keys

douira:
The keys were previously hardcoded to KEY_Y and KEY_N which is problematic for people with non-US keyboard layouts where oftentimes the Y key is actually Z. Now there are two new keybindings that allow the customization of the keys. The dialog box with the update message updates according to the chosen keys.


douira:
It has been noted that adding an extra keybind for this is not a great idea since this is rarely used. A better solution should be found for this. Especially now that wynntils has a separate update screen when connecting by using the icon button, this feature is even less useful.

PR #148, opened 2019-11-11
Add server restart timer

BurstingF:
This should add an optional restart timer based on a suggestion I saw in the discord suggestions channel.


HeyZeer0:
I think we can find a better place to put this on instead of creating a new overlay. Btw you don't need to add an option to disable it since you can simply hide the entire overlay ^^

BurstingF:
I removed the config option to disable it based on what you said. 
The implementation of this restart timer is based on the already existing stopwatch overlay. I guess I could try to group all those timers in one overlay instead, however, that would make you lose the ability to have different timers in different locations.

BurstingF:
@HeyZeer0 is this perhaps more in the direction you were looking for?

BurstingF:
Closing due to no responses.

PR #149, opened 2019-11-14
Added congratulate message

BurstingF:
This adds the option to enable having an additional message in chat when someone reaches level 103 combat for example. Clicking this additional message will send a message/whisper to the person who leveled up based on the format that was configured in the config. Appearance is based on the already existing thanks message you see when someone throws a bomb.
This feature probably isn't that useful but was more just me trying out some stuff.


HeyZeer0:
Hey, @BurstingF this looks cool but there are some issues with it.

1. The message is customizable, that allows people to put swears or similars on it
2. The default message is kinda horrible xD

I hope you find a solution for these issues ^^

BurstingF:
For the default message I just wanted to show all 3 parameters that can be used, but I agree that the current one is kinda bad xD. 
As for the swearing in the message, is that really something we should take care of? Since it's just a regular whisper people can't use words that they otherwise wouldn't be able to use if they just typed the whisper manually. If it is indeed something that you want the mod to take care of then how would you want to see it implemented? A predetermined list of blacklisted regex that checks whether there are any swears or similar in it?

HighCrit:
The thing with blacklists is - people who are willing to bypass it always will. And, it's possible for people to not swear but still insult people (e.g. "Tryhard" or "Find a life").

EDIT:
Also, having a predetermined message will water down the value of a congratulation.

BurstingF:
I modified it a bit which means that by default it will now show this additional message and when you click on the additional message your chat window will now contain "/msg %user% " (with %user% being replaced for the user who leveled up).
You are still able to configure your own predetermined message which will still be send just like before if you were to click on the additional message.
However, since there is now no longer a default message I think this should fix the issue of the value of a congratulation.

BurstingF:
Closing due to no responses.

PR #151, opened 2019-11-15
Improve wynncraft team nametags

### LINKED ISSUE #144: People with [Art] rank do not have the Wynncraft CT tag

BurstingF:
This makes the nametags for the wynncraft teams (such as Moderator, Admin, GM) based on the name of the scoreboard team the player has assigned by wynncraft. 

For Moderators and Admins I don't think anything changes in appearance. People that originally had "Wynncraft CT" will now get their specific rank such as GM or Builder. 

This might also fix [Issue#144](https://github.com/Wynntils/Wynntils/issues/144) but my guess is that [Issue#144](https://github.com/Wynntils/Wynntils/issues/144) is happening because the specific player with the art rank doesn't have a scoreboard team or the art rank in general doesn't have a scoreboard team.

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/managers/NametagManager.java
+++ b/src/main/java/com/wynntils/modules/utilities/managers/NametagManager.java
@@ -67,9 +67,10 @@ public class NametagManager {
             if(PlayerInfo.getPlayerInfo().getFriendList().contains(entity.getName())) customLabels.add(friendLabel); //friend
             else if(PlayerInfo.getPlayerInfo().getGuildList().contains(entity.getName())) customLabels.add(guildLabel); //guild
 
-            if(entity.getDisplayName().getUnformattedText().startsWith(TextFormatting.GOLD.toString())) customLabels.add(moderatorLabel); //moderator
-            if(entity.getDisplayName().getUnformattedText().startsWith(TextFormatting.DARK_RED.toString())) customLabels.add(adminLabel); //admin
-            if(entity.getDisplayName().getUnformattedText().startsWith(TextFormatting.DARK_AQUA.toString())) customLabels.add(wynnContentTeamLabel); //Wynn CT
+            //wynncraft teams (Admin, Moderator, GM, Builder, etc.)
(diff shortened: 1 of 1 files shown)


HeyZeer0 (changes requested):
as safe as it sounds, this actually is called every single rendering tick and creating a new object every time that there's a staff member could cause a massive memory leak. That's the only issue ^^

BurstingF:
That indeed sounds like an issue I didn't think of. 
I think this new commit should fix the issue you described.

src/main/java/com/wynntils/modules/utilities/managers/NametagManager.java
```diff
@@ -69,7 +71,7 @@ public static boolean checkForNametags(RenderLivingEvent.Specials.Pre e) {
 
             //wynncraft teams (Admin, Moderator, GM, Builder, etc.)
             if (entity.getTeam() != null && !(entity.getTeam().getName().equalsIgnoreCase("all") || entity.getTeam().getName().equalsIgnoreCase("afk"))) {
```
HeyZeer0, line 73: ``entity.getTeam().getName().equalsIgnoreCase("afk")`` is this if cause right? Seems like
``entity != null && (!name.equals("all") || name.equals("afk"))``
So you're considerating just afk as a wynncraft team? 

It should probably be
``if (entity.getTeam() != null && !entity.getTeam().getName().equalsIgnoreCase("all") && !entity.getTeam().getName().equalsIgnoreCase("afk")) {``

HeyZeer0 (changes requested):

src/main/java/com/wynntils/modules/utilities/managers/NametagManager.java
BurstingF: `entity.getTeam() != null && (!name.equals("all") || name.equals("afk"))` is not actually how it is written down at the moment. 

Currently it is written down as `entity.getTeam() != null && !(name.equals("all") || name.equals("afk"))` which is equivalent to `entity.getTeam() != null && !name.equals("all") && !name.equals("afk")`. 

I can still change it if you think !A && !B is more readable than !(A || B).

PR #152, opened 2019-11-16
implement a "/s" command to interact with the onlinePlayers API

theniels17:
The command has two main functions, 
1. /s list
this will list the available servers, by default sorted from oldest to newest (largest to smallest uptime). 
server colors will be based on the amount of online players, it follows the same colors as in the compass menu
it accepts options to
    - group the results by server type e.g. lobby servers and wc game servers
    - filter by server type
    - sort by alphabet
    - count the amount of online servers
2. /s info <serverID>
this will give information about the server, currently it is only able to show the player list and the player count. 
Online time would be nice but I can only get the relative online time from the api. Real online time can only be determined by constantly polling (by a third party, not the client)

The main reason to add this is to be able to quickly poll the newest server when you get the notice that your server is going to restart. It is not possible to derive this from the compass menu, and the compass menu is also not available from the game servers.

On the info command I did not implement the feature to poll the rank of each member, this because a full server will trigger 50 api calls for the ranks. This will get you quickly to the API call limit.

the command also supports short versions of the keywords to more easily type the command. This is supported on the [i]nfo [l]ist commands and also all options from the list command (first letter of the word)

### DIFF
--- a/src/main/java/com/wynntils/core/framework/FrameworkManager.java
+++ b/src/main/java/com/wynntils/core/framework/FrameworkManager.java
@@ -126,6 +126,7 @@ public class FrameworkManager {
         ClientCommandHandler.instance.registerCommand(new CommandCompass());
         ClientCommandHandler.instance.registerCommand(new CommandTerritory());
         ClientCommandHandler.instance.registerCommand(new CommandExportDiscoveries());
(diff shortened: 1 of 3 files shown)


HeyZeer0:
Hey @theniels17, if you are in our Discord lemme know your name so I can give you the contributor tag ^^

PR #153, opened 2019-11-16
Update ChatGUI.java

ICT-000:
removed unnecessary code - INTmAker

### DIFF
--- a/src/main/java/com/wynntils/modules/chat/overlays/gui/ChatGUI.java
+++ b/src/main/java/com/wynntils/modules/chat/overlays/gui/ChatGUI.java
@@ -15,10 +15,6 @@ import java.io.IOException;
 
 public class ChatGUI extends GuiChat {
 
-    public ChatGUI() {
-
-    }
-
     public ChatGUI(String defaultInputText) {
         super(defaultInputText);
     }


HighCrit:
It wasn't unnecessary

Bedo9041:
Heya, in Java (and most other OOP languages for that matter) when you create an instance of a subclass, it must also be an instance of the superclass. Java ensures this by calling the superclass' constructor when an instance of the subclass is created.

GuiChat contains 2 constructors: 

```java
public GuiChat() {}
```

and

```java
public GuiChat(String) {...}
```

When we extend the class we have to add each constructor that is also going to be applicable to the subclass. If we don't define one of them we can't use it, therefore by getting rid of the argument-less constructor we can't use it and therefore, if we have used it somewhere, the code won't compile.

*It's worth noting java does add a default argument-less constructor to a class when no other constructor is defined, but because we do have another one, so that doesn't apply here*

Uses of the argumentless constructor: https://github.com/Wynntils/Wynntils/blob/fd983d918c025ac544974c6bbb504dbb99fcb6aa/src/main/java/com/wynntils/modules/chat/overlays/gui/TabGUI.java#L144 https://github.com/Wynntils/Wynntils/blob/fd983d918c025ac544974c6bbb504dbb99fcb6aa/src/main/java/com/wynntils/modules/chat/overlays/gui/TabGUI.java#L151 https://github.com/Wynntils/Wynntils/blob/fd983d918c025ac544974c6bbb504dbb99fcb6aa/src/main/java/com/wynntils/modules/chat/overlays/gui/TabGUI.java#L157

Uses of the other constructor: https://github.com/Wynntils/Wynntils/blob/fd983d918c025ac544974c6bbb504dbb99fcb6aa/src/main/java/com/wynntils/modules/chat/events/ClientEvents.java#L35

Also in future before creating a PR, could you test your code by running it, use the gradle task `buildDependents` to get a jar file in `<projectroot>/build/libs` or setup a run configuration in your IDE with the `--username` and `--password` commandline arguments set.

PR #155, opened 2019-11-17
Added gathering icons as waypoint icons

BurstingF:
This adds a new waypoint icon for each gathering skill.


BurstingF:
fixed the size of the image being doubled

PR #160, opened 2019-12-06
Reduced amount of slots in custom player info

BurstingF:
Removed the last row of the custom player info texture as well as adjusted the corresponding x, y values. This row could be removed because it wouldn't contain any values anyway, since we effectively had 21 slots for each column instead of the original 20 slots wynncraft uses.


HeyZeer0:
Can you take a screenshot of the final result? ^^ @BurstingF

BurstingF:
What exactly do you want a screenshot of? GitHub already provides a nice rich difference between changes in images like this:
firefox_pHVgku6x0f
If it's an in-game screenshot you want then you'll have to wait for a bit till the person who helped me test this comes back.

HeyZeer0:
Mhm, it needs to be in game, so I can verify the alignments =P

theniels17:
This is an ingame screenshot from a server with 20+ players

PR #161, opened 2019-12-08
Added configurable system time based animation duration for player info

BurstingF:
This should make the opening/closing animation of the custom player info based on system time. 
I also added an option to configure the duration of the animation, where a duration of 0 means that it is instant. The default value of 125 ms should look similar to what the animation looks like right now with a good fps.


BurstingF:
Any reason this has no response yet or hasn't been merged yet? If so please tell me so this pr can be closed.

HeyZeer0:
Hey @BurstingF sorry for the late response!
Could you provide a gif or video of the animation?

Thanks!

theniels17:
Here is a video of the animation, one time at 0ms (min), one time at 125ms (default) and one at 500ms (max)
I cannot upload a video to github thus it is a webm inside a zip

[time-based-animation.zip](https://github.com/Wynntils/Wynntils/files/4055315/time-based-animation.zip)

HeyZeer0:
It seems that it actually removed the smoothness aka motionblur of the animation

BurstingF:
The motion blur or smoothness is still present I believe.
125ms
It's true that it's not really noticeable in the video. Perhaps this is because I set the default opening time too high (125ms). For example, I currently get ~100 fps and from testing I know that with that kind of fps RenderGameOverlayEvent.Post is called about a 1000 times on average each second. This means that with the old system the player info hud would open in ~50ms for me. When putting the new system on 50ms I get this kind of image.
50ms

magicus (approved):
Looks good to me.

PR #162, opened 2019-12-08
Prevent hotkey with press enabled from firing multiple times on keydown

BurstingF:
Prevented use of the KeyBinding.isPressed() function which isn't meant for continuous querying according to function description. This should fix the bug where a hotkey action in press mode is fired multiple times during one key press.

### DIFF
--- a/src/main/java/com/wynntils/core/framework/instances/KeyHolder.java
+++ b/src/main/java/com/wynntils/core/framework/instances/KeyHolder.java
@@ -12,6 +12,7 @@ public class KeyHolder {
     private Runnable onAction;
     private boolean press;
     private KeyBinding keyBinding;
+    public boolean wasPressed;
 
     public KeyHolder(String name, int key, String tab, boolean press, Runnable onAction) {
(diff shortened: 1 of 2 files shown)


HeyZeer0:
Hey @BurstingF, I have no idea what you mean by hotkey action, also, when ``isPress`` is ``true``, that means it should just be fired one time, which is already happening. I'm not sure what this commit is supposed to do

BurstingF:
@HeyZeer0 
By hotkey action I meant the `Runnable onPress` of the `KeyBinding`. According to messages on discord and some testing, it appears that sometimes (not always) hotkeys where `isPress` set to `true` can still fire multiple times when the hotkey is held down. Here is an example where the stopwatch button is held down but the action is fired multiple times: 
JI3HdHfh7I
I'm not 100% sure what is causing it but I think it might be a mix between the continuous use of the `isPressed()` function (which in the description says it shouldn't be used continuously) and keyboard repeat delay/rate. Anyways the changes in this commit seem to fix the issue.

PR #164, opened 2019-12-09
Added pvp to teamname check for nametags

BurstingF:
It appears people in pvp(hunted) mode have a different team to add the [PvP] prefix to their name. These changes should make it so that people in pvp mode now also get their wynncraft nametag.


BurstingF:
I believe currently the nametag people will get in pvp mode would be something like "[PvP] Wynncraft %rank%". Should the prefix [PvP] be removed from the nametag?

HeyZeer0 (changes requested):
Hey, I think your idea is actually great, but I think you should actually split the PVP tag into a different one, saying ``Hunted Mode`` in ``red``.

BurstingF:
@HeyZeer0 is this what you were looking for?

PR #177, opened 2020-02-02
Log damage done to Game Update overlay

mmcgill:
First: **Thanks for the awesome mod!**
Second: **Feedback welcome!** I'm happy to make any recommended changes/additions/etc.

This commit adds toggle-able (defaulting to off) logging of damage done (total, and by type) to the Game Update Overlay, as shown below:

2020-02-02_16 21 44_damage-logging-example

Primary use case is for testing actual damage output against various mobs, particularly those with resistances. If find that the in-game text is often not visible long enough to make out, or obscured by particle effects.

In a future commit, I intend to include keeping track of overall DPS over a sliding (e.g. 5-10 second) window of time.

I've added this as a separate module, but acknowledge that might be overkill. If the maintainers prefer, I can refactor to bundle the functionality in the Utilities module instead.


HighCrit:
It also logs the damage done by everyone else that is around though 🤔

mmcgill:
> It also logs the damage done by everyone else that is around though 🤔

Sure does :/ I can't think of any way to reliably tie damage to the damage dealer, based on how the information is conveyed to the client (in armor stand name tags). The current approach for tying damage to _damaged_ entity is also quite approximate.

HeyZeer0:
Hey @mmcgill, thanks for your PR!
I'm glad to see that you followed our code style.

Your PR is amazing, but I see that you created an entire module just for a feature, that's not how it's supposed to be.
Can you move your code to the utilities module?

mmcgill:
> Can you move your code to the utilities module?

Absolutely! I'll amend the PR presently. Thanks for taking a look so quickly :)

HeyZeer0:
Okay so, now I think we just have to figure out some things:

- Be creative and figure out a way to detect only damage done by the player
- Make the GameUpdateTicker messages a bit more friendly and less spammy

For the last one, we could create a new overlay to, instead of calculating each mob separately, calculate the overall damage in some period like 5s

Any ideas? =P

src/main/java/com/wynntils/modules/utilities/configs/UtilitiesConfig.java
```diff
@@ -99,6 +99,9 @@
     @Setting(upload = false)
     public String lastServerResourcePackHash = "";
 
+    @Setting(displayName="Log Damage to Game Update Overlay", description="Should damage be logged to the Game Update Overlay?")
```
HeyZeer0: This needs to be moved to the GameUpdateTicker configurations, inside Overlay Configurations.
Of course only if we're going to keep it this way

HeyZeer0 (changes requested):

src/main/java/com/wynntils/modules/utilities/configs/UtilitiesConfig.java
mmcgill: Moved.

mmcgill:
> Okay so, now I think we just have to figure out some things:
> 
> * Be creative and figure out a way to detect only damage done by the player
> * Make the GameUpdateTicker messages a bit more friendly and less spammy
> 
> For the last one, we could create a new overlay to, instead of calculating each mob separately, calculate the overall damage in some period like 5s
> 
> Any ideas? =P

I've thought about detecting player for a while now, and unfortunately I think it's fundamentally impossible to determine, or even reliably approximate, damage sources on the client without help from the server.

The damage information we're detecting is being pulled out of armor stands spawned from the server, and there's nothing about the armor stands themselves that could be used to infer or even guess at damage source. Imagine, for example, a situation where two mages have landed Meteors in the same area (with burning ground therefore in effect), and both have poison gear. There will be periodic damage occurring for each mob in the area from each burning ground, and afterward from each player's poison. There's no way to tell which damage goes with which player.

I checked into how the server triggers damage animations on the client. It's done via an [Entity Status](https://wiki.vg/Protocol#Entity_Status) packet. Unfortunately, the damage source isn't passed along to the client.

I _do_ like the idea of a separate overlay that can be enabled to display a rolling average of DPS. That should be easy to do, and useful for e.g. judging effectiveness of different builds at a grind spot. I'll take a stab at it.

ghost:
We don't have to put it in the game update ticker. A new overlay that's customizable might also do. With options such as show target and show total damage or damage per type. With the later only showing the types that actually did damage will free up screen real estate making it less obtrusive.

For the damage source however, in my opinion it being off by default with a warning on the toggle would suffice.

mmcgill:
stats-overlay

Here's a first stab at a 'Stats' overlay that shows DPS computed over a configurable interval of time (10 seconds by default).

If you're happy with the direction this is heading, I've got other things in mind for this overlay that might be more interesting/useful.

*EDIT:* Looking at that screenshot, the DPS stats could be improved by putting them in two columns, and by not repeating the 'DPS' string for each damage type.

HighCrit:
I personally like the direction this is heading, and, was just about to suggest your edit. I'll ping Zeer0 to look at it as well.

HeyZeer0:
Wow, that seems great! I think it can be a bit more compact. Try not showing damages with 0 DPS, 
also maybe put it on the same line? It seems insanely big for low resolution screens, thought on something like:
``{earthSymbol} {damage} {fireSymbol} {damage}...``

What if we also deactivate the overlay at big cities? In theory, with that, the overlay would avoid constantly being triggered by nearby players.
So, what do you think? :)

mmcgill:
stats-overlay2

How about this?

I'm aiming to keep the stats overlay columnar because I'd like to add stats to it over time (with visibility of each stat being individually configurable, of course).

For that reason, I'm hesitant to hide '0' values because of the possibility of causing other stats to  frequently jump between positions.

mmcgill:
> What if we also deactivate the overlay at big cities? In theory, with that, the overlay would avoid constantly being triggered by nearby players.

What if we add an input binding to toggle stats overlay visibility? That would give the player more control (e.g. to quickly hide the stats outside a city when crossing paths with another player), and also avoid having to hard-code a bunch of city bounding boxes.

HeyZeer0:
~~totally not clicked close by mistake, twice~~
The only problem with keys is that most players doesn't check it frequently and have no idea that they exists. About the bouding boxes, they already exists in our API, so would be easy to determine if it's a city or not.
The new design looks great tho, the only thing that still gets me it's the size of the overlay and it constantly appearing which pollutes the screen _even more_

mmcgill:
> The only problems with keys is that most players doesn't check it frequently and have no idea that they exists. About the bouding boxes, they already exists in our API, so would be easy to determine if it's a city or not.

Oh! Ok, I can take a stab at auto-hiding in cities then.

> The new design looks great tho, the only thing that still gets me it's the size of the overlay and it constantly appearing which pollutes the screen _even more_

Does it help that it's off-by-default? Anyhow, how about I hide entries for which damage is zero, but keep things in 'absolute' positions so they don't bounce around?

mmcgill:
> About the bouding boxes, they already exists in our API, so would be easy to determine if it's a city or not.

Are you speaking here of the data returned by `https://login.wynntils.com/cache/getTerritoryList`? If so, are you thinking we'd maintain a client-side list of the names of the big cities within which we want to suppress damage statistics? Or would we add a key to the territory profile indicating whether or not it represents a city?

HeyZeer0:
We are going to change the API, including if it's a city or not, but first we need to compile a list of all cities territories! Since I'm completely unavailable right now I'll ask another developer to

mmcgill:
Makes sense. Do you think this PR could be merged prior to that change, if I create an enhancement issue to track suppression in cities? I can circle back when the API's been updated. Or would you prefer to wait?

HeyZeer0:
I think first we should try to figure out a way to make the overlay more compact, after that I would gladly merge. Sorry for not being able to help at all, I'm completely unavailable recently :(

PR #183, opened 2020-05-19
Add labels to map

magicus:
This patch will add text labels (names of places) to the in-game map, just as found on map.wynncraft.com.

The labels are pulled from map.wynncraft.com, and processed into a json format by a simple script. The script is currently inside the `script` directory. You may want to put the json file on your Athena service eventually, instead of bundling it with the plugin. (I've written some non-tested, commented-out code to achieve this, based on other methods in the WebManager class.)

I am pleasantly surprised at how good this looks. It really fills a need I've had for a long time with the Wynntils map. (While I'm not as much of a Wynncraft noob anymore, I still suck at remembering names and places, and I was forced to swap out Wynncraft to map.wynncraft.com to check those.)

### DIFF
--- a/scripts/update-labels-json.sh
+++ b/scripts/update-labels-json.sh
@@ -1,0 +1,77 @@
+#!/bin/sh
+# This script will retrieve the labels.js file from map.wynncraft.com, parse
+# the locations and coordinates that are present in it, and convert it to a
+# json file.
+MYDIR=$(cd $(dirname "$0") && pwd)
+TARGET="$MYDIR/../src/main/resources/assets/wynntils/labels.json"
+
+command -v curl > /dev/null 2>&1
+if test $? -ne 0; then
+  echo curl is required
+  exit 1
+fi
+
+command -v gawk > /dev/null 2>&1
+if test $? -ne 0; then
+  echo gawk is required
+  exit 1
+fi
+
+command -v jq > /dev/null 2>&1
+if test $? -ne 0; then
+  echo jq is required
+  exit 1
+fi
+
+curl -s https://map.wynncraft.com/js/labels.js | gawk '
+function jsonval(v)
+{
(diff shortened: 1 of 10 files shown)


HighCrit:
You are aware territories are already available in the map when holding Ctrl or is this not what you seek to achieve?

magicus:
@HighCrit Yes, I know, but that is not really the same feature. Some differences:
* The names of the territories does not always align with the names of features. For instance, there's no "Saint's Row" territory.
* Instead of just a single "Detlas", there's a lot of Detlas-<direction> territories.
* When you zoom out, the names of all territories just overlap (unlike a proper label support, where zooming out will remove less relevant labels).

The use cases here is:
* To help the user orient themselves and features on the map (I think the importance of this is hard to understand for people with a normal sense of direction :-))
* To find unfamiliar places ("Oh, the blacksmith in Bucie, let's see, where is that now?")

That is not really the same use case as the territories are solving.

HeyZeer0:
Since I can't really preview how it currently looks would you mind indexing a screenshot? ^^
@magicus

magicus:
@HeyZeer0 I posted the images in Discord, since I cannot attach them here to a PR. :(

HeyZeer0:
I moved your script to a static repository
https://github.com/Wynntils/Data-Storage

There we will be able to pull the data from it and implement it over Athena. I'll be including it inside the route [mapLocations](https://athena.wynntils.com/cache/get/mapLocations) over a new JSON Object.

The final structure will look like:
```json
{
     "locations": [
         ...
     ],
     "labels": [
         ...
     ]
}
```

HeyZeer0:
Successfully implemented over Athena, the route is already updated ^^

HeyZeer0:
Merged to development, thanks for your contribution <3

PR #184, opened 2020-05-21
Avoid NPE stack trace if CoreDBConfig singleton is not instantiated

magicus:
At startup, CoreDBConfig singleton might not be instantiated. Check for null to avoid printing a NPE stack trace.

I got annoyed at this always appearing when I was debugging my other branch.

### DIFF
--- a/src/main/java/com/wynntils/core/framework/settings/SettingsContainer.java
+++ b/src/main/java/com/wynntils/core/framework/settings/SettingsContainer.java
@@ -63,8 +63,10 @@ public class SettingsContainer {
         }
 
         try {
-            if (CoreDBConfig.INSTANCE.enableCloudSync) {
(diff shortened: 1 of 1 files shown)


HighCrit (changes requested):
This is not the fix for this issue. As you might as well remove the entire try block. CoreDBConfig should be loaded before that function is run on startup.

magicus:
Fair enough. The SettingsContainer constructor is run multiple times during startup, and only at the first run is CoreDBConfig not instantiated. I tried looking at the code to see if there were some obvious ordering issues, but I did not understand the startup procedure good enough to spot anything out of the ordinary.

HeyZeer0:
Since this configuration was not properly done, I'll be temporarily removing it

PR #187, opened 2020-05-25
Add UI to share coordinates

magicus:
This patch adds a GUI way to share your current location or compass coordinates with either your party, or your friends.

Additionally, a hover help text is added, not just to the new share button, but also to the existing "Add waypoint" button, for consistency.

Finally coordinates made clickable in chat is now underlined to emphasize that they are indeed clickable.


HeyZeer0:
Ok, we might have some problems with this one:

1. The UI is limited to only 15 friends, and there's a loooot of people with tons of friends
2. In a list with a lot of friends, it would be kinda hard to find the one that you're looking for
3. The UI looks kinda ugly since it has no visuals at all
4. The button is quite invisible, took me nearly 2min to see that it was there

We do have plans on retexturing the entire world map and we already have textures done, I'm not sure how exactly this would fit. It seems like it's a separate system that could be integrated with something more complex, like a friend book or similar ^^

magicus:
While I could make a multi page version, as you say, if you have too many friends, this is not an effective way to select them.

I do think this would be usable for players with less friends (my use case is how I play with my kids and wife), but I accept if you do not think this is enough reason to merge this.

Regarding 3, yeah, sorry, I'm not really a UI designed (although I did take courses on HCI in university). I spent longer than I care to admit to even get it to this level. :-)

I'm not sure I understand 4. Do you mean the share arrow from the map, or the "Share" button at the end of the screen? 

The former is sort of intentional -- my intention was for this to be available from the map, but not intrude on the user. I think the tooltip makes it discoverable enough to new users; at least not worse than the "add waypoint" plus sign. 

If you mean the latter, chalk it up to the rest of 3. :)

magicus:
And idea to handle 1, 2 and 3:
I remove the possibility to share the position with friends, and only allow sharing with your party. (After all, that's the main use case as I see it anyway).  This means that I do not need to deal with large number of friends, and I do not even need to display a huge friend selection screen.

If you think this sounds like an acceptable way forward, let me know and I can start transforming this patch. You can always say no later if you still don't think it's okay. :-)

HeyZeer0:
Seems more like an idea for the future when the Map UI be renewed, I'll be closing this for now!

PR #189, opened 2020-05-25
Add WynnData Builder integration

magicus:
This patch will add an additional button when opening the "compass" player profile, below the Ingredient Filter. 

Pressing this button will open the Builder at WynnData.tk, with your currently equipped armor, accessories and weapon.

Current limitations:
* Powders are not parsed
* Weapon is assumed to be in first slot in hotbar


HeyZeer0:
Great addition, made some syntax changes and updated the class name to be proper referring to WynnData ^^

Thanks again! <3

magicus:
Thanks! And thank you for improving the code. At least the "opened" thingy was just embarrassing. :) I can only blame that I moved the code from elsewhere and didn't think it through.

PR #190, opened 2020-05-26
Implement item lookup on WynnData

magicus:
This PR adds the functionality to do item lookups on WynnData. To enter item lookup mode, right click on the WynnData share button. You can then right click on any item to open it in WynnData's item database. You can exit the item lookup mode by leaving the character information page, or by right clicking again on the WynnData share button.

The rationale for the item identification mode is to avoid accidental right clicks which opens the browser unexpectedly.

This PR also fixes a bug with the WynnData share button. I only intended for it to appear on the character info page, but it appeared on all chests (!).

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/overlays/inventories/WynnDataOverlay.java
+++ b/src/main/java/com/wynntils/modules/utilities/overlays/inventories/WynnDataOverlay.java
@@ -14,6 +14,7 @@ import net.minecraft.client.audio.PositionedSoundRecord;
 import net.minecraft.client.gui.GuiButton;
 import net.minecraft.init.Blocks;
 import net.minecraft.init.SoundEvents;
+import net.minecraft.inventory.Slot;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 import net.minecraft.util.NonNullList;
@@ -65,10 +66,12 @@ public class WynnDataOverlay implements Listener {
     @SubscribeEvent
     public void clickOnChest(GuiOverlapEvent.ChestOverlap.HandleMouseClick e) {
         if (!Utils.isCharacterInfoPage(e.getGui()) || !WynnDataOverlay.itemLookupMode ||
-                e.getMouseButton() != 1 || e.getGui().getSlotUnderMouse() == null ||
-                e.getGui().getSlotUnderMouse().inventory == null) return;
+                e.getMouseButton() != 1) return;
 
-        ItemStack stack = e.getGui().getSlotUnderMouse().getStack();
+        Slot slot = e.getGui().getSlotUnderMouse();
+        if (slot == null || slot.inventory == null || !slot.getHasStack()) return;
+
+        ItemStack stack = slot.getStack();
         Utils.openUrl("https://www.wynndata.tk/i/" + Utils.encodeItemNameForUrl(stack));
         e.setCanceled(true);
     }
@@ -97,7 +100,7 @@ public class WynnDataOverlay implements Listener {
             Minecraft.getMinecraft().getSoundHandler().playSound(PositionedSoundRecord.getMasterRecord(SoundEvents.UI_BUTTON_CLICK, 1f));
 
             Map<String, String> itemNames = new HashMap<>();
-            
+
             NonNullList<ItemStack> armorInventory = Minecraft.getMinecraft().player.inventory.armorInventory;
             getItemNameFromInventory(itemNames, "helmet", armorInventory, 3);
             getItemNameFromInventory(itemNames, "chestplate", armorInventory, 2);
@@ -124,5 +127,5 @@ public class WynnDataOverlay implements Listener {
             Utils.openUrl(urlBuilder.toString());
         });
     }
-    
+
 }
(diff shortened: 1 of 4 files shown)


src/main/java/com/wynntils/modules/utilities/overlays/inventories/WynnDataOverlay.java
```diff
@@ -16,36 +17,62 @@
...
+    }
+
+    @SubscribeEvent
+    public void clickOnChest(GuiOverlapEvent.ChestOverlap.HandleMouseClick e) {
+        if (!Utils.isCharacterInfoPage(e.getGui()) || !WynnDataOverlay.itemLookupMode ||
+                e.getMouseButton() != 1 || e.getGui().getSlotUnderMouse() == null ||
+                e.getGui().getSlotUnderMouse().inventory == null) return;
+
+        ItemStack stack = e.getGui().getSlotUnderMouse().getStack();
+        Utils.openUrl("https://www.wynndata.tk/i/" + Utils.encodeItemNameForUrl(stack));
```
coehlrich, line 64: Shouldn't there be a `hasStack` check for the slot or `isEmpty` check for the item stack?

coehlrich (changes requested):
That was just something I noticed a while ago (didn't realise I had to finish the review)

src/main/java/com/wynntils/modules/utilities/overlays/inventories/WynnDataOverlay.java
magicus: Good point, I'll fix.

coehlrich (approved):

HeyZeer0:
As long as this seems a great idea, it doesn't seem to be very necessary to add a button to toggle this option, I think maybe a key shortcut would be actually way smarter and easier for the end-user. Something like ``ctrl + shift + click`` would be the ideal.

magicus:
@HeyZeer0 I tried implementing this. It turns out that is a very bad idea to hold down modifier keys when opening a URL in the browser. Since this will shift focus from the Minecraft Window to the browser, the release of the shift key will not be registered by Minecraft, so it will be effectively stuck. :-(

I tried some ways to work around this, but this is basically a hard limitation due to the interaction between the OS and minecraft, and how lwjgl is implemented.

magicus:
But maybe the button toggling is too much overhead for the user. I'll push an update that only requires the middle mouse button pressed to open on WynnData. Since this is limited to the character info page, hopefully you will not get sent to the browser in the middle of a dungeon fight by mistake, or so...

magicus:
... and now I remember why I did not choose this solution in the first place. :-(

If I do that, the item will be considered picked up, so the slot I'm checking is empty, and everything just fails. 

So unfortunately I cannot find a better way of doing this than the original patch. I recognize that it's not ideal, but I think it's the best we can do given the limitations.

HeyZeer0:
> @HeyZeer0 I tried implementing this. It turns out that is a very bad idea to hold down modifier keys when opening a URL in the browser. Since this will shift focus from the Minecraft Window to the browser, the release of the shift key will not be registered by Minecraft, so it will be effectively stuck. :-(
> 
> I tried some ways to work around this, but this is basically a hard limitation due to the interaction between the OS and minecraft, and how lwjgl is implemented.

Oh, that's true but it's easily fixable, you can just execute the method ``Keyboard#reset`` to force it to release all current hold keys and come back to its original state. This method is currently private so you'll need some Reflections to invoke it, check [ReflectionMethods ](https://github.com/Wynntils/Wynntils/blob/development/src/main/java/com/wynntils/core/utils/reflections/ReflectionMethods.java) for more info ^^

HeyZeer0:
Btw, just a quick tip, don't limit yourself to Minecraft. Minecraft base code is almost 100% replaceable and very hacky things can be done. Wynntils has a lot of these implemented to offer us a better API and faster responses. That's a thing we learned in the process of creating this mod xD

magicus:
I can look into Keyboard.reset(). But I don't know when will be a safe time to call it.

I'm worried what happens if you call it and do hold down a key, but the Minecraft window do not lose focus. I assume this means the "key down" status will be forcefully lost, in a situation where it is not apparent to the user. From looking at the Keyboard.reset code, it looks like it just clears the status of all held-down keys. I think a better approach would have been to re-scan the keyboard for pressed-down keys. But maybe things do not work that way anymore; USB keyboards maybe only send events on key press?  I have not really worked this close to hardware since the PS/2 keyboard adapter was a novelty replacing the old DIN keyboards with their grid scanning :)

HeyZeer0:
You should call it after opening the user browser. It will have no effects if the key is considered as not pressed anymore, since whenever the client press it again it will register and vice-versa. For the events, I'm not exactly sure but I do think it depends on the keyboard driver, but at its base, it sends the key change. You would have to check how every OS handles that xD

Give it a shot!

HeyZeer0:
Ah also, you can do a quick debug if you can check if the game is in focus after calling the browser. Check ``Display#isActive``

magicus:
If this works out fine, I should probably make the change in openUrl() instead. I checked, and opening quests on the wiki suffers from the same problem; if you happen to hold down a key when right clicking, it will become "stuck" in Minecraft.

HeyZeer0:
Awesome, feel free to make that ^^

magicus:
Ok, here's a new implementation that *almost* work. :-) The theory seems sound. It works well in development mode. But when running it deployed, it does not work. My current guess is that the reflection does not work. As far as I can tell the lwjgl library is not obfuscated, but maybe I'm wrong there.

magicus:
I'd appreciate any assistance regarding the Keyboard.reset() call not working. I don't have any clear idea on how to proceed.

magicus:
I think it works fine now. On my Mac, it works good anyway. I believe the lost pressed key is only an issue on macos; it seems to work in Windows, and the code indicates that there is a fix for that in Linux as well.

PR #191, opened 2020-05-27
Add objectives overlay.

magicus:
This patch will capture the sidebar scoreboard showing the (daily) objective, and redirect it to an overlay. Finally you can get rid of that annoying background behind the tracker! As an added bonus, it is also possible to visualize progress towards the objective as a progress bar.

Unfortunately, Forge presented no hooks for reading (or prohibiting display of) sidebar scoreboard updates. :-( So I had to capture the relevant packages and pre-process them.

Limitations: If you are playing as a beginner in your class, you will not have daily objectives, but can have multiple "starter" objectives. The overlay will only present the topmost of these. (I have a hard time finding a good way to show all, without sacrificing usability for the single-line daily objectives.)

From time to time, things can get weird if you turn on and off the overlay, and switch class.

Also note that if you have disabled the sidebar scoreboard using `/toggle sb`, you need to toggle it back on, otherwise Wynntils cannot get hold of the scoreboard data.


magicus:
I have posted a screenshot on Discord.

HighCrit:
You can disable the scoreboard by setting a variable to false. Iirc in the forgeui class

magicus:
Well...  I looked at several options, but I could not find a way to disable the standard sidebar scoreboard from rendering, while at the same time being able to extract the data.

magicus:
For instance, GuiIngameForge calls renderScoreboard unconditionally. Too bad. Found no way to replace GuiIngameForge with a specialized subclass without advanced Forge hackery.

magicus:
I would have loved a pre-hook in renderScoreboard.

HighCrit:
Have you set GuiIngameForge.renderObjective to false? You'll need to undo this on server leave

magicus:
@HighCrit No. The Forge mechanism for this was not usable for this scenario. I drop the scoreboard update packets instead, after extracting the information I need. With no score updates, Vanilla will not render any sidebar scoreboard.

HeyZeer0:
Okay, this overlay is quite gigantic, I think we can actually create a proper texture to it.

I have 3 ideas for it:
1. Create a new and tiny texture for the overlay
2. Let it permanently visible on the game menu (when you press ESC by default)
3. Show it, in the user HUD, for a tiny fraction of a second whenever the objective updates.

Also, have you checked if this is handling Multiple Objectives properly? There are some specifics occasions where this happens ^^

magicus:
Personally, I start my wynncraft sessions most of the time by doing the daily objective. So I really like to be able to see the status and progress all the time, without updating the objective, or pressing Esc. So that definitely needs to be an option to have it like that, otherwise it'll be a regression compared to the current sidebar Wynncraft solution.

However, I can see the appeal in just having it flash up for a while when the objective updates. That should be fairly easy to implement. It can even be on by default.

magicus:
I'm not sure what you mean by 1. Is it the progress bar you are talking about?

magicus:
And finally, I don't understand what you mean by gigantic. Maybe I'm defining the overlay size wrong..? On my computer it's quite small, just big enough for the objective text, counter and the progress bar. But there seem to be some kind of dynamics in resizing that I don't fully understand, so maybe I'm configuring it wrong.

magicus:
And as for multiple objectives, as I said in the PR description, this is a known limitation. Only the first objective will be shown and tracked on the progress bar.

Does this happen in other circumstances than when you start a new class?

I don't have any good ideas on how to solve this, that do not require an inordinate amount of work, for a very special case, or impacts the "normal" use. :(

My two ideas is to either make the overlay box larger, and show multiple objectives at the same time. This means it's tricky to fit the box on the default screen without overlapping. Or, we could let the different objectives "rotate" on the same spot, e.g. every 5th second. (And pop up immediately if they change.) I think it could look good, but will require a lot of work for a special case. A third option is to inactivate the overlay if you have multiple objectives, and just keep the vanilla Wynncraft sidebar for those instances.

magicus:
Maybe overlapping overlays is not a big deal, as long as the actual graphics on them do not overlap most of the time. I'll try to implement a solution where I can display multiple objectives at the same time, and see how that works out.

HeyZeer0:
--------

> And finally, I don't understand what you mean by gigantic.

Using the experience bar for displaying such small info seems like overkill since we kinda already have a loooot of overlays.

> so that definitely needs to be an option

For sure.

--------

I'll need to see what I can do about the texture. But I think that we can create a good one for it ^^

magicus:
I moved this PR back info Draft state, since I have not been fully able to test all changes. (I can only test Daily Objectives once per day! *doh*)

New features/changes/fixes:
* Handle multiple objectives (up to three; never seen more than that)
* Allow setting opacity (alpha) for the overlay.
* Option to hide overlay on inactivity. Overlay will reappear when changed, or when you
access a chest or inventory (to be able to be reminded of current objectives). Overlay fades out when hiding.
* Combine packet snooping with GuiIngameForge.renderObjective; this way we get all information we need, but can still quickly replace the vanilla scoreboard with the overlay.
* When Daily Objective is done, display a reminder until you have fetched your loot.
* Bug fixes for enabling/disabling overlay, and when objectives are finished.

Remaining issues:
 * Not fully tested.
 * Stacking of objectives, if less than the maximum, is not very well handled. The
user should be able to select if they should grow from top to bottom.
 * The code in ObjectivesOverlay could need one more pass at improving code quality.

magicus:
I have posted a screenshot to Discord. (Once again, since attachments cannot be done to Github PRs.)

magicus:
The functionality is confirmed working. I was thinking I needed to make sure there were e.g. no "hole" if you finished the middle objective out of three, but it turned out that (of course) I do not need to worry about that -- Wynncraft will reorganize the objectives for me, and I just mirror that.

However, I'm still not happy with how the code looks in ObjectivesOverlay. And I would like to add an option to select if the list of objectives grows from the top or from the bottom. So I'm keeping this as draft for a bit longer.

magicus:
So, I'm finally happy with this. :)

HeyZeer0 (approved):
I'm happy as well ^^

PR #195, opened 2020-05-29
Locate weapon for WynnData build integration

magicus:
Locate weapon in hotbar, starting from left-most slot. This is the same behavior as the weapon info in the Character Info page. If no weapon is located, the WynnData weapon slot is left empty.


HeyZeer0 (changes requested):
You might wanna check if the class Req fills the user class. Check Player Info to get the user class, the enum should already have the default text

magicus:
Good point. Fixed.

HeyZeer0 (approved):

PR #197, opened 2020-05-31
Print the coordinates of where the loot run start when loading it

magicus:
Print the coordinates of where the loot run start when loading it. (Which are clickable, so you can set your beacon there easily.)


HeyZeer0 (changes requested):
You might wanna color the message, could be a simple gray or dark gray.

PR #200, opened 2020-06-02
Fix bugs on consumables from chat

magicus:
Fix bugs on consumables from chat reported by SirCmt:
 * Use Minecraft.getSystemTime() to avoid time discrepancy on Windows.
 * Subtract a second to better align with actual time set by server.


magicus:
Setting as draft while I await the lastest testing by SirCmt.

magicus:
Confirmed working.

magicus:
Just to be clear: I had missed to push the "other player" fix to github; but the confirmation by SirCmt was with that commit too.

PR #212, opened 2020-06-13
green/red colors for enabled/disabled in settings

McPlayHD:
someone was asking for colored "enabled" and "disabled" in the Discord and since it got upvoted and I like the idea as well I quickly implemented it.


DevChromium:
Tested this and it works as intended

HeyZeer0 (changes requested):
Hey! Thanks for your contribution ^^
I would change this by myself but since I'm currently not able I'll ask this to you.

Instead of using directly color codes would you mind replacing them with the TextFormatting color enum? 

Thanks again!

PR #213, opened 2020-06-14
Update SkillPointOverlay to use a better method of collection and include more overlays 

aws404:
Changes the for loops that were previously used and using substring instead and replacing colour codes with the TextFormatting method rather than regex.
Also adds additional number overlay to the 'Damage Info' icon (average mele damage) and the 'Daily Rewards' icon (current streak).


magicus (approved):
Looks good to me.

PR #214, opened 2020-06-15
Item level requirement overlay

magicus:
Adjust the background color "highlight" overlay so it also indicates the combat level requirements of items. It does this by showing a "clock" style arc in the background, which grows to 100% full at level 105. This can be somewhat distracting, and is disabled by default.

Also shows the combat level as a number on the item (re-using item stack count) when holding down ctrl. For unidentified items, the level is a span (like 60-64); in this case the average value (62) is displayed.


magicus:
I think this worked out OK, but I'd appreciate if someone could give it a second pair of eyes. I unified the code between the two places where the highlight were drawn, for inventory and chest/horse. There might be even more cleanup possible, but this was enough for me.

I think the best part is holding down ctrl to show the level req. as "stack count". It is probably even better than the arc. But it was okay, so I kept it. It is disabled by default though, since it is a bit distracting.

HeyZeer0 (approved):

magicus:
Item level requirement overlay

Adjust the background color "highlight" overlay so it also indicates the combat level requirements of items. It does this by showing a "clock" style arc in the background, which grows to 100% full at level 105. This can be somewhat distracting, and is disabled by default.

Also shows the combat level as a number on the item (re-using item stack count) when holding down ctrl. For unidentified items, the level is a span (like 60-64); in this case the average value (62) is displayed.

PR #215, opened 2020-06-16
created a way to mark trades as "favorite" in the trade market (default key "F")

McPlayHD:
You can mark items as favorite so that you can track them better. The favorites will get cleared if you close the inventory or switch to another item. But you can scroll through pages if there are a lot of offers for a specific item.
You can switch the key under "Controls" if you want to


src/main/java/com/wynntils/modules/utilities/overlays/inventories/FavoriteTradesOverlay.java
```diff
@@ -0,0 +1,108 @@
...
+        if (!Reference.onWorld) return;
+        for (Slot s : e.getGui().inventorySlots.inventorySlots) {
+            if (s.slotNumber >= e.getGui().getLowerInv().getSizeInventory()) continue;
+            if (isNotMarketItem(s.getStack())) continue;
+            renderFavoriteItem(s, e.getGui().getGuiLeft(), e.getGui().getGuiTop());
+        }
+    }
+
+    @SubscribeEvent(priority = EventPriority.NORMAL)
+    public void onInventory(TickEvent.RenderTickEvent e) {
```
HeyZeer0: This is not necessary you can check the inventory name on the event above (onChestGui)

src/main/java/com/wynntils/modules/utilities/overlays/inventories/FavoriteTradesOverlay.java
```diff
@@ -0,0 +1,108 @@
...
+        char[] c = lore.get(2).toCharArray();
+        for (char cc : c)
+            if (cc == EmeraldSymbols.E)
+                return false;
+        return true;
+    }
+
+    private static void renderFavoriteItem(Slot s, int guiLeft, int guiTop) {
+        ItemStack it = s.getStack();
+        ItemIdentificationOverlay.replaceLore(it);
```
HeyZeer0, line 69: Is there any reasons that you're replacing the lore again?

src/main/java/com/wynntils/modules/utilities/overlays/inventories/FavoriteTradesOverlay.java
```diff
@@ -0,0 +1,108 @@
...
+        String lore = Arrays.toString(ItemUtils.getLore(it).toArray());
+        boolean found = false;
+        for (String lore_fav : favorites_trade_items_lore) {
```
HeyZeer0: This is not necessary at all, you could just check if ``Array#contains`` the lore ^^

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
```diff
@@ -228,6 +229,16 @@ public void keyPressOnInventory(GuiOverlapEvent.InventoryOverlap.KeyTyped e) {
         }
     }
 
+    @SubscribeEvent
+    public void keyPressOnTrade(GuiOverlapEvent.ChestOverlap.KeyTyped e) {
```
HeyZeer0: You can place this event inside your FavoriteTradesOverlay class

HeyZeer0 (changes requested):
Hey, thanks for your contribution ^^

I just need some changes before merging

src/main/java/com/wynntils/modules/utilities/overlays/inventories/FavoriteTradesOverlay.java
```diff
@@ -0,0 +1,108 @@
...
+            if (lore_fav.equals(lore)) {
+                found = true;
+                break;
+            }
+        }
+        if (!found) return;
+        ScreenRenderer.beginGL(0, 0);
+
+        // HeyZeer0: this will make the lock appear over the item
```
magicus, line 75: Maybe update the comment? :) Or remove it fully...

src/main/java/com/wynntils/modules/utilities/overlays/inventories/FavoriteTradesOverlay.java
McPlayHD: it kind of is because if you enter the Marketplace and add Favorite markers and then exit the Marketplace there is no Chest inventory so it does not get cleared. Then when opening the Market again the markers are still visible.
The question is: Maybe we want that to be a thing? We would save resources for sure.

src/main/java/com/wynntils/modules/utilities/overlays/inventories/FavoriteTradesOverlay.java
```diff
@@ -0,0 +1,108 @@
...
+        GlStateManager.translate(0, 0, 260);
+
+        ScreenRenderer r = new ScreenRenderer();
+        RenderHelper.disableStandardItemLighting();
+        ScreenRenderer.scale(0.5f);
+        r.drawRect(Textures.UIs.hud_overlays, (int)((guiLeft + s.xPos) / 0.5) + 20, (int)((guiTop + s.yPos) / 0.5) - 3, 51, 0, 17, 16);
+        ScreenRenderer.endGL();
+    }
+
+    public static void checkLockState(ItemStack it) {
```
magicus: I think. `toggleLockState` would be a more descriptive name.

magicus (changes requested):

src/main/java/com/wynntils/modules/utilities/overlays/inventories/FavoriteTradesOverlay.java
McPlayHD: I have to do this because sometimes wynncraft updates the items in the inventory and as long as you don't hover over them the Lore will not get updated -> the Favorite marker is gone

src/main/java/com/wynntils/modules/utilities/overlays/inventories/FavoriteTradesOverlay.java
McPlayHD: yep true... Forgot to update this when I changed from List<String> to String
Will do a commit for that.

src/main/java/com/wynntils/modules/utilities/overlays/inventories/FavoriteTradesOverlay.java
McPlayHD: I let it there because it was HeyZeer0's work and I could just copy it so I let the author there... If he wants to I can remove it.

src/main/java/com/wynntils/modules/utilities/overlays/inventories/FavoriteTradesOverlay.java
HeyZeer0: you can clean the array if the inventory is not the market ^^

src/main/java/com/wynntils/modules/utilities/overlays/inventories/FavoriteTradesOverlay.java
HeyZeer0: Got it

src/main/java/com/wynntils/modules/utilities/overlays/inventories/FavoriteTradesOverlay.java
McPlayHD: I see... But when I remove it then the following will occur:
1. Open market
2. On the front page mark favorites
3. Close market
4. Open market
5. Favorites will still be marked

McPlayHD: because the event only gets called if a chest inventory is open.
Technically it's not a big deal but I just added it for consistency... I can easily remove it because yea... it just takes resources for no real addition.

HeyZeer0: Sure, do it :D

McPlayHD: done

HeyZeer0 (approved):

PR #236, opened 2020-07-13
Fixes to Consumable timer

TrapinchO:
A really little thing. Previously you would get "War Scream" while using Vanish.

EDIT:
* Fixed Assassin getting "War Scream" during Vanish (which also didn't disappear)
* Renamed Timer during Vanish to "Vanish" and based it on "invisiblity" effect instead of "speed boost"
* Added comments

(This is my first PR and I hope I didn't mess up)


magicus (changes requested):
I think would be a step backwards. The term "Resistance" is not used in Wynncraft. If this gets mixed up with Vanish, we need to fix it, but I'd suggest instead figuring out properly the requirements for War Scream (and, for that matter, Vanish).

magicus:
One way to solve this is to complement ConsumableTimerOverlay.addBasicTimer with a "refillBasicTimer", which checks if there already is a timer with that name, and if so, updates it's timer.

This might still not be enough, if you have War Scream active and then get Vanish, it will likely mis-fire. We might need to look for both requirements for War Scream, that is, two different effects.

TrapinchO:
Wouldn't be simpler to create timer for every effect you can get? (speed, resistance, strenght, invisibility)

magicus:
Simpler for the developer, maybe, but not as good a user experience. As a user, you don't really care that slow weapon speed is implemented using "mining fatigue". You have a slow weapon. And you do your War Scream spell.

TrapinchO:
> As a user, you don't really care that slow weapon speed is implemented using "mining fatigue". You have a slow weapon.

From what I know it is only decorative

This is not a final commit (I hope)

found bugs:
* after logging in you sometimes get "Vanish (27:xx)"  EDIT: fixed
* "War Scream" timer might not appear during Vanish (not tested) EDIT: fixed as much as I could

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -677,27 +677,54 @@ public void onPlayerDeath(GameEvent.PlayerDeath e) {
...
 
+        // no idea what this does
         SPacketEntityEffect effect = e.getPacket();
         Potion potion = Potion.getPotionById(effect.getEffectId());
+
+        // if effect belongs to player?
         if (effect.getEntityId() == Minecraft.getMinecraft().player.getEntityId()) {
-            String timerName;
-            if (potion == MobEffects.SPEED) {
+            String timerName = "";
```
magicus:
```suggestion
            String timerName;
```

magicus: This will make timerName effectlively final, so you don't have to copy it before the lambda.

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -677,27 +677,54 @@ public void onPlayerDeath(GameEvent.PlayerDeath e) {
...
-            } else if (potion == MobEffects.RESISTANCE && effect.getAmplifier() == 2) {
-                timerName = "War Scream III";
+            }
+            // if the effect is invisibility timer is "Vanish"
+            else if (potion == MobEffects.INVISIBILITY && effect.getDuration() < 200) {
+                timerName = "Vanish";
+                isVanished = true;
+            }
+            // if the player isn't invisible (didn't use vanish)
+            else if (!isVanished) {
```
magicus: "isVanished" is true if and only if the previous if condition tested true, so you can just turn this into a simple else
```suggestion
            else {
```

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -677,27 +677,54 @@ public void onPlayerDeath(GameEvent.PlayerDeath e) {
...
 
+
+    static boolean isVanished = false;
     @SubscribeEvent
-    public void onEffectApplied(PacketEvent<SPacketEntityEffect> e) {
+    public void onEffectApplied(PacketEvent<SPacketEntityEffect> e)
+    {
         if (!Reference.onWorld || !OverlayConfig.ConsumableTimer.INSTANCE.showSpellEffects) return;
 
+        // no idea what this does
```
magicus: It just extracts the actual effect of the layers of data structure that wrap around it. :)

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -677,27 +677,54 @@ public void onPlayerDeath(GameEvent.PlayerDeath e) {
...
         SPacketEntityEffect effect = e.getPacket();
         Potion potion = Potion.getPotionById(effect.getEffectId());
+
+        // if effect belongs to player?
```
magicus: Yes, the client get all effects, even on mobs and other players. We ignore all other.

magicus (changes requested):

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: I kept getting error before even when I had it this way, no idea why…
changed

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: No. Because each effect is resolved individuallly, without it you get both Vanish and War Scream (II).

magicus: Hm. Ok, if you say...

TrapinchO:
If no one has any suggestions or corrections I would call it finished.

magicus (approved):

TrapinchO:
@HeyZeer0 could you please review it?

TrapinchO:
Sorry, closed on accident

magicus:
You may want to update the title of the PR to reflect the actual changes.

TrapinchO:
> You may want to update the title of the PR to reflect the actual changes.

Done, thanks.

HeyZeer0:
One thing needs to be tested, what happens if you vanish and receive the war scream effect?

TrapinchO:
In theory you shouldn't get it. (But I wasn't able to get better solution at the time of doing it)

It might be possible to add "if effect is 'resistance' and 'isVanish' is true set 'isVanish' to false" after the "War Scream" check (before the last "else") (so it would only disable when "invisibility" is resolved and "War Scream" is skipped), however I am away from my pc (and propably won't be for about a month) so I can't test it.

TrapinchO:
Fixed.

The only potential problem is when using vanish I (below lv. 26) isVanished might get stuck on true (which shouldn't be problem though)

TrapinchO:
@HeyZeer0 I think it is ready to merge

TrapinchO:
@HeyZeer0 do you want to accept the PR? If yes, do you want changes to the code?

PR #241, opened 2020-08-21
Durability bars for crafted items

P0keDev:
Overlays a vanilla durability bar that represents the durability of crafted items in inventories, like so:
image

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/overlays/inventories/RarityColorOverlay.java
+++ b/src/main/java/com/wynntils/modules/utilities/overlays/inventories/RarityColorOverlay.java
@@ -131,7 +131,7 @@ public class RarityColorOverlay implements Listener {
 
         CustomColor colour = getHighlightColor(s, is, lore, name, isChest, guiContainer.getSlotUnderMouse());
         int level = getLevel(lore);
-        int durabilityPct = getDurability(lore);
+        float durability = getDurability(lore);
 
         if (level != -1) {
             if (Keyboard.isKeyDown(Keyboard.KEY_LCONTROL) || Keyboard.isKeyDown(Keyboard.KEY_RCONTROL)) {
@@ -142,9 +142,10 @@ public class RarityColorOverlay implements Listener {
         }
 
         // start rendering
-        drawDurabilityBar(guiContainer, s, durabilityPct);
         drawLevelArc(guiContainer, s, level);
         drawHighlightColor(guiContainer, s, colour);
+        drawDurabilityArc(guiContainer, s, durability);
+        
     }
 
     private static CustomColor getHighlightColor(Slot s, ItemStack is, String lore, String name, boolean isChest, Slot slotUnderMouse) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/wynntils/modules/utilities/overlays/inventories/RarityColorOverlay.java
```diff
@@ -211,6 +214,54 @@ private static int getLevel(String lore) {
...
+    
+    private static void drawDurabilityBar(GuiContainer guiContainer, Slot s, int durabilityPct){
+    	if (!UtilitiesConfig.Items.INSTANCE.craftedDurabilityBars) return;
+    	if(durabilityPct == -1) return;
+    	
+    	int x = guiContainer.getGuiLeft() + s.xPos;
+        int y = guiContainer.getGuiTop() + s.yPos;
+        
+        GlStateManager.disableLighting();
+        //GlStateManager.disableDepth();
```
magicus: Please do not check in commented code... Also down below.

magicus (changes requested):
Fix the comments and the code looks good to me.

src/main/java/com/wynntils/modules/utilities/overlays/inventories/RarityColorOverlay.java
```diff
@@ -211,6 +214,54 @@ private static int getLevel(String lore) {
         }
         return -1;
     }
+    
+    private static int getDurability(String lore) {
+    	Pattern p = Pattern.compile("\\[([0-9]+)/([0-9]+) Durability\\]");
```
HeyZeer0: keep the compiled pattern as a static variable

src/main/java/com/wynntils/modules/utilities/overlays/inventories/RarityColorOverlay.java
```diff
@@ -211,6 +214,54 @@ private static int getLevel(String lore) {
...
+        renderer.begin(7, DefaultVertexFormats.POSITION_COLOR);
+        renderer.pos((double)(x + 0), (double)(y + 0), 500.0D).color(red, green, blue, alpha).endVertex();
+        renderer.pos((double)(x + 0), (double)(y + height), 500.0D).color(red, green, blue, alpha).endVertex();
+        renderer.pos((double)(x + width), (double)(y + height), 500.0D).color(red, green, blue, alpha).endVertex();
+        renderer.pos((double)(x + width), (double)(y + 0), 500.0D).color(red, green, blue, alpha).endVertex();
+        
+        Tessellator.getInstance().draw();
+    }
+    
+    private static void drawDurabilityBar(GuiContainer guiContainer, Slot s, int durabilityPct){
```
HeyZeer0: Instead of using the vanilla durability bar which doesn't fit the Wynncraft style, use the drawLevelArc function that @magicus did some time ago

P0keDev: Should the combat level arc be replaced if enabled? I used durability bars to avoid that conflict (and also because I thought people would be more used to a vanilla durability bar regardless of whether it fits as well).

HeyZeer0: Sure, there are 2 different ways that you can see their levels =)

magicus: I'm not sure I agree with this. If you do that, you could not simultaneously view durability and level. And suddenly it get's confusing if the arc means durability or level. Should it mean durability for crafted items, and level for all other?

magicus: Whether the vanilla bar fits the style is a matter of subjective judgement. But at the very least it's meaning is fully clear to anyone who has ever played Minecraft, and in this case it fills the very same role, so I'd argue that the patch is good as it is.

magicus: One solution, in terms of style, could be to draw a somewhat smaller arc, and still color it in the vanilla style. That way you get part of the vanilla experience (the color indication) and part of the level arc wynncraft theme.

(Or you could do both and let the user select the look with a switch)

HeyZeer0: Sure, but keeping the vanilla bar is not an option. It definitely doesn't fit the game style.
What also could be done is changing the arc color for durability

magicus: Yes, that was what I was trying to suggest: use the same color changing as the vanilla bar (green -> red) but draw it as an arc (but with a smaller diameter, so it does not overlap).

P0keDev:
Updated the indicator to use the custom arc, as suggested:
image

It is slightly smaller than the combat level arc so they fit together nicely

HeyZeer0 (approved):

PR #242, opened 2020-08-24
Improved Powder Manual

P0keDev:
Replaces the default powder manual with an in-chat menu that allows you to select each chapter, for a cleaner and less annoying experience. Can be toggled.
image

### DIFF
--- a/src/main/java/com/wynntils/modules/chat/managers/ChatManager.java
+++ b/src/main/java/com/wynntils/modules/chat/managers/ChatManager.java
@@ -5,6 +5,7 @@
 package com.wynntils.modules.chat.managers;
 
 import com.wynntils.ModCore;
+import com.wynntils.core.framework.enums.PowderManualChapter;
 import com.wynntils.core.utils.StringUtils;
 import com.wynntils.core.utils.helpers.TextAction;
 import com.wynntils.core.utils.objects.Pair;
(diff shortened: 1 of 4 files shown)


src/main/java/com/wynntils/modules/chat/managers/ChatManager.java
```diff
@@ -681,5 +708,57 @@ public static boolean processUserMention(ITextComponent in, ITextComponent origi
...
+                "    §e§l✦ Thunder:  §7§oThunder-imbued weapons can release a streak of lightning that seeks enemies, bouncing from one foe to the next nearby. Thunder-augmented armors will steal the life force of the slain, and in kind increase the power of Thunder-based attacks temporarily.\n" +
+                "    §b§l❉ Water:  §7§oWater-imbued weapons will curse one's nearby enemies. This curse weakens armor, and turns them vulnerable to further attack. Water-augmented armors recycle used mana, transferring it into a short boost to Water-affiliated techniques, depending on how much mana was consumed.\n" +
+                "    §c§l✹ Fire:  §7§oFire-imbued weapons will generate a fan of flames that burn one's enemies and empower one's allies. Fire-augmented armors react to being struck, turning the impact of the blow, no matter how weak, into a temporary power boost to Fire-based attacks.\n" +
+                "    §f§l❋ Air:  §7§oAir-imbued weapons trap nearby enemies in a vortex of wind, and blow the victims away when struck. Air-augmented armors will, so long as they are kept close, leach energy from nearby foes to improve the strength of Air-based attacks.";
```
magicus: I think it would be better if you tried to unify the two almost identical blocks of the literal content of the chapters. If you don't want the formatting in one of the copies, get them through a method that strips off the formatting.

magicus: Looking through this again, I think you could check the incoming chat with getFormattedMessage() instead, to get a line that more closely matches the formatted text above.

magicus (changes requested):
The idea is great! The Powder manual is annoying. But please try to keep code duplication down.

src/main/java/com/wynntils/modules/chat/managers/ChatManager.java
P0keDev: Good idea, I'll see what I can do.

HeyZeer0 (approved):

PR #244, opened 2020-08-27
Consumable timer improvements

P0keDev:
Handful of changes:

- Fixed loot chest cooldown disappearing on /class
- Fixed loot chest cooldown only enabling when redirect cooldown messages was enabled (if the implementation here should be changed to separate the two options, let me know)
- Fixed consumables sometimes doubling up in the timer when used
- Added mana pots to the consumable timer
- Added scrolls used by other players to the consumable timer


HeyZeer0:
I'll keep in hold this PR for now because I've to properly analyze it to check if there's a way to better handle it.

HeyZeer0 (approved):

PR #248, opened 2020-09-03
Bank overlay

P0keDev:
Adds several features to the Wynncraft bank.

- Custom page names, editable using the pencil icon on the top right. Supports color using & in place of the formatting symbol.
- Ability to customize the quick access page buttons in the right column, through the config.
- Option to show the page number that each quick access button will take you to.
- Toggleable bank search bar. Highlights items that match the search keyword.
- Optional (defaulted to off) feature that will automatically scroll the bank to the item(s) you are searching for, if something is typed in the search bar.
- All options stored in the "Bank" config entry, under Utilities.

image
image
image
image
image
https://i.gyazo.com/b96bd8f1d1e938466c697f6c2d8f10a3.gif

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/UtilitiesModule.java
+++ b/src/main/java/com/wynntils/modules/utilities/UtilitiesModule.java
@@ -48,6 +48,7 @@ public class UtilitiesModule extends Module {
         registerEvents(new WynnDataOverlay());
         registerEvents(new ItemSpecificationOverlay());
         registerEvents(new ServerUptimeOverlay());
+        registerEvents(new BankOverlay());
 
         // Real overlays
         registerOverlay(new WarTimerOverlay(), Priority.LOWEST);
@@ -83,6 +84,7 @@ public class UtilitiesModule extends Module {
         registerSettings(UtilitiesConfig.Items.class);
         registerSettings(UtilitiesConfig.Wars.class);
         registerSettings(UtilitiesConfig.Market.class);
+        registerSettings(UtilitiesConfig.Bank.class);
         registerSettings(UtilitiesConfig.Identifications.class);
 
         registerSettings(TranslationConfig.class);
--- a/src/main/java/com/wynntils/modules/utilities/configs/UtilitiesConfig.java
+++ b/src/main/java/com/wynntils/modules/utilities/configs/UtilitiesConfig.java
@@ -64,12 +64,6 @@ public class UtilitiesConfig extends SettingsClass {
     @Setting(displayName = "Startup Resource Pack", description = "Should the Wynncraft server resource pack be applied when starting Minecraft?")
     public boolean autoResourceOnLoad = false;
 
-    @Setting(displayName = "Confirmation for Bank Pages", description = "Should Wynntils display a confirmation when buying bank pages?")
-    public boolean addBankConfirmation = true;
-
-    @Setting(displayName = "Open Chat Bank Search", description = "Should the chat open when the bank search asks you to type a response?")
-    public boolean openChatBankSearch = true;
-
     @Setting(displayName = "Change Window Title", description = "Should the mod change the window title to \"Wynncraft\" while on the server?")
     public boolean changeWindowTitle = true;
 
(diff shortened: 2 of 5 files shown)


magicus:
This sounds fantastic!!! I've been thinking about implementing some assistance to managing the bank, but never anything as ambitious as this, and I never got around to it. :(

The code look very clean and straightforward, too.

I have just a few, very minor, comments or questions:

I don't really understand where the quick access buttons are. Is it in the empty space next to the destination page numbers? Perhaps these needs to be made more visible. I also don't really get if this was an existing functionality? I've completely missed it if so.

What is the search highlight color? In the screenshot with "unravel", I can't see any item standing out..? (Edit: oh, now I see the orange. This looks good; I think seeing it pop up as you search will be even better.)

Finally, the "edit" pen symbol does not really feel like it belongs in the Wynncraft style. Perhaps replace it with a book and quill symbol? Or maybe the content team can help with producing a more fitting icon.

P0keDev:
> I don't really understand where the quick access buttons are. Is it in the empty space next to the destination page numbers? Perhaps these needs to be made more visible. I also don't really get if this was an existing functionality? I've completely missed it if so.

Yes, it's the empty space--it will also tell you in its tooltip where it will take you. It is a vanilla Wynn feature, but you're not the only person I know of who has missed it, so maybe adding a better indicator isn't a bad idea.


> Finally, the "edit" pen symbol does not really feel like it belongs in the Wynncraft style. Perhaps replace it with a book and quill symbol? Or maybe the content team can help with producing a more fitting icon.

Definitely a fair critique, I can try a book and quill. I believe there's already a symbol used in the quest book?

magicus:
To continue talking with myself :) I looked more closely at the code for the QA handling. I see my guess about the QA buttons were correct. Maybe you could populate the slots with an item that looks somewhat like a button? I'm somewhat out of ideas right now, but even something like a stone button would probably be an improvement over an empty space.

As a future follow-up, it would be great if the QA button could tie in with the page names and colors. Let's say you have a page 4 named PURPLE + "Build". Then the QA button could show a purple "B".

P0keDev:
> As a future follow-up, it would be great if the QA button could tie in with the page names and colors. Let's say you have a page 4 named PURPLE + "Build". Then the QA button could show a purple "B".

This is another good idea, I'll see what I can do.

HeyZeer0:
> > As a future follow-up, it would be great if the QA button could tie in with the page names and colors. Let's say you have a page 4 named PURPLE + "Build". Then the QA button could show a purple "B".
> 
> This is another good idea, I'll see what I can do.

You could as well draw the text in the middle of the slot without shadow and with a gray accent. Would look good too ^^

magicus:
There is surprisingly no "write" symbol in the Quest Book what I can find (I thought there was one, too). There is however a "pen" symbol in the new character selection texture. It's still a pen and I honestly don't think it fits there either, but it's colored and I think it at least look a tad better.

HeyZeer0:
> There is surprisingly no "write" symbol in the Quest Book what I can find (I thought there was one, too). There is however a "pen" symbol in the new character selection texture. It's still a pen and I honestly don't think it fits there either, but it's colored and I think it at least look a tad better.

Might just removing the edit button is enough. You just have to make the title editable by clicking on it

magicus:
@HeyZeer0 I see your point, but that severely lacks in discoverability. How will a user know that possibility exists? If given that option only, I prefer the current pen. ;-)

P0keDev:
I did some digging and found an unused texture in the Wynn pack that I think works for the quick access buttons: 
image

I'm not entirely sure if my method of checking whether the texture is present (in the unlikely case that the user doesn't have the texture pack) is the cleanest method, but as far as I can tell the options are pretty limited. Will get to work on the other improvements soon.

P0keDev:
Now using the pencil from the class selector!
image

magicus:
This looks really great! I'm happy with the fix as it is right now.

magicus (approved):

magicus:
@P0keDev I had an idea. Too late for this PR but certainly possible for a follow-up. Why don't you add the custom bank page name (if one exists) as a hover tooltip for the QA buttons? That should probably be quite easy to implement, and give you plenty of space to fill in the full name with colors and. everything, instead of trying to abbreviate it.

P0keDev:
> @P0keDev I had an idea. Too late for this PR but certainly possible for a follow-up. Why don't you add the custom bank page name (if one exists) as a hover tooltip for the QA buttons? That should probably be quite easy to implement, and give you plenty of space to fill in the full name with colors and. everything, instead of trying to abbreviate it.

I added this already, actually.

magicus:
@P0keDev Great minds think alike! :-) Fantastic work, thank you.

PR #250, opened 2020-09-08
Overhauled ears to support transparency and proper mirroring

Cobular:
Hey! This PR improves the way fox ears work. Now, ears support transparency, and are mirrored better, which means you can now make non-square ears. 
To get transparency to work, we had to pull skin data from two separate, disconnected areas on the skin file. To do this, @liamgolly had to extend ModelBox and change how it pulled textures from the skin. This was then wrapped with ExtendedModelRenderer, which takes in 2 pairs of texture offsets defining the upper left corner of each area and correctly passes this information on to ExtendedModelBox. This was then used to handle the rendering of the ears in LayerFoxEars. All of the rest of the functions of these classes should be the same, however the multiple area texturing has not been tested with any shape other than a 6x6x1 rectangular prism, so be sure to test extensively before using these classes with anything else. Additionally, this system is not backwards compatible with the old ears system, since there was no way we could find to get transparency in those pixels. However, I do not believe this will be such a problem, as this system has been improved significantly, and skins can be changed to this system in just a matter of minutes using paint.net or something similar.
Proper mirroring was a lot simpler, just creating two ears instead of duplicating one, and brute-forcing the positioning. I'm aware it isn't totally 100% symmetrical right now, however I think that can help make them look better, and the difference is hard to notice in the first place. I am sure there is a better way to do this, however I don't really know what that would be since I don't really understand openGL very well. Everything seems to have a different origin, so I just went with the easier option at the time. 

Finally, here's some examples for how the new ears work:
* Ears are a 6x6x1 rectangular prism, which can be represented by this cubemap, with blue/purple as the front and red facing backwards:
image
* This unwraps to this shape on the skin image: 
image Notably, yellow moves up from the bottom of the blue/purple to the top of the red. This unusual separation of skin elements is required as these are the only currently known positions where transparency is allowed on a skin file. This example uses a pre-1.8 skin file (as it is 64x32), but the same area also works on a post-1.8 skin file (the 64x64 style that allows for sleeves and suits). It is very possible that there are more valid transparent areas on a post-1.8 file, but we have not looked into them yet. 
---------
* As a further example, here is how to create these ears: 
image

* They can be represented by this cubemap: 
image
* Which is then represented by this skin file: 
image
---------
Anyway, that's basically everything to this PR. Thanks for reviewing, and hopefully this will be able to be included!


oohwooh (approved):
Looks good to me, i guess

src/main/java/com/wynntils/modules/cosmetics/layers/models/ExtendedModelBox.java
```diff
@@ -1,6 +1,140 @@
 package com.wynntils.modules.cosmetics.layers.models;
 
-import net.minecraft.client.model.ModelBox;
+import net.minecraft.client.renderer.BufferBuilder;
+import net.minecraftforge.fml.relauncher.Side;
+import net.minecraftforge.fml.relauncher.SideOnly;
+import net.minecraft.client.model.*;
```
oohwooh: Is this really the best idea?

src/main/java/com/wynntils/modules/cosmetics/layers/models/ExtendedModelRenderer.java
```diff
@@ -1,19 +1,69 @@
...
 import net.minecraft.client.model.ModelBase;
-import net.minecraft.client.model.ModelRenderer;
+import net.minecraft.client.model.ModelBox;
 import net.minecraft.client.renderer.BufferBuilder;
 import net.minecraft.client.renderer.GLAllocation;
 import net.minecraft.client.renderer.GlStateManager;
 import net.minecraft.client.renderer.Tessellator;
 import net.minecraftforge.fml.relauncher.Side;
 import net.minecraftforge.fml.relauncher.SideOnly;
+import net.minecraft.client.model.ModelRenderer;
```
oohwooh: This is not alphebetized

oohwooh: alphabetized *

oohwooh:
This looks great! Good feature, very excited. Any update on this @maintainers????

oohwooh:
Whoops! Forgot to submit my review. Good job jake! 💖

HeyZeer0 (approved):

HeyZeer0:
> Whoops! Forgot to submit my review. Good job jake! 💖

That's so sweet, thanks ❤️ 

------
Sorry for the delay in reviewing this PR @JakeCover , I had a lot to do recently. Thanks for your amazing contribution ^^

oohwooh:
This is great thank you for merging!!

sent from my iPhone

PR #263, opened 2020-10-17
Unicode font toggle

P0keDev:
As requested, added a unicode toggle for Wynntils HUDs instead of simply following the respective Minecraft setting.


HeyZeer0:
I don't get why is this a thing, isn't already there a Minecraft option for it? :thinking:

P0keDev:
> I don't get why is this a thing, isn't already there a Minecraft option for it? 🤔

There is, but it seems some people would prefer to use the default font for Wynntils and unicode for vanilla Minecraft features, so I wanted to accommodate for that.

HeyZeer0 (approved):

PR #266, opened 2020-10-21
Allow accessories to be shift-clicked on/off

P0keDev:
Adds a toggleable feature to let accessories be shift clicked as if they were armor pieces.

Also includes some code to determine what item type (ring, wand, etc.) a given itemstack is.

### DIFF
--- a/src/main/java/com/wynntils/core/utils/ItemUtils.java
+++ b/src/main/java/com/wynntils/core/utils/ItemUtils.java
@@ -104,6 +104,12 @@ public class ItemUtils {
         return toReturn.toString();
     }
     
+    /**
+     * Determines the equipment type of the given item.
+     * 
+     * @param item
+     * @return The ItemType of the item, or null if invalid or not an equipment piece
+     */
     public static ItemType getItemType(ItemStack item) {
         for (Entry<ItemType, String[]> e : WebManager.getMaterialTypes().entrySet()) {
             for (String id : e.getValue()) {
--- a/src/main/java/com/wynntils/core/utils/ItemUtils.java
+++ b/src/main/java/com/wynntils/core/utils/ItemUtils.java
@@ -5,6 +5,9 @@
 package com.wynntils.core.utils;
 
 import com.wynntils.core.utils.reference.EmeraldSymbols;
+import com.wynntils.webapi.WebManager;
+import com.wynntils.webapi.profiles.item.enums.ItemType;
+
 import net.minecraft.init.Blocks;
 import net.minecraft.init.Items;
 import net.minecraft.inventory.IInventory;
@@ -18,6 +21,7 @@ import net.minecraft.util.text.TextFormatting;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map.Entry;
 
 public class ItemUtils {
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
```diff
@@ -563,11 +564,101 @@ public void keyPressOnHorse(GuiOverlapEvent.HorseOverlap.KeyTyped e) {
...
+        
+        if (UtilitiesConfig.INSTANCE.shiftClickAccessories && e.getGui().isShiftKeyDown() && e.getGui().getSlotUnderMouse() != null && ModCore.mc().player.inventory.getItemStack().isEmpty() && e.getGui().getSlotUnderMouse().inventory == ModCore.mc().player.inventory) {
+            if (e.getSlotId() >= 9 && e.getSlotId() <= 12) { // taking off accessory
+                // check if hotbar has open slot; if so, no action required
+                for (int i = 36; i < 45; i++) {
+                    if (!e.getGui().inventorySlots.getSlot(i).getHasStack()) return;
+                }
+                
+                // move accessory into inventory
+                e.setCanceled(true);
```
magicus: Just like when equipping items from the inventory,  the setCancel should come only after you are sure to have found an empty slot.  I think..?

magicus (changes requested):

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
P0keDev: As long as the hotbar is full (which it will be if that code is reached), the shift-click won't do anything anyway so I cancel it regardless.

magicus: Okay, I see your point. It will work just fine. I still think it would be clearer for the next person reading this code if the cancelling is done in tandem with setting an accessoryDestinationSlot, but if you don't want to change that, I'm fine with the code as-is.

magicus (approved):

src/main/java/com/wynntils/core/utils/ItemUtils.java
```diff
@@ -99,6 +103,27 @@ public static String getStringLore(ItemStack is) {
         }
         return toReturn.toString();
     }
+    
+    public static ItemType getItemType(ItemStack item) {
```
HeyZeer0, line 113: Please add documentation on what this method does, just for future reference

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
```diff
@@ -563,11 +564,101 @@ public void keyPressOnHorse(GuiOverlapEvent.HorseOverlap.KeyTyped e) {
...
+                
+            // pick up accessory
+            CPacketClickWindow packet = new CPacketClickWindow(e.getGui().inventorySlots.windowId, e.getSlotId(), 0, ClickType.PICKUP, e.getSlotIn().getStack(), e.getGui().inventorySlots.getNextTransactionID(ModCore.mc().player.inventory));
+            ModCore.mc().getConnection().sendPacket(packet);
+        }
+    }
+    
+    @SubscribeEvent
+    public void handleAccessoryMovement(TickEvent.ClientTickEvent e) {
+        if (!Reference.onWorld || accessoryDestinationSlot == -1) return;
```
HeyZeer0, line 643: check if the tick phase is on END otherwise, the event will be triggered twice which could cause some problems

HeyZeer0 (approved):
LGTM once everything was addressed

magicus (approved):
I'm happy. Thanks for improving readability! (And for implementing this feature!)

PR #267, opened 2020-11-02
Middle-click to view player's gear

P0keDev:
While looking at a player, middle-click (changeable under control settings) to open a GUI and view their armor and held item.

[Confirmation of legality](https://cdn.discordapp.com/attachments/738272813496139797/744670974581932083/image0.jpg)

image
image


src/main/java/com/wynntils/modules/utilities/enums/InternalIdentification.java
```diff
@@ -0,0 +1,71 @@
+package com.wynntils.modules.utilities.enums;
+
+public enum InternalIdentification {
```
HeyZeer0: needs to be moved to athena

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
```diff
@@ -313,7 +314,7 @@ public void onMythicFound(PacketEvent<SPacketWindowItems> e) {
     @SubscribeEvent
     public void onSlotSet(PacketEvent<SPacketSetSlot> e) {
         if (Minecraft.getMinecraft().currentScreen == null) return;
-        if (!(Minecraft.getMinecraft().currentScreen instanceof SkillPointLoadoutUI)) return;
+        if (!(Minecraft.getMinecraft().currentScreen instanceof SkillPointLoadoutUI || Minecraft.getMinecraft().currentScreen instanceof GearViewerUI)) return;
```
HeyZeer0: Create a new type of container class, call it something like FakeContainer (ofcourse it will extend Container), and extends over SkillPointLoadoutUI and GearViewerUI, after that just do ``currentScreen instanceof FakeContainer``. 
Just for future easier scalability

src/main/java/com/wynntils/modules/utilities/instances/ContainerGearViewer.java
```diff
@@ -0,0 +1,55 @@
...
+    public boolean canInteractWith(EntityPlayer playerIn) {
+        return true;
+    }
+
+    @Override
+    public void onContainerClosed(EntityPlayer playerIn) {
+        super.onContainerClosed(playerIn);
+        this.inventory.closeInventory(playerIn);
+    }
+
```
HeyZeer0, line 53: remove this space

src/main/java/com/wynntils/modules/utilities/managers/KeyManager.java
```diff
@@ -81,8 +82,10 @@ public static void registerKeys() {
         });
 
         stopwatchKey = CoreModule.getModule().registerKeyBinding("Start/Stop Stopwatch", Keyboard.KEY_NUMPAD5, "Wynntils", true, StopWatchOverlay::start);
-        
+
         itemScreenshotKey = CoreModule.getModule().registerKeyBinding("Screenshot Current Item", Keyboard.KEY_F4, "Wynntils", true, () -> {});
+      
+        CoreModule.getModule().registerKeyBinding("View Player's Gear", -98, "Wynntils", true, GearViewerUI::openGearViewer);
```
HeyZeer0, line 89: Are you sure there isn't a key enum for mouse buttons? (easier for visualizing in the future)

HeyZeer0 (changes requested):
nice work on this!

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
P0keDev: done

src/main/java/com/wynntils/modules/utilities/instances/ContainerGearViewer.java
P0keDev: fixed

src/main/java/com/wynntils/modules/utilities/managers/KeyManager.java
P0keDev: As far as I can tell Minecraft hard-codes the mouse codes as 100 - mouse button number, with no enum. For now I've added a comment to clarify.

coehlrich: Fields with mouse button numbers was added in lwjgl3 from what i could find and Minecraft started using them directly instead of 100 - mouse button number.

I just send stuff like this when i find something that mojang fixed in newer versions (or in this case lwjgl fixing).

HeyZeer0 (approved):
LGTM once the Athena implementation is done.

PR #268, opened 2020-11-06
Item screenshotter

P0keDev:
While hovering an item, press the hotkey (F4 by default) to copy only the item's tooltip to the clipboard. Removes any Wynncraft item lore to focus on the item's stats. 

image
image


HeyZeer0 (approved):
LGTM

PR #270, opened 2020-11-11
Added 2 keybinds

TrapinchO:
1) Open item guide -> NUMPAD 1
2) Open HUD configuration -> NONE

EDIT: It works

### DIFF
--- a/src/main/java/com/wynntils/modules/questbook/QuestBookModule.java
+++ b/src/main/java/com/wynntils/modules/questbook/QuestBookModule.java
@@ -28,7 +28,7 @@ public class QuestBookModule extends Module {
 
         registerKeyBinding("Open Quest Book", Keyboard.KEY_K, "Wynntils", true, () -> QuestBookPages.QUESTS.getPage().open(true));
         registerKeyBinding("Open Discoveries", Keyboard.KEY_U, "Wynntils", true, () -> QuestBookPages.DISCOVERIES.getPage().open(true));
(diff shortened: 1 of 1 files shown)


src/main/java/com/wynntils/modules/questbook/QuestBookModule.java
```diff
@@ -28,6 +28,8 @@ public void onEnable() {
 
         registerKeyBinding("Open Quest Book", Keyboard.KEY_K, "Wynntils", true, () -> QuestBookPages.QUESTS.getPage().open(true));
         registerKeyBinding("Open Discoveries", Keyboard.KEY_U, "Wynntils", true, () -> QuestBookPages.DISCOVERIES.getPage().open(true));
+        registerKeyBinding("Open Item Guide", Keyboard.KEY_NUMPAD1, "Wynntils", true, () -> QuestBookPages.ITEMGUIDE.getPage().open(true));
```
HeyZeer0: We should not use Numpad for a frequent usable feature

HeyZeer0 (changes requested):

src/main/java/com/wynntils/modules/questbook/QuestBookModule.java
TrapinchO: What key do you propose then? We are running out of them

HeyZeer0 (approved):

PR #271, opened 2020-11-11
Fixed gathering cooldown not working during prof speed bombs

TrapinchO:
This should fix the bug. Again, I could not test it ingame (the code is copied and the change is minimal though)


P0keDev (changes requested):
A much better solution, instead of using redundant code, would be to use a regex expression like the chest cooldown timer does.

TrapinchO:
Due to my modding skills, I only copied code that (presumably) did the same thing except outside prof speed bombs.

magicus (approved):
Looks good now.

P0keDev (approved):

HeyZeer0 (approved):

PR #273, opened 2020-11-13
Added %chestcount% and %maxchestcount% Info Overlays

SzilBalazs:
%chestcount%: Opened chests (Reset when load a lootrun)
%maxchestcount%: Loaded chests


src/main/java/com/wynntils/modules/utilities/managers/LootChestManager.java
```diff
@@ -0,0 +1,13 @@
+package com.wynntils.modules.utilities.managers;
+
+public class LootChestManager {
```
HeyZeer0: should be done inside the loot run container for consistency

HeyZeer0 (changes requested):

src/main/java/com/wynntils/modules/utilities/managers/LootChestManager.java
SzilBalazs: Inside com.wynntils.modules.map.instances.LootRunPath or inside com.wynntils.modules.map.manager.LootRunManager?

PR #274, opened 2020-11-14
Prompt to update before quitting

P0keDev:
If enabled, opens the update screen if an update is available when clicking "quit game" on the main menu.


HeyZeer0:
Could you try copying the update file on a shutdown hook?
If that works, we could always download the update when the user agrees but just apply it on the shutdown which will be 100% more efficient and cleaner, maybe add a button like "update on shutdown?"

You can add a shutdown hook at
```
 Runtime#addShutdownHook
```

Hope that works :pray:

magicus:
Great news! I've been thinking much about doing something like this. My take would have been to show at the start "New upgrade available", and add a button to "Upgrade when quitting Minecraft", similar to what heyzeer0 suggests. I agree that downloading it straight ahead is probably a good idea, and just wait with applying the update until later.

PR #279, opened 2020-12-18
More FOV scaling functions

phantamanta44:
I play with max FOV and I like having dynamic FOV enabled so that I can gauge how much movement speed I have. However, I often run into the problem where I get too much move speed (e.g. with escape's speed effect) and my FOV becomes so distorted as to make the game unplayable, and so I end up having to manually toggle dyanmic FOV off until the effect wears off.

To remedy this, this PR replaces the "disable dynamic FOV" config option with a selection between several different FOV scaling functions. The ones I've implemented are:
* Vanilla: the no-op option that leaves the FOV alone
* Arctangent: uses an arctan curve to tone down FOV scaling at very high speeds
* Sprint-only: ignores FOV changes from everything except sprinting; equivalent to the current FOV setting
* None: the FOV will not change at all, except by manually modifying the FOV slider


HeyZeer0:
Really good PR, thanks! =)

PR #280, opened 2020-12-18
Better item level/tier overlays

phantamanta44:
Currently, item level overlays are implemented with this hacky thing in `RarityColorOverlay` that just modifies stack sizes for identifiable items when the ctrl key is held. Not only is this a rather ugly implementation, but it also precludes the ability to show an overlay over stackable items, such as ingredients or powders, as it relies on the assumption that the item is unstackable.

This PR remedies this by tossing out the entire stack size trick entirely. Instead, it creates a class `WynnRenderItem extends RenderItem`, then uses reflection to inject an instance of this into `Minecraft#renderItem`. This means that all `renderItemOverlayIntoGUI` calls will go through `WynnRenderItem`, giving us the opportunity to intercept them and do our own rendering instead.

The solution for specifying the overlay text that I came up with is to create a new event `RenderEvent.DrawItemOverlay` which any interested party can listen for and use to modify the text that renders in the item overlay. Then, I've created a handler class `ItemLevelOverlay` in the utilities module that replicates the original item level overlay behaviour using the new event-based system. The exact features are:
* Combat level overlays for identified items, exactly as in the old implementation
* Combat level overlays for unidentified items, with a config option for toggling between displaying the entire level range or display only the average level
* Crafting level overlays for crafting ingredients
* Powder tier overlays, with a config option for toggling between roman and arabic numerals

### DIFF
--- a/src/main/java/com/wynntils/core/framework/rendering/WynnRenderItem.java
+++ b/src/main/java/com/wynntils/core/framework/rendering/WynnRenderItem.java
@@ -35,8 +35,6 @@ public class WynnRenderItem extends RenderItem {
 
     private static final int GUI_OVERLAY_WIDTH_THRESH = 16;
 
-    private final ScreenRenderer renderer = new ScreenRenderer();
-
     private WynnRenderItem(RenderItem parent, TextureManager texMan, ItemColors itemCols) {
         super(texMan, parent.getItemModelMesher().getModelManager(), itemCols);
         ReflectionFields.RenderItem_itemModelMesher.setValue(this, ReflectionFields.RenderItem_itemModelMesher.getValue(parent));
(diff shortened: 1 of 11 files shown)


src/main/java/com/wynntils/core/utils/ItemUtils.java
```diff
@@ -184,4 +188,45 @@ public static String describeMoney(int total) {
...
+
+    public static CombatLevel getLevel(String lore) {
+        Matcher m = COMBAT_LEVEL_PATTERN.matcher(lore);
+        if (m.find()) return new CombatLevel(Integer.parseInt(m.group(1)));
+        m = COMBAT_LEVEL_RANGE_PATTERN.matcher(lore);
+        if (m.find()) return new CombatLevel(Integer.parseInt(m.group(1)), Integer.parseInt(m.group(2)));
+        return null;
+    }
+
+    public static class CombatLevel {
```
HeyZeer0: We avoid using this kind of nested classes. When they are used by the same class it's fine, but when they are useful for other ones, we create proper files for them.

Move it to the following package: ` com.wynntils.core.utils.objects`

src/main/java/com/wynntils/core/framework/rendering/WynnRenderItem.java
```diff
@@ -0,0 +1,73 @@
...
+                GlStateManager.translate(0f, ScreenRenderer.fontRenderer.FONT_HEIGHT * (1f - scaleRatio) / 2f, 0f);
+                GlStateManager.scale(scaleRatio, scaleRatio, 1f);
+            }
+            ScreenRenderer.fontRenderer.drawString(event.getOverlayText(), 0, 0, event.getOverlayTextColor(),
+                    SmartFontRenderer.TextAlignment.RIGHT_LEFT, SmartFontRenderer.TextShadow.NORMAL);
+            GlStateManager.popMatrix();
+            GlStateManager.enableLighting();
+            GlStateManager.enableDepth();
+            GlStateManager.enableBlend();
+        } else {
```
HeyZeer0: This else is not necessary! Replace it with a return for better code consistency.

src/main/java/com/wynntils/modules/utilities/overlays/inventories/ItemLevelOverlay.java
```diff
@@ -0,0 +1,85 @@
...
+        Item item = stack.getItem();
+        String name = stack.getDisplayName();
+
+        // powder tier
+        if (item == Items.DYE || item == Items.GUNPOWDER || item == Items.CLAY_BALL || item == Items.SUGAR) {
+            Matcher powderMatcher = POWDER_NAME_PATTERN.matcher(StringUtils.normalizeBadString(name));
+            if (powderMatcher.find()) {
+                if (UtilitiesConfig.Items.INSTANCE.romanNumeralPowderTier) {
+                    event.setOverlayText(powderMatcher.group(1));
+                } else {
```
HeyZeer0: This else is not necessary! Replace it with a return for better code consistency.

HeyZeer0 (changes requested):
Awesome work!
It's great to see that you perfectly followed our code patterns of creating events for future scalability 😄 

Just these small changes and it LGTM!

HeyZeer0 (approved):

magicus:
@phantamanta44 Nice! :) 🎉

PR #282, opened 2020-12-22
Improvements to level overlays

phantamanta44:
Some small improvements to level overlays (and level arcs, by extension):
* Level overlays can now be disabled outside GUIs using a config option, which renders the corresponding keybind inert when a GUI is not open.
* Level overlays (and also level arcs) now display for most items related to professions, including raw materials and gathering tools.


quick007:
why is only one verified 😳

phantamanta44:
the green check mark doesn't mean it's a verified commit; it means that the github status check was run on that commit and that it passed

HeyZeer0 (approved):

PR #285, opened 2021-01-01
Rewritten Server Restart Timer

TrapinchO:
1) rewritten the restart timer because it didn't seem to work
2) Changed 2 words in the settings so it is easier to find

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
+++ b/src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
@@ -387,29 +387,7 @@ public class OverlayEvents implements Listener {
                 return;
             }
         }
-        /*
-        if (OverlayConfig.GameUpdate.RedirectSystemMessages.INSTANCE.redirectServer) {
-            if (messageText.matches("The server is restarting in \\d+ (seconds?|minutes?)\\.")) {
-                String[] res = messageText.split(" ");
-                GameUpdateOverlay.queueMessage(TextFormatting.DARK_RED + res[5] + " " + res[6].replace(".", "") + " until server restart");
-                e.setCanceled(true);
 
-                if (OverlayConfig.ConsumableTimer.INSTANCE.showServerRestart) {
-                    try {
-                        int seconds = Integer.parseInt(res[5]);
-                        if (res[6].startsWith("minute")) {
-                            seconds = seconds * 60;
-                        }
-                        ConsumableTimerOverlay.addBasicTimer("Server restart", seconds);
-                    } catch (NumberFormatException ignored) {
-                        // ignore
-                    }
-                }
-
-                return;
-            }
-        }
-        */
         if (OverlayConfig.GameUpdate.RedirectSystemMessages.INSTANCE.redirectQuest) {
             if (messageText.startsWith("[Quest Book Updated]")) {
                 GameUpdateOverlay.queueMessage(TextFormatting.GRAY + "Quest book updated.");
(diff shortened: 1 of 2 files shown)


src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -592,6 +573,23 @@ else if (messageText.matches("[a-zA-Z0-9_ ]{1,19} has given you 20% resistance a
             }
             return;
         }
+
+        // Server restart message handling
+        Matcher matcher_restart = SERVER_RESTART_PATTERN.matcher(messageText);
```
HighCrit: Snake case and camel case for variable names on the same line 🤢

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -592,6 +573,23 @@ else if (messageText.matches("[a-zA-Z0-9_ ]{1,19} has given you 20% resistance a
...
+        if (matcher_restart.find()) {
+            if (OverlayConfig.ConsumableTimer.INSTANCE.showServerRestart) { // if you want the timer
+                int seconds = Integer.parseInt(matcher_restart.group(1));
+
+                if (matcher_restart.group(2).equals("minute") || matcher_restart.group(2).equals("minutes")) { // if it is in minutes
```
HighCrit: Check for minutes first so you'll leave the if statement earlier more often

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: What? It is the same as the rest (or more exactly like the 2 above)

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: How exactly?

Like do mins and divide by 60 when seconds?

HighCrit: Swap "minute" and "minutes"

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
HighCrit: Oh I see. Who did that?

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: ok, will do

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: I have no idea, I am known for not doing original code

HighCrit: It was you 🕵️ https://github.com/Wynntils/Wynntils/commit/a7006c9fa844245222d0098f07c00abfbc064b68

TrapinchO: lol

ok, should I rename it to "matcherRestart" or something?

HighCrit: restartMatcher would be better imo

TrapinchO: ok, I will rename the other ones as well

coehlrich (approved):

PR #287, opened 2021-01-02
Rewritten Blacksmith redirect message

TrapinchO:
I can change the formatting if you want

@HeyZeer0 please notice it

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
+++ b/src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
@@ -417,7 +417,6 @@ public class OverlayEvents implements Listener {
                 EnumMap<ItemTier, Integer> itemCounts = new EnumMap(ItemTier.class); // counter for sold/scrapped items
 
                 String[] res = formattedText.split("§");
-                int[] itemCount = {0, 0, 0, 0, 0, 0, 0, 0}; // normal, unique, rare, set, legendary, fabled, mythic, crafted
 
                 for (String s : res) {
                     if (s.equals("dYou sold me: ") //non-item founds
@@ -429,7 +428,11 @@ public class OverlayEvents implements Listener {
                         continue;
                     }
                     else if (s.matches("e\\d+")) {
-                        int total = IntStream.of(itemCount).sum(); // sum of the items
+                        //int total = IntStream.of(itemCount).sum(); // sum of the items
+                        int total = 0;
+                        for (ItemTier tier : ItemTier.values()) {
+                            total += itemCounts.getOrDefault(tier, 0);
+                        }
 
                         StringBuilder message = new StringBuilder(); // creates the message
                         for (ItemTier tier : ItemTier.values()) {
@@ -453,22 +456,14 @@ public class OverlayEvents implements Listener {
                         e.setCanceled(true); // remove the chat message
                     }
                     // item counter
-                    else if (s.startsWith("f")) { itemCount[0]++;
-                        itemCounts.put(ItemTier.NORMAL, itemCounts.getOrDefault(ItemTier.NORMAL, 0) + 1); } // normal
-                    else if (s.startsWith("e")) { itemCount[1]++;
-                        itemCounts.put(ItemTier.UNIQUE, itemCounts.getOrDefault(ItemTier.UNIQUE, 0) + 1); } // unique
-                    else if (s.startsWith("d")) { itemCount[2]++;
-                        itemCounts.put(ItemTier.RARE, itemCounts.getOrDefault(ItemTier.RARE, 0) + 1); } // rare and ingredients
-                    else if (s.startsWith("a")) { itemCount[3]++;
-                        itemCounts.put(ItemTier.SET, itemCounts.getOrDefault(ItemTier.SET, 0) + 1); } // set
-                    else if (s.startsWith("b")) { itemCount[4]++;
-                        itemCounts.put(ItemTier.LEGENDARY, itemCounts.getOrDefault(ItemTier.LEGENDARY, 0) + 1); } // legendary
-                    else if (s.startsWith("c")) { itemCount[5]++;
-                        itemCounts.put(ItemTier.FABLED, itemCounts.getOrDefault(ItemTier.FABLED, 0) + 1); } // fabled
-                    else if (s.startsWith("5")) { itemCount[6]++;
-                        itemCounts.put(ItemTier.MYTHIC, itemCounts.getOrDefault(ItemTier.MYTHIC, 0) + 1); } // mythic
-                    else if (s.startsWith("3")) { itemCount[7]++;
-                        itemCounts.put(ItemTier.CRAFTED, itemCounts.getOrDefault(ItemTier.CRAFTED, 0) + 1); } // crafted
+                    else if (s.startsWith("f")) { itemCounts.put(ItemTier.NORMAL, itemCounts.getOrDefault(ItemTier.NORMAL, 0) + 1); } // normal
+                    else if (s.startsWith("e")) { itemCounts.put(ItemTier.UNIQUE, itemCounts.getOrDefault(ItemTier.UNIQUE, 0) + 1); } // unique
+                    else if (s.startsWith("d")) { itemCounts.put(ItemTier.RARE, itemCounts.getOrDefault(ItemTier.RARE, 0) + 1); } // rare and ingredients
+                    else if (s.startsWith("a")) { itemCounts.put(ItemTier.SET, itemCounts.getOrDefault(ItemTier.SET, 0) + 1); } // set
+                    else if (s.startsWith("b")) { itemCounts.put(ItemTier.LEGENDARY, itemCounts.getOrDefault(ItemTier.LEGENDARY, 0) + 1); } // legendary
+                    else if (s.startsWith("c")) { itemCounts.put(ItemTier.FABLED, itemCounts.getOrDefault(ItemTier.FABLED, 0) + 1); } // fabled
+                    else if (s.startsWith("5")) { itemCounts.put(ItemTier.MYTHIC, itemCounts.getOrDefault(ItemTier.MYTHIC, 0) + 1); } // mythic
+                    else if (s.startsWith("3")) { itemCounts.put(ItemTier.CRAFTED, itemCounts.getOrDefault(ItemTier.CRAFTED, 0) + 1); } // crafted
                 }
                 return;
             } else if (messageText.equals("Blacksmith: I can't buy that item! I only accept weapons, accessories, potions, armour, ingredients, resources, and crafted items.")) {
--- a/src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
+++ b/src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
@@ -412,49 +412,68 @@ public class OverlayEvents implements Listener {
                 // TODO jesus christ rewrite this thing
                 boolean sold = formattedText.split(" ")[2].equals("sold");
                 String[] res = formattedText.split("§");
-                int countCommon = 0;
-                int countUnique = 0;
-                int countRare = 0;
-                int countSet = 0;
-                int countLegendary = 0;
-                int countFabled = 0;
-                int countMythic = 0;
-                int countCrafted = 0;
+                int[] countTest = {0, 0, 0, 0, 0, 0, 0, 0};
                 int total = 0;
                 for (String s : res) {
                     if (s.startsWith("f")) {
-                        countCommon++;
+                        countTest[0]++;
                         total++;
                     } else if (s.startsWith("c")) {
-                        countFabled++;
+                        countTest[5]++;
                         total++;
                     } else if (s.startsWith("b")) {
-                        countLegendary++;
+                        countTest[4]++;
                         total++;
                     } else if (s.startsWith("5") && !s.equals("5Blacksmith: ")) {
-                        countMythic++;
+                        countTest[6]++;
                         total++;
                     } else if (s.startsWith("d") && !s.equals("dYou sold me: ") && !s.equals("dYou scrapped: ") && !s.equals("d, ") && !s.equals("d and ") && !s.equals("d for a total of ")) {
-                        countRare++;
+                        countTest[2]++;
                         total++;
                     } else if (s.startsWith("a")) {
-                        countSet++;
+                        countTest[3]++;
                         total++;
                     } else if (s.startsWith("3")) {
-                        countCrafted++;
+                        countTest[7]++;
+
                         total++;
                     } else if (s.startsWith("e")) {
                         if (s.matches("e\\d+")) {
-                            String message;
+                            String message2 = "";
                             if (sold) {
-                                message = TextFormatting.LIGHT_PURPLE + "Sold " + total + " (" + TextFormatting.WHITE + countCommon + TextFormatting.LIGHT_PURPLE + "/" + TextFormatting.YELLOW + countUnique + TextFormatting.LIGHT_PURPLE + "/" + countRare + "/" + TextFormatting.GREEN + countSet + TextFormatting.LIGHT_PURPLE + "/" + TextFormatting.AQUA + countLegendary + TextFormatting.LIGHT_PURPLE + "/" + TextFormatting.RED + countFabled + TextFormatting.LIGHT_PURPLE + "/" + TextFormatting.DARK_PURPLE + countMythic + TextFormatting.LIGHT_PURPLE + "/" + TextFormatting.DARK_AQUA + countCrafted + TextFormatting.LIGHT_PURPLE + ") item(s) for " + TextFormatting.GREEN + s.replace("e", "") + (char) 0xB2 + TextFormatting.LIGHT_PURPLE + ".";
+                                int sum = 0;
+                                for (int i : countTest) {
+                                    sum += i;
+                                }
+
+                                message2 = TextFormatting.LIGHT_PURPLE + "Sold " + sum
+                                        + " (" + TextFormatting.WHITE + countTest[0] + TextFormatting.LIGHT_PURPLE
+                                        + "/" + TextFormatting.YELLOW + countTest[1] + TextFormatting.LIGHT_PURPLE
+                                        + "/" + countTest[2]
+                                        + "/" + TextFormatting.GREEN + countTest[3] + TextFormatting.LIGHT_PURPLE
+                                        + "/" + TextFormatting.AQUA + countTest[4] + TextFormatting.LIGHT_PURPLE
+                                        + "/" + TextFormatting.RED + countTest[5] + TextFormatting.LIGHT_PURPLE
+                                        + "/" + TextFormatting.DARK_PURPLE + countTest[6] + TextFormatting.LIGHT_PURPLE
+                                        + "/" + TextFormatting.DARK_AQUA + countTest[7] + TextFormatting.LIGHT_PURPLE
+                                        + ") item(s) for " + TextFormatting.GREEN + s.replace("e", "")
+                                        + (char) 0xB2 + TextFormatting.LIGHT_PURPLE + ".";
                             } else {
-                                message = TextFormatting.LIGHT_PURPLE + "Scrapped " + total + " (" + TextFormatting.WHITE + countCommon + TextFormatting.LIGHT_PURPLE + "/" + TextFormatting.YELLOW + countUnique + TextFormatting.LIGHT_PURPLE + "/" + countRare + "/" + TextFormatting.GREEN + countSet + TextFormatting.LIGHT_PURPLE + "/" + TextFormatting.AQUA + countLegendary + TextFormatting.LIGHT_PURPLE + "/" + TextFormatting.RED + countFabled + TextFormatting.LIGHT_PURPLE + "/" + TextFormatting.DARK_PURPLE + countMythic + TextFormatting.LIGHT_PURPLE + "/" + TextFormatting.DARK_AQUA + countCrafted + TextFormatting.LIGHT_PURPLE + ") item(s) for " + TextFormatting.YELLOW + s.replace("e", "") + " scrap" + TextFormatting.LIGHT_PURPLE + ".";
+                                message2 = TextFormatting.LIGHT_PURPLE + "Scrapped " + total
+                                        + " (" + TextFormatting.WHITE + countTest[0] + TextFormatting.LIGHT_PURPLE
+                                        + "/" + TextFormatting.YELLOW + countTest[1] + TextFormatting.LIGHT_PURPLE
+                                        + "/" + countTest[2]
+                                        + "/" + TextFormatting.GREEN + countTest[3] + TextFormatting.LIGHT_PURPLE
+                                        + "/" + TextFormatting.AQUA + countTest[4] + TextFormatting.LIGHT_PURPLE
+                                        + "/" + TextFormatting.RED + countTest[5] + TextFormatting.LIGHT_PURPLE
+                                        + "/" + TextFormatting.DARK_PURPLE + countTest[6] + TextFormatting.LIGHT_PURPLE
+                                        + "/" + TextFormatting.DARK_AQUA + countTest[7] + TextFormatting.LIGHT_PURPLE
+                                        + ") item(s) for " + TextFormatting.YELLOW + s.replace("e", "")
+                                        + " scrap" + TextFormatting.LIGHT_PURPLE + ".";
                             }
-                            GameUpdateOverlay.queueMessage(message);
+                            GameUpdateOverlay.queueMessage(message2);
                             e.setCanceled(true);
                         } else {
-                            countUnique++;
+                            countTest[1]++;
                             total++;
                         }
                     }
--- a/src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
+++ b/src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
@@ -429,7 +429,7 @@ public class OverlayEvents implements Listener {
                 }
                 */
 
-                EnumMap<ItemTier, Integer> item = new EnumMap(ItemTier.class);
+                EnumMap<ItemTier, Integer> itemCounts = new EnumMap(ItemTier.class);
 
                 boolean sold = formattedText.split(" ")[2].equals("sold");
                 String[] res = formattedText.split("§");
@@ -458,14 +458,14 @@ public class OverlayEvents implements Listener {
                                 + ") item(s) for ";
 
                         String message2 = total
-                                + " (" + WHITE + item.getOrDefault(ItemTier.NORMAL, 0) + LIGHT_PURPLE
-                                + "/" + YELLOW + item.getOrDefault(ItemTier.UNIQUE, 0) + LIGHT_PURPLE
-                                + "/" + item.getOrDefault(ItemTier.RARE, 0)
-                                + "/" + GREEN + item.getOrDefault(ItemTier.SET, 0) + LIGHT_PURPLE
-                                + "/" + AQUA + item.getOrDefault(ItemTier.LEGENDARY, 0) + LIGHT_PURPLE
-                                + "/" + RED + item.getOrDefault(ItemTier.FABLED, 0) + LIGHT_PURPLE
-                                + "/" + DARK_PURPLE + item.getOrDefault(ItemTier.MYTHIC, 0) + LIGHT_PURPLE
-                                + "/" + DARK_AQUA + item.getOrDefault(ItemTier.CRAFTED, 0) + LIGHT_PURPLE
+                                + " (" + WHITE + itemCounts.getOrDefault(ItemTier.NORMAL, 0) + LIGHT_PURPLE
+                                + "/" + YELLOW + itemCounts.getOrDefault(ItemTier.UNIQUE, 0) + LIGHT_PURPLE
+                                + "/" + itemCounts.getOrDefault(ItemTier.RARE, 0)
+                                + "/" + GREEN + itemCounts.getOrDefault(ItemTier.SET, 0) + LIGHT_PURPLE
+                                + "/" + AQUA + itemCounts.getOrDefault(ItemTier.LEGENDARY, 0) + LIGHT_PURPLE
+                                + "/" + RED + itemCounts.getOrDefault(ItemTier.FABLED, 0) + LIGHT_PURPLE
+                                + "/" + DARK_PURPLE + itemCounts.getOrDefault(ItemTier.MYTHIC, 0) + LIGHT_PURPLE
+                                + "/" + DARK_AQUA + itemCounts.getOrDefault(ItemTier.CRAFTED, 0) + LIGHT_PURPLE
                                 + ") item(s) for ";
 
                         if (sold) { // normal selling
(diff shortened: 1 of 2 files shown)


magicus:
I don't understand. What's the point? I can't see what has changed.

... except that variables with comprehensible names has now become anonymous indexes into an array. Seems like a step backwards.

TrapinchO:
> I don't understand. What's the point? I can't see what has changed.

There was a TODO "Jesus Christ rewrite it", so I rewritted it.

EDIT:
> ... except that variables with comprehensible names has now become anonymous indexes into an array. Seems like a step backwards.

There are comments explaining almost everything

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -409,53 +409,57 @@ else if (messageText.matches("[a-zA-Z0-9_ ]{1,19} has given you 20% resistance a
...
-                            } else {
-                                message = TextFormatting.LIGHT_PURPLE + "Scrapped " + total + " (" + TextFormatting.WHITE + countCommon + TextFormatting.LIGHT_PURPLE + "/" + TextFormatting.YELLOW + countUnique + TextFormatting.LIGHT_PURPLE + "/" + countRare + "/" + TextFormatting.GREEN + countSet + TextFormatting.LIGHT_PURPLE + "/" + TextFormatting.AQUA + countLegendary + TextFormatting.LIGHT_PURPLE + "/" + TextFormatting.RED + countFabled + TextFormatting.LIGHT_PURPLE + "/" + TextFormatting.DARK_PURPLE + countMythic + TextFormatting.LIGHT_PURPLE + "/" + TextFormatting.DARK_AQUA + countCrafted + TextFormatting.LIGHT_PURPLE + ") item(s) for " + TextFormatting.YELLOW + s.replace("e", "") + " scrap" + TextFormatting.LIGHT_PURPLE + ".";
+                            String message2;
+                            int total = 0;
+                            for (int i : countTest) {
+                                total += i;
+                            }
+                            if (sold) { // normal selling
+
+                                message2 = TextFormatting.LIGHT_PURPLE + "Sold " + total
```
HeyZeer0: Import TextFormatting.* as static to remove the TextFormating text, will make it cleaner

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -409,53 +409,57 @@ else if (messageText.matches("[a-zA-Z0-9_ ]{1,19} has given you 20% resistance a
...
-                        total++;
-                    } else if (s.startsWith("3")) {
-                        countCrafted++;
-                        total++;
-                    } else if (s.startsWith("e")) {
+                    if (s.startsWith("f")) { countTest[0]++; } // normal
+                    else if (s.startsWith("d") && !s.equals("dYou sold me: ") && !s.equals("dYou scrapped: ") && !s.equals("d, ") && !s.equals("d and ") && !s.equals("d for a total of ")) {
+                        countTest[2]++; // rare
+                    }
+                    else if (s.startsWith("a")) { countTest[3]++; } // set
```
HeyZeer0: you can definitely use a regex with name identifications to match each rarity!

HeyZeer0 (changes requested):
Needs further review after changes.

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: ok

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: This one only or all?

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: Did for blacksmith, might do the rest later

magicus:
I don't think this rewrite is any improvement. It's more or less just a reformatting of the same structure. I think the explicative comment might not be warranted -- it's not that bad -- but if you really want to rewrite it, you need to redo it in a more structured way.

For instance, create a holder class that holds the count of the different kinds of rarities, and with a parse() method that can take in a message and update the count correctly. This could in turn use a map with regexpes -> int:s for each rarity. And also a format() or getMessage() method which returns a properly formatted string. And to do this general, you would iterate over a map of rarities -> colors. (This might be available elsewhere.)

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -409,53 +410,57 @@ else if (messageText.matches("[a-zA-Z0-9_ ]{1,19} has given you 20% resistance a
...
-                            } else {
-                                message = TextFormatting.LIGHT_PURPLE + "Scrapped " + total + " (" + TextFormatting.WHITE + countCommon + TextFormatting.LIGHT_PURPLE + "/" + TextFormatting.YELLOW + countUnique + TextFormatting.LIGHT_PURPLE + "/" + countRare + "/" + TextFormatting.GREEN + countSet + TextFormatting.LIGHT_PURPLE + "/" + TextFormatting.AQUA + countLegendary + TextFormatting.LIGHT_PURPLE + "/" + TextFormatting.RED + countFabled + TextFormatting.LIGHT_PURPLE + "/" + TextFormatting.DARK_PURPLE + countMythic + TextFormatting.LIGHT_PURPLE + "/" + TextFormatting.DARK_AQUA + countCrafted + TextFormatting.LIGHT_PURPLE + ") item(s) for " + TextFormatting.YELLOW + s.replace("e", "") + " scrap" + TextFormatting.LIGHT_PURPLE + ".";
+                            String message2;
+                            int total = 0;
+                            for (int i : countTest) {
+                                total += i;
+                            }
+                            if (sold) { // normal selling
```
magicus: You know that you could have just done `total += countTest`? 🤣

TrapinchO: What, I can add integer and integer array in java?

Wait, what?

magicus: ... or wait, that's not even what you are doing. You are adding 1+2+3....countTest to total. That's not just inefficient, it's completely wrong.

magicus (changes requested):
I think it's better to just remove the comment requesting a rewrite. Otherwise, a completely more structured way to do this is needed. The current patch is a regression in both style, readability and correctness. Sorry to be the deliverer of bad news.

TrapinchO:
> I don't think this rewrite is any improvement. It's more or less just a reformatting of the same structure. I think the explicative comment might not be warranted -- it's not that bad -- but if you really want to rewrite it, you need to redo it in a more structured way.
> 
> For instance, create a holder class that holds the count of the different kinds of rarities, and with a parse() method that can take in a message and update the count correctly. This could in turn use a map with regexpes -> int:s for each rarity. And also a format() or getMessage() method which returns a properly formatted string. And to do this general, you would iterate over a map of rarities -> colors. (This might be available elsewhere.)

I have no idea how (and why) I would make it this complicated.

Right now I am attempting to use regex to the the rarities somehow.

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: I haven't found a better way to get a sum in an intenger array. Imo at least the sum is better than before, because at least i do not increment it in every if

TrapinchO:
> I think it's better to just remove the comment requesting a rewrite. Otherwise, a completely more structured way to do this is needed. The current patch is a regression in both style, readability and correctness. Sorry to be the deliverer of bad news.

Well, I have always wondered when it becomes "Just get away". It wasn't really a suprise since Zeer0 mentioned regex here, so you don't have to be sorry :)

Also I have no idea why in correctness, it works as the previous one

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
magicus: Ah, sorry, the countTest is the array.  It's a horrible name. 

Yes, I'm misreading the code. I see what you are trying to do. 

I would have used something along the lines of `IntStream.of(countTest).sum()` instead, but perhaps that's overdoing it.

Anyway, I think storing the values in an array like this is a bad idea. If anything, you should store it mapped with the rarities as keys.

TrapinchO: countTest was a placeholder name and then I forgot about it, sorry.

I saw this once and just asked "wtf", but yes, it could be used
```
IntStream.of(countTest).sum()
```

Imo better than 8 variables (+ 1 total).
I considered this, but since I have no idea how to initialise hash map with key-pairs already in I dropped it

TrapinchO: changed the 2 things (yes, I know it doesn't have much sense since I will propably rewrite it)

magicus:
Fair enough; the patch is likely correct -- as I said above, I thought I found an error with the totals, but this was not the case. I still fail to see how it is an improvement though. Now we have hard-coded indexes into an array instead of named variables. In what way do you think this is better? If you have an argument for this, I'd like to hear it. To me, a code piece of code that was quite understandable, like:
```
if (s.startsWith("b")) {
                         countLegendary++;
                         total++;
```
is now
```
 if (s.startsWith("b")) { countTest[4]++; } // legendary
 ```

Now suddenly everyone needs to agree that index 4 is legendary. And instead of enforcing this easily by the name, you have to add a comment. That is not an improvement.

I do agree that increasing the total in every if block is a bit silly. I would support a refactoring where the sum was extracted to the end.

**Comment by TrapinchO** [comment]
> Fair enough; the patch is likely correct -- as I said above, I thought I found an error with the totals, but this was not the case. I still fail to see how it is an improvement though. Now we have hard-coded indexes into an array instead of named variables. In what way do you think this is better? If you have an argument for this, I'd like to hear it. To me, a code piece of code that was quite understandable, like:
> 
> ```
> if (s.startsWith("b")) {
>                          countLegendary++;
>                          total++;
> ```
> 
> is now
> 
> ```
>  if (s.startsWith("b")) { countTest[4]++; } // legendary
> ```
> 
> Now suddenly everyone needs to agree that index 4 is legendary. And instead of enforcing this easily by the name, you have to add a comment. That is not an improvement.
> 
> I do agree that increasing the total in every if block is a bit silly. I would support a refactoring where the sum was extracted to the end.

I thought the bad thing was the 8 variables there so I tried to make it look nice? in my eyes. I can't really say anything, it is me and well, you know my coding at least somewhat... (Which is why this resulted in me attempting to do redo it completely anyways)

If you tell me how to initialise hashmaps with content in it I can try to do it using these

**Comment by magicus** [comment]
You have not extracted the common part of the scrapped and sold messages. It is still duplicated, and it's part of what made this code messy.

I still think we have a data class somewhere that connects rarity to their colors. If we do have that, and you wanted to clean this up, you should use it. If we do not have one, you should create one.

I assume the startsWith letters are color codes. You could match explicitly to the named color codes to make it clearer. I assume it would be the same colors that we use later on when formatting the resulting message, giving yet another opportunity to unify and clarify things. 

The strings that we look for do avoid mismatch ("5Blacksmith: ", "dYou sold me: ", etc) can be checked first, and if they match, we will do no further procesing. After that, you will have a simple 1-to-1 matching from color code to rarity type. This can, once again, be nicely supported by a map.

Any, or all, of these changes would constitute the kind of rewrite the author of the "Jesus Christ" rewrite comment meant.

**Comment by TrapinchO** [comment]
> You have not extracted the common part of the scrapped and sold messages. It is still duplicated, and it's part of what made this code messy.

Actually I have no idea why I scrapped this one

> If we do not have one, you should create one.

I think I am lost enough already, thanks

> I assume the startsWith letters are color codes. You could match explicitly to the named color codes to make it clearer. I assume it would be the same colors that we use later on when formatting the resulting message, giving yet another opportunity to unify and clarify things.

I know and I am trying to work with this in the current attempt

> The strings that we look for do avoid mismatch ("5Blacksmith: ", "dYou sold me: ", etc) can be checked first, and if they match, we will do no further procesing. After that, you will have a simple 1-to-1 matching from color code to rarity type. This can, once again, be nicely supported by a map.

Good idea actually

> This can, once again, be nicely supported by a map.
.
> If you tell me how to initialise hashmaps with content in it I can try to do it using these

**Comment by magicus** [comment]
We have an ItemTier class, but it is badly placed in com.wynntils.webapi.profiles.item.enums. I'm getting a bit rusty on the code base, perhaps HeyZeer0 has a good idea on where it should be moved so it could be shared by more parts of the code. From it you can get the rarity level name (or "item tier" name) and color, and and an arbitrarily index.

**Comment by magicus** [comment]
We also have a lot of similar code with explicit code paths for all rarities in `HotbarOverlay`, `ItemPage`, `NametagManager`, `ItemIdentificationOverlay`, `RarityColorOverlay`, and a list of the rarity text colors rendered as CustomColor in `UtilitiesConfig`.

It would make sense to create a RarityLevel or ItemTier class that could support all these use cases.

**Comment by TrapinchO** [comment]
> We have an ItemTier class, but it is badly placed in com.wynntils.webapi.profiles.item.enums. I'm getting a bit rusty on the code base, perhaps HeyZeer0 has a good idea on where it should be moved so it could be shared by more parts of the code. From it you can get the rarity level name (or "item tier" name) and color, and and an arbitrarily index.

Yet it is not useful for me because there is (presumably) no way to get the character that identifies the color from the color enum aside from the enum implementation.
Also it is missing crafted tied btw

**Comment by magicus** [comment]
I think this is starting to look better! I was not aware of the EnumMap class; but it's perfect! Please rename the variable to something more descriptive. It is typically helpful to name maps and arrays using a plural -s, to indicate it contain many items (unless it end in -Map or -Array). So something like itemMap or `itemCounts`.

There is a way to get the color code.  Add something like this to ItemTier:
```
    public char getColorCode() {
        return color.charAt(1);
    }
```

Unfortunately the TextFormatting class does not export the color code letter separately, but only as part of the control string, where the letter is the second character. The ItemTier class also converts the incoming TextFormatting object in the constructor to a String, before storing it. Personally, I would have preferred to store it in its original form, and make the toString() call in getColor. In this case, it does not matter, since we have to take the detour with using toString().charAt(1), but, still... If you change it, I'll support you. :)

Furthermore, the order in ItemTier is odd. The SET tier does not really fit in with the rest, but it is usually listed (as in the Blacksmith selling message) between RARE and LEGENDARY. I have checked the usage of priority in ItemTier, and I think it's perfectly fine to reorder them to match the Blacksmith order.

**Comment by magicus** [comment]
After that, you can compact the matching to something like:

```
char colorCode = s.charAt(0);
if (Character.isLetterOrDigit(colorCode)) {
  Optional<ItemTier> tier = ItemTier.fromColorCode(colorCode)
  tier.ifPresent(itemTier ->  itemCounts.put(itemTier, itemCounts.get(itemTier) + 1));
}
```

and add a method to the enum class to get the correct item tier from color code:
```
    public static Optional<ItemTier> fromColorCode(char ch) {
        return Stream.of(ItemTier.values()).filter(tier -> tier.getColorCode() == ch).findFirst();
    }
```

(code is not tested and might need some tweaking)

**Comment by TrapinchO** [comment]
> I think this is starting to look better! I was not aware of the EnumMap class; but it's perfect! Please rename the variable to something more descriptive. It is typically helpful to name maps and arrays using a plural -s, to indicate it contain many items (unless it end in -Map or -Array). So something like itemMap or itemCounts.

Since I didn't think the idea would work (like regex etc.) it is just a quick testing name

> There is a way to get the color code. Add something like this to ItemTier:
> ```
    public char getColorCode() {
        return color.charAt(1);
    } ```

I do not see a reason for it but sure.

> Unfortunately the TextFormatting class does not export the color code letter separately, but only as part of the control string, where the letter is the second character. The ItemTier class also converts the incoming TextFormatting object in the constructor to a String, before storing it. Personally, I would have preferred to store it in its original form, and make the toString() call in getColor. In this case, it does not matter, since we have to take the detour with using toString().charAt(1), but, still... If you change it, I'll support you. :)
uh, what?

> Furthermore, the order in ItemTier is odd. The SET tier does not really fit in with the rest, but it is usually listed (as in the Blacksmith selling message) between RARE and LEGENDARY. I have checked the usage of priority in ItemTier, and I think it's perfectly fine to reorder them to match the Blacksmith order.

Again, sure

**Comment by magicus** [comment]
And then finally you can replace the message part with:

```
        StringBuilder sb = new StringBuilder();
        for (ItemTier tier : ItemTier.values()) {
            sb.append('/');
            sb.append(tier.color);
            sb.append(itemCounts.getOrDefault(tier, 0));
            sb.append(TextFormatting.LIGHT_PURPLE);
        }
        sb.setCharAt(0, '(');
        sb.append(") item(s) for ");
```

**Comment by magicus** [comment]
At this point, you have transferred all the "knowledge" about rarity levels in Wynntils to a single point, the ItemTier class. If Wynncraft ever introduces a new EPIC item tier (or a MUNDANE below RARE, for that matter), we would only have to update it in a single place, give it it's correct color, and the Blacksmith message code would still continue to work!

**Comment by TrapinchO** [comment]
> And then finally you can replace the message part with:
> 
> ```
>         StringBuilder sb = new StringBuilder();
>         for (ItemTier tier : ItemTier.values()) {
>             sb.append('/');
>             sb.append(tier.color);
>             sb.append(itemCounts.getOrDefault(tier, 0));
>             sb.append(TextFormatting.LIGHT_PURPLE);
>         }
>         sb.setCharAt(0, '(');
>         sb.append(") item(s) for ");
> ```

This part is missing
```total + " ("```



> And then finally you can replace the message part with:
> 
> ```
>         StringBuilder sb = new StringBuilder();
>         for (ItemTier tier : ItemTier.values()) {
>             sb.append('/');
>             sb.append(tier.color);
>             sb.append(itemCounts.getOrDefault(tier, 0));
>             sb.append(TextFormatting.LIGHT_PURPLE);
>         }
>         sb.setCharAt(0, '(');
>         sb.append(") item(s) for ");
> ```

I will try it if I won't tomorrow forget

File: src/main/java/com/wynntils/webapi/profiles/item/enums/ItemTier.java [hunk 4]
```diff
@@ -12,12 +12,13 @@
...
     NORMAL    (0, TextFormatting.WHITE,        (lvl) -> 0),
-    SET       (1, TextFormatting.GREEN,        (lvl) -> (int)Math.ceil(12d + lvl * 1.6)),
     UNIQUE    (2, TextFormatting.YELLOW,       (lvl) -> (int)Math.ceil(5d + lvl * 0.5)),
     RARE      (3, TextFormatting.LIGHT_PURPLE, (lvl) -> (int)Math.ceil(15d + lvl * 1.2)),
+    SET       (1, TextFormatting.GREEN,        (lvl) -> (int)Math.ceil(12d + lvl * 1.6)),
     LEGENDARY (4, TextFormatting.AQUA,         (lvl) -> (int)Math.ceil(35d + (4.8d * lvl))),
     FABLED    (5, TextFormatting.RED,          (lvl) -> (lvl + 5) * 60),
-    MYTHIC    (6, TextFormatting.DARK_PURPLE,  (lvl) -> (lvl + 5) * 18);
+    MYTHIC    (6, TextFormatting.DARK_PURPLE,  (lvl) -> (lvl + 5) * 18),
+    CRAFTED   (7, TextFormatting.DARK_AQUA, (lvl) -> lvl);
```
magicus: Please try and follow the indentation of the rest of the table for the lambda expression.

src/main/java/com/wynntils/webapi/profiles/item/enums/ItemTier.java
```diff
@@ -12,9 +12,9 @@
 public enum ItemTier {
 
     NORMAL    (0, TextFormatting.WHITE,        (lvl) -> 0),
-    SET       (1, TextFormatting.GREEN,        (lvl) -> (int)Math.ceil(12d + lvl * 1.6)),
     UNIQUE    (2, TextFormatting.YELLOW,       (lvl) -> (int)Math.ceil(5d + lvl * 0.5)),
     RARE      (3, TextFormatting.LIGHT_PURPLE, (lvl) -> (int)Math.ceil(15d + lvl * 1.2)),
+    SET       (1, TextFormatting.GREEN,        (lvl) -> (int)Math.ceil(12d + lvl * 1.6)),
```
magicus: This means you should reorder the priority from SET, UNIQUE and RARE.

magicus:
Yes, don't forget about the total. :-) The code snippets I'm sending you here are sketches, there might be syntax errors and omissions (like the total) and other adaptations you need to do.

TrapinchO:
Is this enough or do you want me to try:

```
for (ItemTier tier : ItemTier.values()) {
    if (tier.getColorCode().equals(s.charAt(0))) {
        itemCounts.put(ItemTier.NORMAL, itemCounts.getOrDefault(ItemTier.NORMAL, 0) + 1)
    }
}
```

Also how to get a sum of values within EnumMap?

TrapinchO:
Ok, I think I am done with this PR if you have nothing else

magicus:
Yes, please also unify the adding of the values.

magicus:
As for the EnumMap, perhaps you can do something along the lines of `Stream.of(itemCounts.values().sum()`?

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -409,55 +413,55 @@ else if (messageText.matches("[a-zA-Z0-9_ ]{1,19} has given you 20% resistance a
...
-                            total++;
+                    if (s.equals("dYou sold me: ") //non-item founds
+                            || s.equals("dYou scrapped: ")
+                            || s.equals("d, ")
+                            || s.equals("d and ")
+                            || s.equals("d for a total of ")
+                            || s.equals("5Blacksmith: ")) {
+                        continue;
+                    }
+                    else if (s.matches("e\\d+")) {
```
HeyZeer0: else not necessary you're continuing above

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -409,55 +413,55 @@ else if (messageText.matches("[a-zA-Z0-9_ ]{1,19} has given you 20% resistance a
...
+                        for (ItemTier tier : ItemTier.values()) {
+                            total += itemCounts.getOrDefault(tier, 0);
+                        }
+
+                        StringBuilder message = new StringBuilder(); // creates the message
+                        for (ItemTier tier : ItemTier.values()) {
+                            message.append('/' + tier.getColor() + itemCounts.getOrDefault(tier, 0));
+                            message.append(LIGHT_PURPLE);
+                        }
+                        message.append(") item(s) for ");
```
HeyZeer0: space above this

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -409,55 +413,55 @@ else if (messageText.matches("[a-zA-Z0-9_ ]{1,19} has given you 20% resistance a
...
+
+                        if (formattedText.split(" ")[2].equals("sold")) { // normal selling
+                            message.insert(0, LIGHT_PURPLE + "Sold " + total + " ");
+                            message.append(GREEN + s.replace("e", "") + (char) 0xB2 + LIGHT_PURPLE + ".");
+
+                        } else { // scrapping
+                            message.insert(0, LIGHT_PURPLE + "Scrapped " + total + " ");
+                            message.append(YELLOW + s.replace("e", "") + " scrap" + LIGHT_PURPLE + ".");
                         }
+                        GameUpdateOverlay.queueMessage(message.toString()); // send the redirect messsage
```
HeyZeer0, line 459: space above this

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -409,55 +413,55 @@ else if (messageText.matches("[a-zA-Z0-9_ ]{1,19} has given you 20% resistance a
...
+                        e.setCanceled(true); // remove the chat message
                     }
+                    // item counter
+                    else if (s.startsWith("f")) { itemCounts.put(ItemTier.NORMAL, itemCounts.getOrDefault(ItemTier.NORMAL, 0) + 1); } // normal
```
HeyZeer0: Don't make all the code in a single line, remove the else and place a continue. Same for all below ones

src/main/java/com/wynntils/webapi/profiles/item/enums/ItemTier.java
```diff
@@ -39,6 +40,8 @@ public int getRerollPrice(int level, int rolledAmount) {
         return rolledAmount == 0 ? basePrice : basePrice * (int)Math.pow(5, rolledAmount);
```
HeyZeer0: nice catch on this class but I just noticed this thing is also wrong, mind changing it as well?
it should be just:
`return basePrice * (int)Math.pow(55, rolledAmount);`

because:
0 rolls -> basePrice * 5^0 = basePrice * 1 = basePrice
1 roll -> basePrice * 5^1 = basePrice * 5

basically, we are doing an unnecessary check.

HeyZeer0 (changes requested):
LGTM once changes done.

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -15,12 +15,14 @@
...
 import com.wynntils.modules.utilities.overlays.hud.*;
 import com.wynntils.webapi.WebManager;
+import com.wynntils.webapi.profiles.item.enums.ItemTier;
 import net.minecraft.client.Minecraft;
 import net.minecraft.init.MobEffects;
 import net.minecraft.inventory.IInventory;
 import net.minecraft.network.play.server.*;
 import net.minecraft.potion.Potion;
 import net.minecraft.util.text.TextComponentString;
+import static net.minecraft.util.text.TextFormatting.*;
```
magicus, line 26: We don't normally do static imports. Now that the color codes are used in just a few places, I think it is better to be explicit about what class they come from, and remove the static import.

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -409,55 +413,55 @@ else if (messageText.matches("[a-zA-Z0-9_ ]{1,19} has given you 20% resistance a
...
+                    if (s.equals("dYou sold me: ") //non-item founds
+                            || s.equals("dYou scrapped: ")
+                            || s.equals("d, ")
+                            || s.equals("d and ")
+                            || s.equals("d for a total of ")
+                            || s.equals("5Blacksmith: ")) {
+                        continue;
+                    }
+                    else if (s.matches("e\\d+")) {
+                        //int total = IntStream.of(itemCount).sum(); // sum of the items
```
magicus: Remove the commented out code.

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -409,55 +413,55 @@ else if (messageText.matches("[a-zA-Z0-9_ ]{1,19} has given you 20% resistance a
...
+                        StringBuilder message = new StringBuilder(); // creates the message
+                        for (ItemTier tier : ItemTier.values()) {
+                            message.append('/' + tier.getColor() + itemCounts.getOrDefault(tier, 0));
+                            message.append(LIGHT_PURPLE);
+                        }
+                        message.append(") item(s) for ");
+                        message.setCharAt(0, '(');
+
+                        if (formattedText.split(" ")[2].equals("sold")) { // normal selling
+                            message.insert(0, LIGHT_PURPLE + "Sold " + total + " ");
```
magicus: Instead of using insert(), which is an expensive operation, create a new string builder and append the common part. E.g. rename the current SB message to "messageCounts" or so, and then create a new SB `message`, and start with `message.append(messageCounts)`

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
magicus: Or, better, as I suggested, do this in a general way instead of testing each rarity by itself.

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -409,55 +413,55 @@ else if (messageText.matches("[a-zA-Z0-9_ ]{1,19} has given you 20% resistance a
...
+                            message.append(GREEN + s.replace("e", "") + (char) 0xB2 + LIGHT_PURPLE + ".");
```
magicus: Instead of the hardcoded 0xB2, use `EmeraldSymbols.EMERALDS`

magicus: Also, I think we should skip the trailing dot at the end of this (and the scrapped) message.

I don't fully understand what the replace "e" -> "" is doing, but I see this is in the original code as well. If you understand this (perhaps by looking at the original `s` string), please add a comment.

magicus (changes requested):

magicus:
Almost done! Just a few fixes left, hang on and you'll be good to go! :)

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: Discuss this with @HeyZeer0, he told me to do it

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: Sorry, forgot

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: Because instead of the "e" you use the emerald sign

TrapinchO:
> Almost done! Just a few fixes left, hang on and you'll be good to go! :)

10 change requests - few fixes... :)

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: done in the next commit

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: > Don't make all the code in a single line, remove the else and place a continue. Same for all below ones

Sure

src/main/java/com/wynntils/webapi/profiles/item/enums/ItemTier.java
TrapinchO: done for the next commit

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: done for the next commit

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: > Now that the color codes are used in just a few places, I think it is better to be explicit about what class they come from, and remove the static import.

The green rectangles are places where colors are used (With `TextFormatting.` before the color). They are used in many palces
mc_colors

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: done for the next commit

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: done now?

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: done now?

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
magicus: Oh, great, I get to fight @HeyZeer0. ;-) I think consistency is important. If you do this static import, maybe you can come back in a separate PR and remove the TextFormatting prefix for the other places..?

(But personally I think it's mostly helpful to keep the prefix; it helps you understand that it is a text formatting code and not an RGB color triple. But I will not take that fight with Zeer0 here in this PR. :-))

TrapinchO: > I think consistency is important. If you do this static import, maybe you can come back in a separate PR and remove the TextFormatting prefix for the other places..?

That was my intention.

> (But personally I think it's mostly helpful to keep the prefix; it helps you understand that it is a text formatting code and not an RGB color triple. But I will not take that fight with Zeer0 here in this PR. :-))

For me personally those are just colors, meaning it is not reakky necessary. In other cases I would propably recommend the prefix too

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: Since we are using another way to do it I am resolving this one

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -444,24 +444,48 @@ else if (s.matches("e\\d+")) {
...
+                        itemCounts.put(ItemTier.FABLED, itemCounts.getOrDefault(ItemTier.FABLED, 0) + 1);
+                        continue;
+                    }
+                    if (s.startsWith("5")) { // mythic
+                        itemCounts.put(ItemTier.MYTHIC, itemCounts.getOrDefault(ItemTier.MYTHIC, 0) + 1);
+                        continue;
+                    }
+                    if (s.startsWith("3")) { // crafted
+                        itemCounts.put(ItemTier.CRAFTED, itemCounts.getOrDefault(ItemTier.CRAFTED, 0) + 1);
+                    }
```
magicus, line 468: Just needs to compact this to the loop I provided. :)

magicus (changes requested):

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
TrapinchO: done

magicus:
Now it really looks great!

All I want is a final confirmation that you have tested this code -- sell a bunch of items of different rarities to a blacksmith and verify that the message looks correct. When you confirm that this works, I'll integrate the PR.

TrapinchO:
> Now it really looks great!
> 
> All I want is a final confirmation that you have tested this code -- sell a bunch of items of different rarities to a blacksmith and verify that the message looks correct. When you confirm that this works, I'll integrate the PR.

You need to fight HeyZeer0 first :)

Yes, the code works. I only tested it with normal-rare rarities (because I don't have rarer items to sell and TM is down), but it works

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
HeyZeer0: Oh, we do use static imports in a lot of places! they drastically reduce the spam on your facing while calling a bunch of color enums.
We use it on GlStateManager as well, to not keep having to call a hell lot of functions that keep
repeating GlStateManager.

Totally fine 👍🏻

src/main/java/com/wynntils/webapi/profiles/item/enums/ItemTier.java
```diff
@@ -36,9 +37,11 @@ public String getColor() {
     public int getRerollPrice(int level, int rolledAmount) {
         int basePrice = rerollFormula.applyAsInt(level);
 
-        return rolledAmount == 0 ? basePrice : basePrice * (int)Math.pow(5, rolledAmount);
+        return basePrice * (int)Math.pow(55, rolledAmount);
```
HeyZeer0: just 5, not 55!

HeyZeer0 (changes requested):
LGTM once changed from 55 to 5!

src/main/java/com/wynntils/webapi/profiles/item/enums/ItemTier.java
TrapinchO: you were the one who told me to use 55 (you sent the code) btw.

I will correct it soon

Btw what is the formula for crafted?

TrapinchO: corrected

HeyZeer0: who knows

HeyZeer0 (approved):

src/main/java/com/wynntils/webapi/profiles/item/enums/ItemTier.java
TrapinchO: You do.
You are Wynncraft dev after all....

(Yes, I know you propably can't use it for Wynntils, just a joke)

magicus:
I will not fight Zeer0. :) I accept his desire to do static imports.

magicus (approved):

PR #288, opened 2021-01-04
Add option to apply update at Minecraft exit instead of immediately

magicus:
This is inspired by https://github.com/Wynntils/Wynntils/pull/274 and hints from @P0keDev.

I did some general cleanup in the updating area as well. It still seems like we have a bit of duplicated logic here, but I can't be bothered to untangle it completely right now.

Anyway, this patch makes the "Update before join?" screen instead give the user two options: `Update now` and `Update at exit`. The first works like it always has done and restarts Minecraft forcefully. The second will just delay updating until the user exits Minecraft by themselves. I also added some more description on what the different options mean.


magicus:
Maybe I could have pushed directly but I'm feeling a bit rusty and would appreciate some reviews. :)

HeyZeer0 (approved):
Nicely done 😄 
Nice to see you again!

PR #289, opened 2021-01-04
Athena offline checks

P0keDev:
Some small changes to make the mod run more smoothly during Athena outages.

This adds some checks to see if the mod successfully authenticated with Athena, and if not, avoids sending additional requests. Importantly, attempting to upload the user's config files does not reattempt an Athena login if the previous logins failed--this would cause Minecraft to hang for a long time and seems unnecessary. I also added a small message warning the user of decreased functionality when joining Wynncraft after failing to connect to Athena.


quick007:
pog!!! This fixes the crashes as well right?

HeyZeer0 (approved):
LGTM!

PR #290, opened 2021-01-05
Removed prof icons, used the "ProfessionType" enum instead

TrapinchO:


TrapinchO:
mentioning @HeyZeer0 because I can't request review

magicus:
Nice catch!

PR #294, opened 2021-01-15
Code quality improvements

HighCrit:
Fixed among other things potential memory leaks and concurrency issues

### DIFF
--- a/src/main/java/com/wynntils/Reference.java
+++ b/src/main/java/com/wynntils/Reference.java
@@ -45,7 +45,7 @@ public class Reference {
         onLobby = onServer && !onWorld;
     }
 
-    public static String getUserWorld() {
+    public static synchronized String getUserWorld() {
         return userWorld;
     }
 
--- a/src/main/java/com/wynntils/Reference.java
+++ b/src/main/java/com/wynntils/Reference.java
@@ -24,7 +24,7 @@ public class Reference {
     public static final String MINECRAFT_VERSIONS = "1.12,1.12.2";
     public static String VERSION = "";
     public static int BUILD_NUMBER = -1;
-    public static final File MOD_STORAGE_ROOT = ModCore.mc() == null ? new File("wynntils") : new File(ModCore.mc().gameDir, "wynntils");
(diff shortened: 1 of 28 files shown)


src/main/java/com/wynntils/core/framework/FrameworkManager.java
```diff
@@ -52,6 +52,8 @@
         registeredOverlays.put(Priority.HIGHEST, new ArrayList<>());
     }
 
+    private FrameworkManager() {}
```
HeyZeer0: no reason to have a constructor for a class that's not instantiable.

src/main/java/com/wynntils/core/CoreManager.java
```diff
@@ -10,6 +10,7 @@
 import net.minecraftforge.common.MinecraftForge;
 
 public class CoreManager {
+    private CoreManager() {}
```
HeyZeer0: no reason to have a constructor for a class that's not instantiable.

src/main/java/com/wynntils/Reference.java
```diff
@@ -64,9 +64,9 @@ public static boolean inClassSelection() {
     public static boolean developmentEnvironment = false;
 
     public static class ServerIPS {
+        private ServerIPS() {}
```
HeyZeer0: no reason to have a constructor for a class that's not instantiable.

HeyZeer0 (changes requested):
LGTM once changes done.

Nice work 👍🏻

src/main/java/com/wynntils/core/CoreManager.java
magicus: Actually, you always have a default constructor unless you specify another explicitly.

This change is implementing the recommended pattern for non-instantiable classes like singletons.

magicus (approved):

src/main/java/com/wynntils/core/CoreManager.java
HighCrit: what he said

HeyZeer0: still, it doesn't matter and just pollutes your view with a completely useless constructor.

HeyZeer0 (approved):

PR #297, opened 2021-01-25
Block trades/duels if an item is held

P0keDev:
Blocks the player from making trade/duel requests without an empty hand if enabled. Does not interfere with the actual click (spells and melee attacks will still work fine). Useful to avoid accidentally opening GUIs in group fights, etc.


quick007:
is there a toggle for this?

HeyZeer0 (approved):
LGTM

PR #299, opened 2021-01-27
Minor map features + small fixes

phantamanta44:
This PR implements the following small map features:

* A config option is added that lets the minimap remain visible while in non-mapped areas, in case players want to use the compass in those areas.
* A config option is added that draws the distance to the compass position over the minimap icon.

Additionally, this PR fixes a couple small issues:

* The scaling algorithm for the compass beacon when it's too far away doesn't account for partial ticks, so it looks jittery while the player is moving.
* Many Wynntils sounds are played as positioned records, so they fade out if the player is moving while they play.


HeyZeer0 (approved):
Awesome work, thanks :D

PR #300, opened 2021-01-31
Lootrun improvements

P0keDev:
Adds several helpful commands for lootrunning:

/lootrun undo - Standing on a previous part of the lootrun path while recording and using this command will roll the recording back to the position you are on.
/lootrun note - Adds holographic text to the current lootrun where you are standing. Useful for explaining how to do certain chest puzzles.
/lootrun addchest/removechest - Manually adds or removes a marked chest from the lootrun. Useful for going back and adding chests on the route that weren't present during recording (someone else looting them, etc.)

Any lootruns recorded prior to this change will still work, and new lootruns will be compatible with previous versions of Wynntils.

image

### DIFF
--- a/src/main/java/com/wynntils/modules/map/instances/LootRunNote.java
+++ b/src/main/java/com/wynntils/modules/map/instances/LootRunNote.java
@@ -1,0 +1,94 @@
+package com.wynntils.modules.map.instances;
+
+import com.wynntils.ModCore;
+import com.wynntils.core.framework.rendering.ScreenRenderer;
+import com.wynntils.core.framework.rendering.SmartFontRenderer;
+import com.wynntils.core.framework.rendering.colors.CustomColor;
+import com.wynntils.core.utils.StringUtils;
+import com.wynntils.core.utils.objects.Location;
+
+import net.minecraft.client.entity.EntityPlayerSP;
+import net.minecraft.client.gui.FontRenderer;
+import net.minecraft.client.renderer.entity.RenderManager;
+
+import static net.minecraft.client.renderer.GlStateManager.*;
+
+public class LootRunNote {
+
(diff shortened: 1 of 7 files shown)


src/main/java/com/wynntils/modules/map/instances/LootRunNote.java
```diff
@@ -0,0 +1,94 @@
...
+import com.wynntils.core.utils.StringUtils;
+import com.wynntils.core.utils.objects.Location;
+
+import net.minecraft.client.entity.EntityPlayerSP;
+import net.minecraft.client.gui.FontRenderer;
+import net.minecraft.client.renderer.entity.RenderManager;
+
+import static net.minecraft.client.renderer.GlStateManager.*;
+
+public class LootRunNote {
```
HeyZeer0, line 16: Use the FakeEntity system here, you'll have much more control over and it'll generify the code a bit more!

HeyZeer0 (changes requested):
Looks really good! Nice work on this!
Just this small change and LGTM!

magicus:
Fantastic improvements! 

One thing I was thinking of but never got around to implement was some kind of "lootrun merge". The idea was partly that of being able to add missing chests, but also something along the lines of being able to get "smoother" lootruns. The first time you figure out a good lootrun you run around pretty erratically, but later on, as you can follow the previous lootrun, you can make straighter or more clever paths. The idea I had was that you could take two (or more) recorded lootruns and "merge" them together, to form a average of the path (and the superset of all chests).

I still think that could be a useful feature, but I got stuck at trying to figure out the maths for successfully averaging the lootrun curves. Just bringing it up as an idea, if you feel like taking it on. :-)

PR #306, opened 2021-02-18
Fix allowRetry

iwei20:
Thanks to @McPlayHD for mentioning the problem.  allowRetry was at the beginning, so when the horse spawned and the client event triggered for a second time it returned NO HORSE.  I moved that back into the null/far if statement.

Further testing revealed that there might need to be some modification to client events to actually get it to respawn, or maybe something else.  Current behavior is only despawning the horse on pressing the button which is in my opinion better than spitting out an error message, but I would prefer to get it to my original intention.

This PR is just to fix my original mistake.

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/managers/MountHorseManager.java
+++ b/src/main/java/com/wynntils/modules/utilities/managers/MountHorseManager.java
@@ -35,6 +35,8 @@ public class MountHorseManager {
     private static final int remountTickDelay = 5;
     private static final int spawnAttempts = 8;
 
+    public static boolean ingamePrevention = false;
+
     public static boolean isPlayersHorse(Entity horse, String playerName) {
         return (horse instanceof AbstractHorse) && isPlayersHorse(horse.getCustomNameTag(), playerName);
     }
@@ -65,9 +67,15 @@ public class MountHorseManager {
     }
 
     private static void tryDelayedSpawnMount(Minecraft mc, HorseData horse, int attempts) {
+        if (ingamePrevention) {
+            ingamePrevention = false;
+            return;
+        }
+
         if (attempts <= 0) {
             String message = getMountHorseErrorMessage(MountHorseStatus.NO_HORSE);
             GameUpdateOverlay.queueMessage(TextFormatting.DARK_RED + message);
+            ingamePrevention = false;
             return;
         }
 
@@ -121,7 +129,11 @@ public class MountHorseManager {
                 tryDelayedSpawnMount(mc, horse, spawnAttempts);
                 return MountHorseStatus.SPAWNING;
             } else {
-                ClientEvents.isAwaitingHorseMount = true;
+                if (ingamePrevention) {
+                    ingamePrevention = false;
+                } else {
+                    ClientEvents.isAwaitingHorseMount = true;
+                }
                 return MountHorseStatus.SUCCESS;
             }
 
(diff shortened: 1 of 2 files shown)


iwei20:
Finally got behavior to intended state by adding 25-tick delay between despawn and respawn, thanks to @P0keDev for telling me Delay exists

iwei20:
Thanks to @HighCrit for the suggestion on multi attempt spawning

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -182,6 +183,7 @@ public void onChatToRedirect(ChatEvent.Pre e) {
                 case "There is no room for a horse.":
                     GameUpdateOverlay.queueMessage(DARK_RED + "There is no room for a horse.");
                     e.setCanceled(true);
+                    MountHorseManager.ingamePrevention = true;
```
HeyZeer0: Make this a proper setter on MountHorseManager.
e.g `MountHorseManager.setInGamePrevention(true);`

src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -190,6 +192,7 @@ public void onChatToRedirect(ChatEvent.Pre e) {
                 case "Your horse is scared to come out right now, too many mobs are nearby.":
                     GameUpdateOverlay.queueMessage(DARK_RED + "Too many mobs nearby to spawn your horse");
                     e.setCanceled(true);
+                    MountHorseManager.ingamePrevention = true;
```
HeyZeer0: Make this a proper setter on MountHorseManager.
e.g `MountHorseManager.setInGamePrevention(true);`

src/main/java/com/wynntils/modules/utilities/managers/MountHorseManager.java
```diff
@@ -40,58 +48,100 @@ public static boolean isPlayersHorse(String horseName, String playerName) {
...
+            }
+
+            player.inventory.currentItem = horse.getInventorySlot();
+            playerController.processRightClick(player, player.world, EnumHand.MAIN_HAND);
+            player.inventory.currentItem = prev;
             if (far) {
-                mc.playerController.processRightClick(player, player.world, EnumHand.MAIN_HAND);
+                tryDelayedSpawnMount(mc, horse, spawnAttempts);
+                return MountHorseStatus.SPAWNING;
+            } else {
```
HeyZeer0, line 138: else is not required since you're returning above

src/main/java/com/wynntils/modules/utilities/managers/MountHorseManager.java
```diff
@@ -40,58 +48,100 @@ public static boolean isPlayersHorse(String horseName, String playerName) {
...
 
         int prev = mc.player.inventory.currentItem;
+        new Delay(() -> {
+            mc.player.inventory.currentItem = horse.getInventorySlot();
+            mc.playerController.processRightClick(mc.player, mc.player.world, EnumHand.MAIN_HAND);
+            mc.player.inventory.currentItem = prev;
+
+            if (findHorseInRadius(mc) != null) {
+                ClientEvents.isAwaitingHorseMount = true;
+            } else {
```
HeyZeer0: replace the else with a return:
```
if (findHorseInRadius(mc) != null) {
    ClientEvents.isAwaitingHorseMount = true;
    return;
}

// below
```

HeyZeer0 (changes requested):
LGTM once addressed, nice work on this, and sorry for the delay!

PR #307, opened 2021-02-22
Server selector displays server number properly

ryanzhoudev:
Replaces the # of items display with actual server numbers, for servers above 64. Uses same method as the display item level feature, except it doesn't check if you're holding down a key. Tested on server 70, and also seems to not affect anything else in-game.

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/overlays/inventories/ServerNumberOverlay.java
+++ b/src/main/java/com/wynntils/modules/utilities/overlays/inventories/ServerNumberOverlay.java
@@ -6,9 +6,7 @@ package com.wynntils.modules.utilities.overlays.inventories;
 
 import com.wynntils.core.events.custom.RenderEvent;
 import com.wynntils.core.framework.interfaces.Listener;
-import net.minecraft.client.Minecraft;
 import net.minecraft.item.ItemStack;
-import net.minecraft.util.text.TextComponentString;
 import net.minecraft.util.text.TextFormatting;
 import net.minecraftforge.fml.common.eventhandler.SubscribeEvent;
 
--- a/src/main/java/com/wynntils/modules/utilities/overlays/inventories/ServerNumberOverlay.java
+++ b/src/main/java/com/wynntils/modules/utilities/overlays/inventories/ServerNumberOverlay.java
@@ -1,0 +1,39 @@
+/*
+ *  * Copyright © Wynntils - 2021.
+ */
+
+package com.wynntils.modules.utilities.overlays.inventories;
+
+import com.wynntils.core.events.custom.RenderEvent;
+import com.wynntils.core.framework.interfaces.Listener;
+import com.wynntils.modules.utilities.configs.UtilitiesConfig;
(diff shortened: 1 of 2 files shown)


src/main/java/com/wynntils/modules/utilities/overlays/inventories/ServerNumberOverlay.java
```diff
@@ -0,0 +1,39 @@
...
+import net.minecraft.client.Minecraft;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.text.TextFormatting;
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent;
+
+public class ServerNumberOverlay implements Listener {
+
+    @SubscribeEvent
+    public void onItemOverlay(RenderEvent.DrawItemOverlay event) {
+        if (!UtilitiesConfig.Items.INSTANCE.itemLevelOverlayOutsideGui && Minecraft.getMinecraft().currentScreen == null) return;
```
P0keDev: The config check is unnecessary, since this code is only applicable when a GUI is open (there will never be server items in the hotbar, which is what the config option is for)

src/main/java/com/wynntils/modules/utilities/overlays/inventories/ServerNumberOverlay.java
```diff
@@ -0,0 +1,39 @@
...
+
+        ItemStack serverStack = event.getStack();
+        String serverName = TextFormatting.getTextWithoutFormattingCodes(serverStack.getDisplayName());
+
+        assert serverName != null;
```
P0keDev: There should not be asserts in any production code, replace with an if statement that returns when null

src/main/java/com/wynntils/modules/utilities/overlays/inventories/ServerNumberOverlay.java
```diff
@@ -0,0 +1,39 @@
...
+        // There is probably a better way to do this, but I can't get the container title "Wynncraft Servers" in this event, so here we are
+
+        // server number replacements
+        try {
+            int serverNumber = Integer.parseInt(serverName.replaceAll("[^0-9]+", ""));
+            event.setOverlayText(String.valueOf(serverNumber));
+        } catch (NumberFormatException ignored) {
+            // I know ignoring exceptions is probably not good, but I don't see a reason to throw some error or two into the user's logs
+            // This is only triggered when it scans for server names, looking for integers
+        }
```
P0keDev: You can assure more accurate matching and handle the issue of number parsing errors with a regex pattern for server names, something like `World (\d+)(?: \(Recommended!\))?` (I can't recall if that's the exact format for a recommended world, so don't use that without checking). That way, you can verify the item name and grab the server number at the same time, without the ugly try/catch, and it won't try to parse random items whose names start with 'World'.

P0keDev (changes requested):
Feel free to tag me on discord if you need any clarifications

P0keDev (approved):

HeyZeer0 (approved):

PR #309, opened 2021-02-28
Bugfixes & Various QoL

P0keDev:
Lots of small but important things

### DIFF
--- a/src/main/java/com/wynntils/webapi/profiles/item/enums/ItemTier.java
+++ b/src/main/java/com/wynntils/webapi/profiles/item/enums/ItemTier.java
@@ -11,17 +11,16 @@ import com.wynntils.modules.utilities.configs.UtilitiesConfig;
 import net.minecraft.util.text.TextFormatting;
 
 import java.util.concurrent.Callable;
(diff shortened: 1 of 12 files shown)


src/main/java/com/wynntils/webapi/profiles/item/enums/ItemTier.java
```diff
@@ -106,7 +103,7 @@ public int getItemIdentificationCost(int level) {
     }
 
     public int getRerollPrice(int level, int rolledAmount) {
-        int basePrice = rerollFormula.applyAsInt(level);
+        int basePrice = getItemIdentificationCost(level);
```
magicus, line 106: @P0keDev This will change the price, afaict. Are you sure the new way to calculate it is correct? (ie meaning the old one was buggy)

P0keDev: rerollFormula and getItemIdentificationCost accomplished the same exact thing (except for a typo in the fabled one), so I removed the former. The numbers themselves were changed because the cost formulas were updated in 1.20.

PR #310, opened 2021-03-01
Powder & Damage Calculation Overhaul

ParkerHitch:
Big redo of powders. This includes code to calculate the damage values on weapons so we can get what tier powders they have. This could also be used in future building capabilities.
This could maybe still use some optimization.


ParkerHitch:
Had to make a new PR cuz I screwed with my repo -_- sry

PR #311, opened 2021-03-05
Added decimal ID slider option, fixed rounding of values near color boundaries

ryanzhoudev:
Decimal ID slider:
This is just a slider that goes from 0-4 that decides how many decimals to display for your % ID's. Pretty self-explanatory, requires inventory reload, either by opening bank or tapping shift while hovering over an item.
Value rounding:
Since I changed up how the % id's work somewhat, I might as well just fix rounding to be more normal. Old version was truncate, new version rounds up at half. This ends up being more consistent as well, for different decimal settings. See linked example for how the colors may change:
[https://donkeyblaster69.tk/xbackbone/QOMa4/KAbUTEku18.png](https://donkeyblaster69.tk/xbackbone/QOMa4/KAbUTEku18.png)

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/enums/IdentificationType.java
+++ b/src/main/java/com/wynntils/modules/utilities/enums/IdentificationType.java
@@ -10,11 +10,8 @@ import com.wynntils.modules.utilities.interfaces.IIdentificationAnalyser;
 import com.wynntils.webapi.profiles.item.objects.IdentificationContainer;
 import org.apache.commons.lang3.math.Fraction;
 
-import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.text.DecimalFormat;
-import java.text.NumberFormat;
-import java.util.regex.Pattern;
 
 import static net.minecraft.util.text.TextFormatting.*;
 
--- a/src/main/java/com/wynntils/modules/utilities/enums/IdentificationType.java
+++ b/src/main/java/com/wynntils/modules/utilities/enums/IdentificationType.java
@@ -4,10 +4,17 @@
 
 package com.wynntils.modules.utilities.enums;
 
+import com.wynntils.modules.utilities.configs.UtilitiesConfig;
 import com.wynntils.modules.utilities.instances.IdentificationResult;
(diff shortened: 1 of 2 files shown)


src/main/java/com/wynntils/modules/utilities/enums/IdentificationType.java
```diff
@@ -17,9 +21,22 @@
 
         @Override
         public String getTitle(double specialAmount) {
-            int amount = normalize(specialAmount);
-
-            return amount != -1 ? getColor(amount) + "[" + amount + "%]" : GOLD + " NEW";
+            // Use DecimalFormat and a switch to determine how our decimals are formatted
+            String decimalPattern;
+            switch (UtilitiesConfig.Identifications.INSTANCE.decimalPlaces) {
```
HeyZeer0, line 26: Use an enum that stores the decimal format, compiling it every time is not necessary and will avoid this switch.

src/main/java/com/wynntils/modules/utilities/enums/IdentificationType.java
```diff
@@ -41,20 +58,23 @@ public double getRelativeValue(IdentificationContainer container, int current, b
             return isInverted ? 1 - value : value;
         }
 
-        String getColor(double amount) {
+        String getColor(double rawamount) {
+            // Round the provided decimals to make sure values such as 79.8% show up as green instead of yellow
+            // This is so zero decimals and four decimals are consistent with true roll percentages
+            // Also most people seem to prefer proper rounding anyway
+            int amount = (int)Math.round(rawamount);
```
HeyZeer0, line 65: This will be misleading without decimals.

HeyZeer0 (changes requested):

src/main/java/com/wynntils/modules/utilities/enums/IdentificationType.java
ryanzhoudev: Could you elaborate a bit on what you mean by misleading?
I'm pretty sure 0.49 -> 0 and 0.5 -> 1, and colors are based on the rounded decimals, so we shouldn't get any green 79's or blue 95's

PR #313, opened 2021-03-19
Powders

ParkerHitch:
Created a branch in my repo to try to be organized and forgot that my old pr was on a different branch.
This is a redo of powders, now introducing tiers! This includes code to calculate the damage values on weapons so we can get what tier powders they have. Currently, this is only used in the Wynndata export feature but I plan on making a Wynnbuilder export soon as well. This could also be used in future building capabilities.
This could maybe still use some optimization.

### DIFF
--- a/src/main/java/com/wynntils/core/framework/enums/Powder.java
+++ b/src/main/java/com/wynntils/core/framework/enums/Powder.java
@@ -4,25 +4,78 @@
 
 package com.wynntils.core.framework.enums;
 
+import com.wynntils.core.utils.ItemUtils;
+import com.wynntils.webapi.WebManager;
+import com.wynntils.webapi.profiles.item.ItemProfile;
+import net.minecraft.item.ItemStack;
 import net.minecraft.util.text.TextFormatting;
 
+import java.util.Arrays;
+import java.util.Hashtable;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 public enum Powder {
 
     EARTH('✤', TextFormatting.DARK_GREEN),
+    E1(3, 6, 17, 2, 1), E2(6, 9, 21, 4, 2), E3(8, 14, 25, 8, 3), E4(11, 16, 31, 14, 5), E5(15, 18, 38, 22, 9), E6(18, 22, 46, 30, 13),
+
     THUNDER('✦', TextFormatting.YELLOW),
+    T1(1, 8, 9, 3, 1), T2(1, 13, 11, 5, 1), T3(2, 18, 14, 9, 2), T4(3, 24, 17, 14, 4), T5(3, 32, 22, 20, 7), T6(5, 40, 28, 28, 10),
+
     WATER('❉', TextFormatting.AQUA),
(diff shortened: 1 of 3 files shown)


magicus:
I have just skimmed through the patch, but my main worry here is that this will just break through the combinatorial roof! Can you give me a proof how this will have a reasonable upper bound on the number of combinations you will have to look through? Otherwise this might be a potential game crasher.

P0keDev:
> 
> 
> I have just skimmed through the patch, but my main worry here is that this will just break through the combinatorial roof! Can you give me a proof how this will have a reasonable upper bound on the number of combinations you will have to look through? Otherwise this might be a potential game crasher.

Singularity's 15 powder slots would result in a number of combinations over 200 times that of the Java integer limit, immediately breaking things before the loop even starts. No other weapons are this bad, but singu being a commonly-used mage mythic would result in it being exported to wynndata relatively frequently.

Reverse engineering powders from damage/defense is certainly interesting, but also impractical. If the purpose is wynndata/wynnbuilder exporting, assuming all powders are tier 6 is probably going to be correct 99% of the time - I'm not sure that chasing those edge cases is worth the complexity of doing so.

magicus:
My initial implementation assumed powders to be tier 1, but redirected you to Wynndata's Powder setting screen, so you can manually update them. If this is an unlikely default, maybe it is more convenient in assuming tier 6 in the redirect to Wynndata, so that it is most likely that you can just confirm that value.

(I'm personally just a measly sub-80 level so I have not much experience of that kind of powder tiers... maybe that's caused me to select an improper default)

ParkerHitch:
Ok that makes sense so thanks for the feedback. Perhaps it would just be best to check if the powders are all t6, all t5, all t4, etc. I would think that people would be more likely to use the same tier powders just for symmetry most of the time. This also reduces the number of combinations to 6 or less for all equipment. If none of the combos work we can just use whichever one got closest to the real value. Maybe send a notification in chat.
Another option might be some type of pruning function. Now I am probably a bit out of my depth with this, but I saw a couple youtube vids where they used alpha-beta pruning to dramatically reduce the number of combinations a chess engine needs to check. This would be ideal if we cared about every scenario. This might be a bit advanced but I bet I could figure it out. The question is if we care about people who used a random combo that much.
Let me know what you think.

magicus:
Checking for the "symmetric" options seems reasonable. That means you have a upper bound of O(C) with C = max_tiers. I think we can live with that. Still, you should make sure this is calculated only on demand, i.e. when someone wants to open the build on Wynndata, and not in the background ahead of time. Maybe this is already the case, I didn't check the code as carefully for that.

And sure, if you can get a simple check which symmetric option is closest if no one match exactly, and pick that, it might be a better fallback than going to tier 1 for all, or tier 6 for all.

ParkerHitch:
Pretty sure this is the final draft unless I did something really dumb. Sorry for the delay between commits.

magicus (commented):
This version seem to avoid the excessive big-O complexity of the initial patch. However, I'mm not convinced this is a good addition. What if you have like T4 T4 T3? That can't be that uncommon, can it? If I read the source correctly, you try to pick the closest thing, so I guess that means you would pick T4 T4 T4 in this case. But how would the user know that you picked wrong, and that it only was a "best guess"?

I'm also not 100% following in the related refactorings (but that might just be because I spent too little time looking into them).

PR #314, opened 2021-03-20
Scoreboard overhaul

P0keDev:
Adds a custom scoreboard overlay that replaces and mimics the vanilla scoreboard, with options to reposition, remove the numbers on the side, remove the title of the scoreboard, change the color/opacity of the background, and remove some unnecessary text from the default quest tracker.

This implementation also fixes the issues with the objective overlay. Instead of directly manipulating the scoreboard, which caused it to disappear when it shouldn't have, lines are individually removed while rendering, ensuring that everything else is always visible.

I also added a small change to the filter for the welcome messages, which now displays them if they contain info on market items being bought/sold.


magicus:
Cool! 🎉

magicus (approved):

HeyZeer0 (approved):
LGTM, nice work on this :D

PR #315, opened 2021-03-25
Add config option for disabling automatic update checks

phantamanta44:
I usually play on a custom build of Wynntils, and it's inconvenient having to click through an update screen every time I log into Wynncraft. This PR remedies the problem by adding a config option that disables automatic update checks.


HeyZeer0:
The auto-updater being annoying is a feature, I definitely don't want people to be stuck in an old version and also be able to disable the warnings.

If you are running a custom Wynntils build I recommend you start up the game with the environment variable `wynntils.development` set as `true` which will disable the updater entirely

PR #317, opened 2021-04-21
Make /lootrun list open the lootrun page

Incompleteusern:


Incompleteusern:
Only the last 2 matter, and how do I get rid of all the previous commits from appearing?

PR #320, opened 2021-04-23
Translation revamp

Incompleteusern:
Code should work, but I can't test it because gradle won't build correctly.

The error is because when directly translating, when a number is at the end, it is never added to the component.


Incompleteusern:
Currently a wip

Incompleteusern:
Making a new pr because this one is messy

PR #323, opened 2021-05-02
Don't show distance to compass marker on minimap while it is nearby.

magicus:
This is unnecessary and only clobbers the minimap view. Instead only show it when the marker is off the minimap.


HeyZeer0 (approved):

phantamanta44:
i think this should probably be configurable, e.g. with an option to choose between showing the distance:
* always
* only when far away
* never

magicus:
@phantamanta44 I understand your thinking, and that was my initial reaction as well, but then I thought about it and realized it never makes sense to show the distance when you see the marker on the map. If you are unsure of the distance when you see the marker, what you need is a scale on the minimap, not a distance marker.

phantamanta44:
i think it's rather presumptuous of you to dictate what users do and don't want. i prefer having the distance shown always, which is why i implemented it that way in the first place :I

magicus:
@phantamanta44 Okay, if you feel so strongly about it, I'll re-implement it as a tri-state logic.

PR #325, opened 2021-05-04
Translation Revamp

Incompleteusern:
This is a rewrite of the translation code and adding translation conditions, or what is necessary for a translation.
The 4 conditions are:
- **always** - Translate no matter what; no condition
- **discovery** - Translate if the corresponding discovery for the transcriber is discovered
-  **book** - Translate if the transcriber is in the inventory
- **never** - Don't translate ever; Disabled

image

If a condition isn't met the following hover text is shown

image

Another change is that Wynnic is colored green and Gavellian is colored purple.

imageimage


Incompleteusern:
Should be last commit until feedback/review.

Incompleteusern:
Should be last commit until feedback as the callback function is stable and everything else works.

src/main/java/com/wynntils/modules/questbook/overlays/ui/LootRunPage.java
```diff
@@ -361,7 +361,11 @@ public void mouseClicked(int mouseX, int mouseY, int mouseButton) throws IOExcep
                     if (result) {
                         try {
                             Location start = LootRunManager.getActivePath().getPoints().get(0);
-                            String startingPointMsg = "Loot run " + LootRunManager.getActivePathName() + " starts at [" + (int) start.getX() + ", " + (int) start.getZ() + "]";
+                            String message = TextFormatting.GREEN + "Loaded loot run " + LootRunManager.getActivePathName() + " successfully! " + TextFormatting.GRAY + "(" + LootRunManager.getActivePath().getChests().size() + " chests)";
+
+                            Minecraft.getMinecraft().player.sendMessage(new TextComponentString(message));
+
+                            String startingPointMsg = "Loot run starts at [" + (int) start.getX() + ", " + (int) start.getZ() + "]";
```
magicus, line 369: This seems like an unrelated change..? It looks good, but not like it belongs here

src/main/java/com/wynntils/modules/chat/overlays/ChatOverlay.java
```diff
@@ -272,11 +289,116 @@ private void updateLine(ChatTab tab, ITextComponent chatComponent, int updateCou
...
         }
     }
 
+    public void processQueue(ChatTab tab) {
+        updateLines(tab, tab.getQueue());
+    }
+
+    public void processQueues() {
+        for (ChatTab tab : TabManager.getAvailableTabs()) {
+            processQueue(tab);
```
magicus, line 305: I'm not sure I understand the queues fully, but my understanding is that if the quest book is not analyzed, you queue chat messages awaiting analysis to see if it should be translated. This is perhaps my only worry... 1) Are you sure questbook analysis cannot get stuck in some kind of loop/deadlock, where lack of chat feedback prevents it from finishing? And 2) this means that whenever you start playing the first chat message will be delayed until the quest book is analysed.

magicus: It might be okay, I just want to point this out. An alternative is to start background analysis if it's missing, but skip trying to translate messages until the analysis is done.

src/main/java/com/wynntils/modules/chat/configs/ChatConfig.java
```diff
@@ -62,30 +62,43 @@
     @Setting(displayName = "Alter Chat Tab by Presets", description = "Which premade selection of chat tabs should be used?\n\na - Global, Guild, Party\n\nb - Global, Shouts, Guild/Party, PMs\n\nvanilla - All", order = 0)
     public Presets preset = Presets.vanilla;
 
-    @Setting(displayName = "Clickable Party Invites", description = "Should party invites provide a clickable command?")
+    @Setting(displayName = "Wynnic Translation Condition", description = "What is the condition for translating Wynnic and Gavellian be translated?\n\n§8Discovery - The discovery for the transcriber has been discovered\n\n§8Book - The transcriber is in the player's inventory", order = 12)
+    public TranslateConditions translateCondition = TranslateConditions.always;
```
magicus, line 66: Should it be always as default? Maybe one of the more limiting choices (i'd vote for discoveries). Otherwise it's unlikely that anyone would ever actively limit this.

src/main/java/com/wynntils/modules/questbook/overlays/ui/LootRunPage.java
Incompleteusern: I can remove it if necessary, but It doesn't seem deserving of its own PR

src/main/java/com/wynntils/modules/chat/overlays/ChatOverlay.java
Incompleteusern: Its not queuing the message so much as queuing the change. It already starts background analysis and reprocesses the message later based on groupID

`
@SubscribeEvent
    public void onAnalyzeDiscoveries(QuestBookUpdateEvent.Partial e) {
        if (e.getAnalysed() == AnalysePosition.SECRET_DISCOVERIES && !ChatManager.getDiscoveriesLoaded()) {
            ChatManager.setDiscoveriesLoaded(true);
            ChatOverlay.getChat().processQueues();
        }
    }
`
    
`
if (QuestManager.getCurrentDiscoveries().isEmpty() && !discoveriesLoaded) {
    QuestManager.updateAnalysis(EnumSet.of(AnalysePosition.DISCOVERIES, AnalysePosition.SECRET_DISCOVERIES), true, true);
    return new Pair<>(original, new Pair<>(ChatManager::getDiscoveriesLoaded, s -> ChatManager.processRealMessage(s).a));
}
`

src/main/java/com/wynntils/modules/chat/configs/ChatConfig.java
Incompleteusern: I don't really know what to have other than that it shouldn't be never, and some people don't really do secret discoveries especially because both of these discoveries are ultimate. And true, people wouldn't really want to limit their choices unless for RP reasons.

src/main/java/com/wynntils/modules/questbook/overlays/ui/LootRunPage.java
magicus: Okay.

src/main/java/com/wynntils/modules/chat/configs/ChatConfig.java
magicus: I'm basically okay with this patch. If you think always is correct, go for it. Otherwise a quick check on discord to see if people agree..?

magicus (approved):

Incompleteusern:
@HeyZeer0 Can I have a review?

Incompleteusern:
Fixed conflicts with 1.16 preparations too

magicus:
I think this looks good now. I don't think we need to wait more for additional reviews.

PR #329, opened 2021-05-22
Fixed a bug with wynnic/gavellian translation

Incompleteusern:
Fixed a bug with translating made by me in earlier commit #328


Incompleteusern:
if more bugs are found from my prs ill add them here too

PR #330, opened 2021-05-31
Quest book list page

Incompleteusern:
- Made it so the scale of the map in LootRunPage is no longer a config and depends on player scroll
- Created a Quest book list page class, and had dialogue page, quests page, discovery page, item page, and lootrun page extend it, imo makes the pages more internally consistent
- Changed the layout of the MainPage so it's easier to know there's a second page, pretty sure some people don't even know it exists
  image

- Implemented a shadow for the DialoguePage, could be made configurable
  image


Incompleteusern:
Should be final commit

magicus (approved):
Looks like a good cleanup.

HeyZeer0:
Sorry for the delay, I think we should try to find a better solution to the background color rather than just adding a block box 🤔, have you tried shadows?

Incompleteusern:
As in the shadows on the quest/discovery pages?

HeyZeer0:
> As in the shadows on the quest/discovery pages?

No I mean shadows on the text you're writing into the page!

Incompleteusern:
image
Like this?

HeyZeer0:
> image
> Like this?


This looks better, what about the other shadow type? 🤔 
(also I definitely didn't edited your message instead of quoting it)

Incompleteusern:
image

Incompleteusern:
Last commit implemented the following

- Changed the layout of the MainPage so it's easier to know there's a second page, pretty sure some people don't even know it exists
  image

- Implemented the first shadow layout, could be made configurable
  image

PR #331, opened 2021-06-19
Add health percentage info variable

ryanzhoudev:
Allows user to determine if SSAC or Rage is active


P0keDev (changes requested):
Add it as a variable in the Variables enum in OverlayConfig and it should be good.

P0keDev (approved):

PR #333, opened 2021-07-11
Improve quest dialogue and quest book analysis

P0keDev:
**Quest Dialogue**
Instead of showing "x delayed messages" when receiving chat messages while in dialogue, the actual messages are displayed (in grey, as that is how they are sent by Wynn). Also fixes a bug where leaving a particular dialogue segment open for too long would result in /toggle chat spam once finished with the dialogue.
image

**Quest Book Analysis**
Instead of analyzing the quest book after every piece of dialogue, which is unnecessary and potentially disruptive, Wynn's quest autotracking system is used to determine when the quest book should be analyzed. The scoreboard text is hidden through the scoreboard overlay if this setting is enabled. The entire system can be disabled if the user only wants to update the quest book manually (or never use the Wynntils quest book). Wynn's autotracking will automatically enable/disable based on the respective settings.

Also includes some miscellaneous improvements I noticed.


magicus:
Thank you @P0keDev for looking into this! This seems to be solving lot of the "papercut" issues with the new Wynncraft version. (I have not looked at the code, just the description)

magicus (commented):
Now I've looked at the code as well :-) Generally looks good, as always, and it seems to solve the problems stated in the PR.

However, my one question is about the changes regarding to the Powder manual. Have these disappeared in the latest Wynncraft? Or are they filtered out some other way?

P0keDev:
> However, my one question is about the changes regarding to the Powder manual. Have these disappeared in the latest Wynncraft? Or are they filtered out some other way?

The in-game powder manual was completely removed a little while ago, so I figured there wasn't any reason to keep the old code around

HeyZeer0 (approved):
Really cool, thanks for solving these issues 🎉 🎉 
Also do you mind taking a look on what is making a bunch of pop pop sfx when the analysis finishes? Not sure who placed that but it sounds weird

Incompleteusern:
Are you referring to https://github.com/Wynntils/Wynntils/blob/85bc1f2aecf00cf86f3fb029e6d9fd2eaf438cdd/src/main/java/com/wynntils/modules/questbook/managers/QuestManager.java#L316 

There's also some deprecated features such as Reference.onNether and unprocessed material stats in the InfoOverlay

P0keDev:
That sound file doesn't have any pop sfx - it's being caused by something else

PR #334, opened 2021-07-13
Added a Seaskipper Map replacing the normal gui

Incompleteusern:
Creates a Seaskipper map, which replaces the normal Seaskipper gui. It can be disabled by config.

Images [here](https://github.com/Wynntils/Wynntils/pull/334#issuecomment-882127973)

It has 3 settings
- Whether to display territory borders
- Whether to display inaccessible destinations
- Whether to display routes to accessible destinations

Fixes a bug where guild territory names wouldn't be rendered when position is being updated. 

Replaces the trade route fix by P0ke a while ago too.

Names 2 methods that canceled packets, see https://github.com/Wynntils/Wynntils/commit/0a255158417762fcf1fd922bf14fdab1a80ce533

Api part of commit at
https://github.com/Wynntils/WynntilsWebsite-API/pull/2


Incompleteusern:
@HeyZeer0 Can I have a review?

magicus:
I have not looked at the code, just the description. From the images, it looks like you might need to zoom out, or pan around, in the map. That'd make for a much worse UI experience than the current "vanilla" Wynncraft interface. 

Can you make sure that all destinations fit in a single screen? I know the current map cannot be zoomed out to allow this (why this is so I don't know, but it is annoying, so part of the fix can certainly be so that this map too can be zoomed out to max). And you'll need to store the zoom level differently between the Seaskipper map and the main map.

HeyZeer0:
Yeah, there are some design heuristics that bothers me. So I'll just embed a couple of suggestions:

1. You can remove the square and text from inaccessible regions by default.
2. Having the text on the main map area sounds a bit weird, maybe you could use the same popup thing as I do on the guild map?
3. I don't think the connection lines are necessary and might confuse some people (and it's over the player position cursor)
4. You don't need the square on your origin position since you already have the position cursor.

Rather than that, looks really good! Nice work on this and sorry for the delay :D

Incompleteusern:
> I have not looked at the code, just the description. From the images, it looks like you might need to zoom out, or pan around, in the map. That'd make for a much worse UI experience than the current "vanilla" Wynncraft interface.
> 
> Can you make sure that all destinations fit in a single screen? I know the current map cannot be zoomed out to allow this (why this is so I don't know, but it is annoying, so part of the fix can certainly be so that this map too can be zoomed out to max). And you'll need to store the zoom level differently between the Seaskipper map and the main map.

For the vanilla gui, you have to hover over all the tags, which could be more annoying for places like Nemract. There's also the potential to miss click on the wrong tag. I'll increase zoom then

Halfway through implementation, https://github.com/Wynntils/Wynntils/pull/334/commits/a6a5bdea2c916b49ab2ba502bbd0ce7b830307a2
is untested

Incompleteusern:
> Yeah, there are some design heuristics that bothers me. So I'll just embed a couple of suggestions:
> 
> 1. You can remove the square and text from inaccessible regions by default.
> 2. Having the text on the main map area sounds a bit weird, maybe you could use the same popup thing as I do on the guild map?
> 3. I don't think the connection lines are necessary and might confuse some people (and it's over the player position cursor)
> 4. You don't need the square on your origin position since you already have the position cursor.
> 
> Rather than that, looks really good! Nice work on this and sorry for the delay :D

Currently, implementation for each is
1. Turned off setting by default
2. Implemented
3. Turned off setting by default
4. Removed pointer

Also increased max zoom for the maps

Incompleteusern:
Default: 
image
Routes:
image
Inaccessible:
image

Incompleteusern:
Found a bug

Incompleteusern:
TODO: redo regions

magicus:
The new maps look really good! Thank you for the work on this.

src/main/java/com/wynntils/modules/core/events/ClientEvents.java
```diff
@@ -143,23 +143,23 @@ public void workAroundWynncraftNPEBug(PacketEvent<SPacketTeams> e) {
     }
 
     @SubscribeEvent
-    public void cancelSomeVelocity(PacketEvent<SPacketEntityVelocity> e) {
-        // I'm not sure what this does, but the code has been here a long time,
-        // just moving it here. /magicus, 2021
+    public void cancelBoatSinking(PacketEvent<SPacketEntityVelocity> e) {
+        // Prevents boats from sinking like a submarine
+        //from https://forums.wynncraft.com/threads/wynn-extra-fix-mod-for-1-11-2-boat-finally-not-function-as-submarine.227231/
```
magicus, line 148: Ah, so that's where it's from! Thanks for the added documentation! :-)

xanndra2004:
Hi Incompleteusern! I've compiled the build and tested it out, and I can't seem to get the new Seaskipper map working.

Incompleteusern:
Put the json file at https://github.com/Incompleteusern/WynntilsWebsite-API/blob/master/seaskipper.json in .minecraft/wynntils/apicache

xanndra2004:
Ah, thank you! Works now!

Incompleteusern:
can't rename with an open pr

HeyZeer0 (approved):
LGTM, nice work on this the changes are perfect, sorry for the delay in reviewing!

PR #335, opened 2021-07-18
Hoverable item tooltips in chat

P0keDev:
Adds an option to copy an item as a string to be put in a chat message after pressing the item screenshot button. This string will automatically be converted into the color-coded and underlined item name, which can be moused over to view the item's exact stats, powders, and rerolls. Multiple items can be put in one message and still be properly viewed. When the string is pasted into the chat box, it will be replaced with a simple shorthand (`<ItemName>`) until the message is actually sent, to avoid any confusion due to the invisible characters used. Only regular gear items are supported: no crafted items or any non-gear item like powders, ingredients, etc.

The chat tooltips also contain an extra line specifying that it is this type of tooltip - a dark gray "_From chat_" below the item's name. My motivation for adding this was to combat fake items being spread. If this is not deemed a significant enough issue, I can easily remove it.

The "encoded" item strings are created using Unicode characters in the private use block; the exact implementation can be found in a comment at the top of ChatItemManager. To other players without Wynntils (or with older versions of Wynntils) in 1.12, these characters will not be rendered, and all they will see is the item's name. However, in newer versions of Minecraft, they will be seen as rectangles. In order to avoid spamming the chats of players on these versions, I attempted to make the encoding as efficient and short as possible, while retaining all the necessary information. Below are some examples of encoded strings, with the outliers Singularity (15 powder slots) and Melange (15 identifications) demonstrating some of the longest possible encodings.
```
󵿰Fate's Shear󵿲󵄽󵅹󵂰󵀀󵀵󵀭󵀼󵀜󵿲󵁵󵀀󵿱
󵿰Singularity󵿲󵃴󵀤󵃀󵀜󵄹󵁠󵿲󵏾󵈆󵈆󵁖󵀂󵿱
󵿰Melange󵿲󵀌󵀈󵀈󵀌󵀐󵀞󵀌󵀄󵀀󵀙󵀈󵀕󵀐󵀌󵀄󵀀󵿱
󵿰Suppression󵿲󵀍󵃬󵀊󵀈󵀀󵿱
󵿰Unravel󵿲󵁐󵆓󵀊󵀸󵀽󵀀󵿱
```
I did write code to encode the item's name as well, but I decided it would be best to at least leave some part of it intelligible, so even if someone can't view the tooltip, they at least know what is going on. However, if desired, I could enable the encoding of names as well.

image
image
image

### DIFF
--- a/src/main/java/com/wynntils/modules/chat/managers/ChatManager.java
+++ b/src/main/java/com/wynntils/modules/chat/managers/ChatManager.java
@@ -331,26 +331,32 @@ public class ChatManager {
                     continue;
                 }
 
-                String text = McIf.getUnformattedText(comp);
-                Style style = comp.getStyle();
+                do {
 
-                ITextComponent item = ChatItemManager.decodeItem(m.group());
-                if (item == null) { // couldn't decode, skip
-                    ITextComponent newComponent = new TextComponentString(comp.getUnformattedComponentText());
-                    newComponent.setStyle(comp.getStyle().createShallowCopy());
-                    temp.appendSibling(newComponent);
-                    continue;
-                }
+                    String text = McIf.getUnformattedText(comp);
+                    Style style = comp.getStyle();
 
-                ITextComponent preText = new TextComponentString(text.substring(0, m.start()));
-                preText.setStyle(style.createShallowCopy());
-                temp.appendSibling(preText);
+                    ITextComponent item = ChatItemManager.decodeItem(m.group());
+                    if (item == null) { // couldn't decode, skip
+                        comp = new TextComponentString(comp.getUnformattedComponentText());
+                        comp.setStyle(style.createShallowCopy());
+                        continue;
+                    }
 
-                temp.appendSibling(item);
+                    ITextComponent preText = new TextComponentString(text.substring(0, m.start()));
+                    preText.setStyle(style.createShallowCopy());
+                    temp.appendSibling(preText);
 
-                ITextComponent postText = new TextComponentString(text.substring(m.end()));
-                postText.setStyle(style.createShallowCopy());
-                temp.appendSibling(postText);
+                    temp.appendSibling(item);
+
+                    comp = new TextComponentString(text.substring(m.end()));
+                    comp.setStyle(style.createShallowCopy());
+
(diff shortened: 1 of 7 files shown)


magicus (approved):
Looks good. Great idea! 

I really do miss a working testing framework in Wynntils, though... :( The encoding/decoding process would have been perfect to have as a unit test.

src/main/java/com/wynntils/modules/chat/managers/ChatManager.java
```diff
@@ -316,6 +317,50 @@
             in = temp;
         }
 
+        // chat item tooltips
+        if (ChatItemManager.ENCODED_PATTERN.matcher(McIf.getUnformattedText(in)).find()) {
+
```
magicus: The extra linebreaks here feels a bit off; I don't think that is the usual Wynntils style

src/main/java/com/wynntils/modules/utilities/managers/ChatItemManager.java
```diff
@@ -0,0 +1,393 @@
+package com.wynntils.modules.utilities.managers;
```
magicus, line 5: The copyright header is missing. See any other file.

magicus (commented):
Just a thought: is it possible to send the HoverEvent over the chat channel, so you can skip the encoding process? I'm not sure what minecraft filters out, but apparently some actions are possible to send; Wynncraft itself sends these "Click here to say thanks to GenerousPlayer123!" links...

coehlrich:
For chat Client's can only send text to the server while the server send's text components to client's and in newer versions of Minecraft any characters which don't have a texture get rendered as an empty white square including the characters used by this.

magicus:
@coehlrich Ah, that makes sense.

src/main/java/com/wynntils/modules/utilities/managers/ChatItemManager.java
P0keDev: Whoops, thanks

HeyZeer0 (approved):
Very well done! 
Compressing all the item info in such a small text is incredible, nice work on this!

Maybe for the future, we could create a standard method to create the item lore, I think this there's 3 different places with the same thing

PR #340, opened 2021-08-08
Fixed typo with the license badge

Kirdow:
The license badge below the logo showed ``AGLP 3.0`` instead of the correct ``AGPL 3.0``.


HeyZeer0:
Nice catch! 😄

PR #343, opened 2021-08-25
Fixed translator returning empty messages if failed

theobeaudenon:
Sometime Google display an error when calling too often the url. Prevent displaying a blank text

### DIFF
--- a/src/main/java/com/wynntils/webapi/services/GoogleApiTranslationService.java
+++ b/src/main/java/com/wynntils/webapi/services/GoogleApiTranslationService.java
@@ -6,6 +6,7 @@ package com.wynntils.webapi.services;
 
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
+import com.google.gson.JsonParser;
 import com.wynntils.core.utils.Utils;
 import com.wynntils.webapi.request.Request;
 import com.wynntils.webapi.request.RequestHandler;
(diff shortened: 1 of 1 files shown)


src/main/java/com/wynntils/webapi/services/GoogleApiTranslationService.java
```diff
@@ -39,6 +39,11 @@ protected void translateNew(String message, String toLanguage, Consumer<String>
                     String translatedMessage = builder.toString();
                     saveTranslation(toLanguage, message, translatedMessage);
                     handleTranslation.accept(translatedMessage);
+
+                    return true;
+                }).onError(integer -> {
+                    //If Google trad return no data ( 500 error ) , display default lang
+                    handleTranslation.accept(message);
```
magicus, line 46: This will make it look like the original text was the translation, so it will be printed twice. 

Why not just return false?

magicus:
Does this happen for specific strings? Or is it a timeout/request overflow thing?

theobeaudenon:
> Does this happen for specific strings? Or is it a timeout/request overflow thing?

It's a request overflow thing caused the Google “unusual traffic” message.

> This will make it look like the original text was the translation, so it will be printed twice.

Nope, the message is only printed once in the original message 

> Why not just return false?

Maybe we could, I need to test this possibility. But I dont think that the message get displayed

theobeaudenon:
> > Why not just return false?
> 
> Maybe we could, I need to test this possibility. But I dont think that the message get displayed

return false does not work. Display Empty message

[](https://postimg.cc/S23CrVFK)

HeyZeer0 (approved):
Since you checked the return, LGTM. Nice work! Lemme know your discord username so I can give you the contributor tag 😄 

Thanks for the help!

theobeaudenon:
Thèor#0955

PR #344, opened 2021-08-26
Fixed Champion Rank not being handled properly

1uf3:
I want CHAMPION RANK in wynntils, so i add champion rank


HeyZeer0 (approved):
Nice catch! Thanks for the help! 😄

PR #345, opened 2021-08-28
Item favoriting

P0keDev:
Adds a system for 'favoriting' items so that they can more easily be noticed in loot chests. Users can favorite and unfavorite items through the item guide by shift left clicking, which is noted in each item's tooltip similar to the Wynndata feature. Any boxes that might contain a favorited item have a star icon to signify the fact, and the name of the item(s) in question will be underlined in the possibilities line of the tooltip, if item possibilities are enabled. Wynntils will also prevent the closing of chests with such boxes, in the same way it does with mythic boxes (this feature can be toggled off). Currently, only regular gear items are supported (basically, anything that shows up in the item guide) - I do want to expand the feature to ingredients eventually.

This PR also fixes a crash that occurs when extremely large numbers are entered as prices in the trade market, and disables clicking the ingredient pouch while in a chest, since it does nothing but force close the chest.

image
image
image

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
+++ b/src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
@@ -279,7 +279,7 @@ public class ClientEvents implements Listener {
         if (!priceInput) return;
 
         priceInput = false;
-        int price = StringUtils.convertEmeraldPrice(e.getMessage());
+        long price = StringUtils.convertEmeraldPrice(e.getMessage());
         if (price != 0) // price of 0 means either garbage input or actual 0, can be ignored either way
             e.setMessage("" + price);
     }
@@ -666,6 +666,11 @@ public class ClientEvents implements Listener {
 
     @SubscribeEvent
     public void clickOnChest(GuiOverlapEvent.ChestOverlap.HandleMouseClick e) {
+        if (e.getSlotIn() != null && e.getSlotId() - e.getGui().getLowerInv().getSizeInventory() == 4) { // prevent accidental pouch clicking
+            e.setCanceled(true);
+            return;
+        }
+
         if (UtilitiesConfig.INSTANCE.preventSlotClicking && e.getSlotIn() != null) {
             if (e.getSlotId() - e.getGui().getLowerInv().getSizeInventory() >= 0 && e.getSlotId() - e.getGui().getLowerInv().getSizeInventory() < 27) {
                 e.setCanceled(checkDropState(e.getSlotId() - e.getGui().getLowerInv().getSizeInventory() + 9, McIf.mc().gameSettings.keyBindDrop.getKeyCode()));
--- a/src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
+++ b/src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
@@ -104,7 +104,7 @@ public class ClientEvents implements Listener {
     }
 
     @SubscribeEvent
-    public void onKeyboardEven(InputEvent.KeyInputEvent e) {
+    public void onKeyboardEvent(InputEvent.KeyInputEvent e) {
         long currentTime = System.currentTimeMillis();
         // Events triggered just after the user pressed the Toggle AFK Protection key
         // should be ignored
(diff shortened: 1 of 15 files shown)


src/main/java/com/wynntils/modules/utilities/overlays/inventories/FavoriteItemsOverlay.java
```diff
@@ -0,0 +1,103 @@
...
+
+        for (Slot s : gui.inventorySlots.inventorySlots) {
+            ItemStack stack = s.getStack();
+            if (stack.isEmpty() || !stack.hasDisplayName() || !stack.hasTagCompound()) continue;
+
+            NBTTagCompound nbt = stack.getTagCompound();
+            if (nbt.hasKey("wynntilsFavorite")) continue; // already been checked
+
+            String itemName = StringUtils.normalizeBadString(getTextWithoutFormattingCodes(stack.getDisplayName()));
+            if (!itemName.contains("Unidentified")) {
```
Incompleteusern, line 62: I don't think this should be inverted

src/main/java/com/wynntils/modules/utilities/overlays/inventories/FavoriteItemsOverlay.java
```diff
@@ -0,0 +1,103 @@
...
+            ItemStack stack = s.getStack();
+            if (stack.isEmpty() || !stack.hasDisplayName() || !stack.hasTagCompound()) continue;
+            NBTTagCompound nbt = stack.getTagCompound();
+            if (!nbt.getBoolean("wynntilsFavorite")) continue;
+
+            // draw star
+            ScreenRenderer.beginGL(0, 0);
+            GlStateManager.translate(0, 0, 300);
+
+            ScreenRenderer r = new ScreenRenderer();
```
HeyZeer0: This will cause some heavy GC, don't we have a static instance of this stored somewhere? If not just create one on this class and it should be fine.

src/main/java/com/wynntils/modules/utilities/overlays/inventories/FavoriteItemsOverlay.java
```diff
@@ -0,0 +1,103 @@
...
+            RenderHelper.disableStandardItemLighting();
+            float scaleFactor = 0.5f;
+            ScreenRenderer.scale(scaleFactor);
```
HeyZeer0: Declaring the variable here seems unnecessary? Probably could be
```suggestion
            ScreenRenderer.scale(0.5f);
```

src/main/java/com/wynntils/modules/utilities/overlays/inventories/FavoriteItemsOverlay.java
```diff
@@ -0,0 +1,103 @@
...
+            r.drawRect(Textures.Map.map_icons, (int)((gui.getGuiLeft() + s.xPos + 10) / scaleFactor), (int)((gui.getGuiTop() + s.yPos - 5) / scaleFactor), 208, 36, 18, 18);
```
HeyZeer0: Instead of translating into 0, 0 with beginGl, translate to `gui.getGuiLeft()` and `gui.getGuiTop()` including the offset to the center of the slot, that will make you not need to divide by the scale factor reducing overhead and improving aliasing.

src/main/java/com/wynntils/modules/utilities/overlays/inventories/FavoriteItemsOverlay.java
```diff
@@ -0,0 +1,103 @@
...
+                if (itemProfile == null) continue;
+
+                if (itemProfile.isFavorited()) {
+                    nbt.setBoolean("wynntilsFavorite", true);
+                    break;
+                }
+            }
+
+            if (!nbt.hasKey("wynntilsFavorite")) // wasn't favorited
+                nbt.setBoolean("wynntilsFavorite", false);
```
HeyZeer0, line 81: Any reason why you are setting it to false instead of checking if it contains? This will probably add a bunch of unnecessary data but might have a reason

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
```diff
@@ -659,6 +666,11 @@ public void handleAccessoryMovement(TickEvent.ClientTickEvent e) {
 
     @SubscribeEvent
     public void clickOnChest(GuiOverlapEvent.ChestOverlap.HandleMouseClick e) {
+        if (e.getSlotIn() != null && e.getSlotId() - e.getGui().getLowerInv().getSizeInventory() == 4) { // prevent accidental pouch clicking
+            e.setCanceled(true);
+            return;
+        }
```
HeyZeer0, line 672: Does this block clicks on the pouch with the pouch open? If so might wanna check if that doesn't cause any weird behavior

HeyZeer0 (changes requested):
LGTM once changes were made. Really nice work on this! Definitely a cool feature and thanks for the bug fixes 😄

src/main/java/com/wynntils/modules/utilities/overlays/inventories/FavoriteItemsOverlay.java
P0keDev: I did it so that any item only needs to be checked once, and then can be ignored on every tick afterwards (line 59) - I figured that would be more efficient

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
P0keDev: Clicking the pouch while in the pouch menu force closes it in the same way as any other chest gui, so disabling it shouldn't be an issue

HeyZeer0 (approved):

PR #341, opened 2021-08-10
Add a boat button to seaskipper map

Incompleteusern:

PR #346, opened 2021-08-28
Remove return

Incompleteusern:
Don't know how this got in there

### DIFF
--- a/src/main/java/com/wynntils/modules/map/overlays/ui/SeaskipperWorldMapUI.java
+++ b/src/main/java/com/wynntils/modules/map/overlays/ui/SeaskipperWorldMapUI.java
@@ -154,7 +154,6 @@ public class SeaskipperWorldMapUI extends WorldMapUI {
                 location.setCost(cost);
                 location.setActiveType(SeaskipperLocation.Accessibility.ACCESSIBLE);
                 routes.put(location, s.slotNumber);
-                return;
             }
 
             if (stack.getItem() == Items.BOAT) { //Buy boat


src/main/java/com/wynntils/modules/map/overlays/ui/SeaskipperWorldMapUI.java
```diff
@@ -154,7 +154,6 @@ private void updateItems() {
                 location.setCost(cost);
                 location.setActiveType(SeaskipperLocation.Accessibility.ACCESSIBLE);
                 routes.put(location, s.slotNumber);
-                return;
```
HeyZeer0, line 157: Should be a continue no?

HeyZeer0 (changes requested):

src/main/java/com/wynntils/modules/map/overlays/ui/SeaskipperWorldMapUI.java
HighCrit: Or else if with next statement

Incompleteusern: It doesn't matter if its a continue statement, I like at as an else if but https://github.com/Wynntils/Wynntils/pull/341#discussion_r693610062

Incompleteusern: It's just important that this bug gets fixed, fine either way

Incompleteusern:
Fixed in https://github.com/Wynntils/Wynntils/commit/863e7d97cd2d79cb03eea8d4aafb8f799d52e03d

PR #351, opened 2021-10-05
Emerald pouch hotkey option

ryanzhoudev:
Code is similar to ingredient pouch when the emerald pouch is in the main inventory
Function is unbound by default.

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/managers/KeyManager.java
+++ b/src/main/java/com/wynntils/modules/utilities/managers/KeyManager.java
@@ -13,14 +13,20 @@ import com.wynntils.modules.core.CoreModule;
 import com.wynntils.modules.map.overlays.MiniMapOverlay;
 import com.wynntils.modules.utilities.UtilitiesModule;
 import com.wynntils.modules.utilities.events.ClientEvents;
+import com.wynntils.modules.utilities.overlays.hud.GameUpdateOverlay;
 import com.wynntils.modules.utilities.overlays.hud.StopWatchOverlay;
 import com.wynntils.modules.utilities.overlays.ui.GearViewerUI;
 import com.wynntils.webapi.WebManager;
 import net.minecraft.client.entity.EntityPlayerSP;
 import net.minecraft.inventory.ClickType;
+import net.minecraft.item.ItemStack;
 import net.minecraft.network.play.client.CPacketClickWindow;
+import net.minecraft.util.NonNullList;
+import net.minecraft.util.text.TextFormatting;
 import net.minecraftforge.client.settings.KeyConflictContext;
 import org.lwjgl.input.Keyboard;
+import java.util.ArrayList;
+import java.util.List;
 
 public class KeyManager {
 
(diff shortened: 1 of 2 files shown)


Incompleteusern:
Try checking the questbook analysis code for right clicking in the hotbar

HeyZeer0:
Probably the reason you can't click on the Hotbar is because you are using the incorrect slot id, the packet requires the raw slot and the Hotbar is not 0-8 on raw slots

Check this out:
Inventory Slots

src/main/java/com/wynntils/modules/utilities/managers/KeyManager.java
```diff
@@ -109,19 +109,43 @@ public static void registerKeys() {
...
-                       // Instead, use the same method we use to spawn a horse
-                       int prev = McIf.player().inventory.currentItem;
-                       McIf.player().inventory.currentItem = emeraldPouchSlots.get(0);
-                       McIf.mc().playerController.processRightClick(McIf.player(), McIf.player().world, EnumHand.MAIN_HAND);
-                       McIf.player().inventory.currentItem = prev;
+                   int slotNumber = emeraldPouchSlots.get(0);
+                   if (slotNumber < 9) {
+                       // sendPacket uses raw slot numbers, we need to manually remap the hotbar
+                       switch (emeraldPouchSlots.get(0)) {
+                           case 0:
```
magicus: ...! 😮 

So, you are "unrolling" the calculation `slotNumber + 36`? :-D

src/main/java/com/wynntils/modules/utilities/managers/KeyManager.java
```diff
@@ -83,6 +90,37 @@ public static void registerKeys() {
...
+
+           for (int i = 0; i < inventory.size(); i++) {
+               ItemStack stack = inventory.get(i);
+               if (!stack.isEmpty() && stack.hasDisplayName() && stack.getDisplayName().contains("§aEmerald Pouch§2 [Tier")) { // Match as much as possible of the emerald pouch name to prevent false positives
+                   emeraldPouchSlots.add(i);
+               }
+           }
+           switch (emeraldPouchSlots.size()) {
+               default:
+                   GameUpdateOverlay.queueMessage(TextFormatting.DARK_RED + "You have more than one emerald pouch in your inventory.");
```
magicus: Just thinking out loud here. Maybe it would be better to open one of the pouches, if we have many? Like the first, or the last, or the one with most/least emeralds/free space in them?

Otoh, I have not really played with the new emerald pouches, so I can't say I really understand the need for this hotkey anyway. :-)

magicus (approved):
The code looks good now imho.

src/main/java/com/wynntils/modules/utilities/managers/KeyManager.java
ryanzhoudev: I was originally considering something like that as well, but I thought it'd be confusing at times and imo it's better to avoid it when dealing with currency and that sort of stuff

If anyone disagrees and wants it in though, I can work something out

magicus: This is good for me. Leave it as it is.

P0keDev (changes requested):
The pouch code (for both pouches) should be moved to a different file like for all the other hotkeys, to avoid cluttering this file

magicus (approved):

P0keDev (approved):

PR #352, opened 2021-10-06
Add Emerald Pouch Usage Arc

ryanzhoudev:
Usage arc similar to the crafted item durability arc
- Starts at green when empty
- Progresses to red when getting full
- Setting to toggle on/off


magicus (approved):

magicus:
Cool idea. I wonder if we could/should show the ingredient pouch "fullness" with the same kind of functionality? (Not saying you should do that in this PR)

src/main/java/com/wynntils/modules/utilities/overlays/inventories/RarityColorOverlay.java
```diff
@@ -190,6 +196,32 @@ private static CustomColor getHighlightColor(Slot s, ItemStack is, String lore,
...
+
+        float capacity = Float.parseFloat(capacity_m.group(1)) * 64;
+
+        switch (capacity_m.group(2)) {
+            case LE:
+                capacity *= 64;
+            case "stx":
+                capacity *= 4096;
+        }
+        return usage / capacity;
```
HeyZeer0: Space here!
```suggestion
        }
        
        return usage / capacity;
```

src/main/java/com/wynntils/modules/utilities/overlays/inventories/RarityColorOverlay.java
```diff
@@ -190,6 +196,32 @@ private static CustomColor getHighlightColor(Slot s, ItemStack is, String lore,
         return null;
     }
 
+    private static float getPouchUsage(String lore) {
+        Matcher usage_m = POUCH_USAGE_PATTERN.matcher(lore);
+        Matcher capacity_m = POUCH_CAPACITY_PATTERN.matcher(lore);
```
HeyZeer0: Better name these variables, they don't follow the camelCase pattern we use and don't really explain what they really are, something like ``usageMatcher`` would work better! 😄

src/main/java/com/wynntils/modules/utilities/overlays/inventories/RarityColorOverlay.java
```diff
@@ -190,6 +196,32 @@ private static CustomColor getHighlightColor(Slot s, ItemStack is, String lore,
...
+        if (!capacity_m.find() || !usage_m.find()) return -1;
+
+        switch (usage_m.groupCount()) {
+            case 1:
+                usage = Float.parseFloat(usage_m.group(1));
+            case 2:
+                usage = Float.parseFloat(usage_m.group(1) + usage_m.group(2));
+            case 3:
+                usage = Float.parseFloat(usage_m.group(1) + usage_m.group(2) + usage_m.group(3));
+        }
```
HeyZeer0: Okay, this adds some limitations to the code and make it harder to read, took me like 3 minutes to understand that you were grouping together the number format splits, you can generify your regex to match them all together then replace all spaces in the group with an empty character and then convert it to a number.

src/main/java/com/wynntils/modules/utilities/overlays/inventories/RarityColorOverlay.java
```diff
@@ -190,6 +196,32 @@ private static CustomColor getHighlightColor(Slot s, ItemStack is, String lore,
...
+
+        float capacity = Float.parseFloat(capacity_m.group(1)) * 64;
```
HeyZeer0: Emeralds are integers not floats

src/main/java/com/wynntils/modules/utilities/overlays/inventories/RarityColorOverlay.java
```diff
@@ -199,6 +231,42 @@ private static float getDurability(String lore) {
...
+        // Obviously, when the pouch is full, we want to turn the colour more red, and the hue scale
+        // does this perfectly - green is 120', red is 0'.
+        // However, (MathHelper.hsvToRGB) takes a float input from 0-1, which means green is 0.33333, red is still 0.
+        // Since we don't like manually typing out infinitely repeating decimals, we can do (120 * inverse_usage) / 360.
+        // In this case, when usage is 1 (100%), it'll be 120 * 0 = 0, 0 / 360 = 0 (red), correct
+        // When usage is 0 (0%), it'll be 120 * 1 = 120, 120 / 360 = 0.333333 (green), correct
+        // When usage is 0.5 (50%), it'll be 120 * 0.5 = 60, 60 / 360 = 0.166666 (yellow), correct
+        float inverse_usage = 1 - usage;
+        int arcColor = MathHelper.hsvToRGB(120 * inverse_usage / 360.0F, 1.0F, 1.0F);
+        int radius = (UtilitiesConfig.Items.INSTANCE.itemLevelArc) ? 7 : 8;
```
HeyZeer0: Move this out of the rendering block, place it together with int x and y so we don't pollute the rendering code. Nice documentation tho, although you could move it to the method documentation since it is a bit long to make it better structured.

src/main/java/com/wynntils/modules/utilities/overlays/inventories/RarityColorOverlay.java
```diff
@@ -199,6 +231,42 @@ private static float getDurability(String lore) {
...
+        drawArc(bufferbuilder, x, y, usage, radius, arcColor >> 16 & 255, arcColor >> 8 & 255, arcColor & 255, 160);
+
+        GlStateManager.disableBlend();
+        GlStateManager.enableAlpha();
+        GlStateManager.enableTexture2D();
+        GlStateManager.enableDepth();
+        GlStateManager.enableLighting();
```
HeyZeer0: OpenGL works with rendering stacks and Minecraft on 1.12 doesn't really accepts attribute stacks so make sure these are not breaking the render of something else that is on the same matrix stack

src/main/java/com/wynntils/modules/utilities/overlays/inventories/RarityColorOverlay.java
```diff
@@ -47,6 +47,11 @@
     public static final float MAX_LEVEL = 105.0f;
     public static final float MAX_CIRCLE_STEPS = 16.0f;
     private static final Pattern DURABILITY_PATTERN = Pattern.compile("\\[([0-9]+)/([0-9]+) Durability\\]");
+    // Shorten! - Also we're not using the EmeraldSymbols in reference because that's not a constant
+    private static final String EB = "\u00B2\u00BD";
+    private static final String LE = "\u00BC\u00B2";
```
HeyZeer0: Uh, use the emerald symbols anyways, just replace the switch with an if since it is 2 conditions.

HeyZeer0 (changes requested):
Nice additions, some changes are required, nice work on this!

src/main/java/com/wynntils/modules/utilities/overlays/inventories/RarityColorOverlay.java
ryanzhoudev: I actually have no idea what these do at all, I just copied the arc rendering from the crafted durability function

ryanzhoudev: It looks like removing all of it breaks the arc rendering though

src/main/java/com/wynntils/modules/utilities/overlays/inventories/RarityColorOverlay.java
```diff
@@ -199,6 +231,42 @@ private static float getDurability(String lore) {
...
+        // So (float usage) returns a number from 0-1 representing the percentage used of the emerald pouch
+        // Obviously, when the pouch is full, we want to turn the colour more red, and the hue scale
+        // does this perfectly - green is 120', red is 0'.
+        // However, (MathHelper.hsvToRGB) takes a float input from 0-1, which means green is 0.33333, red is still 0.
+        // Since we don't like manually typing out infinitely repeating decimals, we can do (120 * inverse_usage) / 360.
+        // In this case, when usage is 1 (100%), it'll be 120 * 0 = 0, 0 / 360 = 0 (red), correct
+        // When usage is 0 (0%), it'll be 120 * 1 = 120, 120 / 360 = 0.333333 (green), correct
+        // When usage is 0.5 (50%), it'll be 120 * 0.5 = 60, 60 / 360 = 0.166666 (yellow), correct
+        float inverse_usage = 1 - usage;
+        int arcColor = MathHelper.hsvToRGB(120 * inverse_usage / 360.0F, 1.0F, 1.0F);
```
Incompleteusern: Inverse_usage could be inlined, as it is only used once.
Is there a reason 120 * inverse_usage / 360.0F is used instead of inverse_usage / 3.0F?

Nice work!

ryanzhoudev: Nope, I just directly threw my thought process into there, fixed

src/main/java/com/wynntils/modules/utilities/overlays/inventories/RarityColorOverlay.java
ryanzhoudev: After using this quite a bit, it looks like this doesn't break anything in-game, at least on my machine

ryanzhoudev:
Can't merge until [this](https://discord.com/channels/394189072635133952/743928502314598511/908050918602342451) is fixed

PR #353, opened 2021-10-06
Tier overlay sync

ryanzhoudev:
### Tier Overlay Sync

- Previously, Corkian Amplifiers and elemental powders did not have the option to show their tiers when the "Show Item Levels" hotkey was pressed. Added.
- Added emerald pouch specification overlay as well as "Show Item Levels" hotkey
- Cleaned up roman numeral converting in ItemLevelOverlay.java
- Added a slider to allow users to adjust size of tier specifications. Currently used for powders, emerald pouches, and amplifiers.
- Combined all "Show Item Levels" tier overlays into one setting.

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/overlays/inventories/ItemLevelOverlay.java
+++ b/src/main/java/com/wynntils/modules/utilities/overlays/inventories/ItemLevelOverlay.java
@@ -23,8 +23,8 @@ import java.util.regex.Pattern;
 
 public class ItemLevelOverlay implements Listener {
 
-    private static final Pattern POWDER_NAME_PATTERN = Pattern.compile("(?:Earth|Thunder|Water|Fire|Air|Blank) Powder (VI|IV|V|I{1,3})");
-    private static final Pattern EMERALD_POUCH_PATTERN = Pattern.compile("§aEmerald Pouch§2 \\[Tier (IX|X|VI{1,3}|IV|V|I{1,3})]");
+    private static final Pattern POWDER_NAME_PATTERN = Pattern.compile("(?:Earth|Thunder|Water|Fire|Air|Blank) Powder ([IV]{1,3})");
+    private static final Pattern EMERALD_POUCH_PATTERN = Pattern.compile("§aEmerald Pouch§2 \\[Tier ([XIV]{1,4})]");
     private static final Pattern CORKIAN_AMPLIFIER_PATTERN = Pattern.compile("§bCorkian Amplifier (I{1,3})");
 
     private int romanToArabic(String romanNumeral) {
@@ -90,7 +90,7 @@ public class ItemLevelOverlay implements Listener {
 
         // emerald pouch tier
         if (item == Items.DIAMOND_AXE) {
-            Matcher emeraldPouchMatcher = EMERALD_POUCH_PATTERN.matcher(StringUtils.normalizeBadString(name));
+            Matcher emeraldPouchMatcher = EMERALD_POUCH_PATTERN.matcher(name);
             if (emeraldPouchMatcher.find()) {
                 if (!UtilitiesConfig.Items.INSTANCE.levelKeyShowsItemTiers) return;
                 if (UtilitiesConfig.Items.INSTANCE.romanNumeralItemTier) {
@@ -105,7 +105,7 @@ public class ItemLevelOverlay implements Listener {
 
         // cork amp tier
         if (item == Item.getItemFromBlock(Blocks.STONE_BUTTON)) {
-            Matcher amplifierMatcher = CORKIAN_AMPLIFIER_PATTERN.matcher(StringUtils.normalizeBadString(name));
+            Matcher amplifierMatcher = CORKIAN_AMPLIFIER_PATTERN.matcher(name);
             if (amplifierMatcher.find()) {
                 if (!UtilitiesConfig.Items.INSTANCE.levelKeyShowsItemTiers) return;
                 if (UtilitiesConfig.Items.INSTANCE.romanNumeralItemTier) {
--- a/src/main/java/com/wynntils/modules/utilities/overlays/inventories/ItemLevelOverlay.java
+++ b/src/main/java/com/wynntils/modules/utilities/overlays/inventories/ItemLevelOverlay.java
@@ -7,6 +7,7 @@ package com.wynntils.modules.utilities.overlays.inventories;
 import com.wynntils.McIf;
 import com.wynntils.core.events.custom.RenderEvent;
 import com.wynntils.core.framework.interfaces.Listener;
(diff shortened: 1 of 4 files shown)


src/main/java/com/wynntils/modules/utilities/overlays/inventories/ItemLevelOverlay.java
```diff
@@ -23,7 +23,46 @@
 
 public class ItemLevelOverlay implements Listener {
 
-    private static final Pattern POWDER_NAME_PATTERN = Pattern.compile("(?:Earth|Thunder|Water|Fire|Air|Blank) Powder (VI|IV|V|I{1,3})");
+    private static final Pattern POWDER_NAME_PATTERN = Pattern.compile("(?:Earth|Thunder|Water|Fire|Air|Blank) Powder ([IV]{1,3})");
```
magicus: I'm just curious. What is a "Blank" powder? Never heard about...

src/main/java/com/wynntils/modules/utilities/overlays/inventories/ItemSpecificationOverlay.java
```diff
@@ -150,7 +149,48 @@ private void renderOverlay(GuiContainer gui) {
                     destinationName = m.group(1);
                     color = MinecraftChatColors.AQUA;
                     xOffset = -1;
-                    scale = 0.8f;
+                    scale = UtilitiesConfig.Items.INSTANCE.specificationTierSize;
+                }
+            }
+
+            if (UtilitiesConfig.Items.INSTANCE.powderSpecification) {
+                Pattern powder = Pattern.compile("§[2ebcf8].? ?(Earth|Thunder|Water|Fire|Air|Blank) Powder ([IV]{1,3})");
```
magicus: Maybe you can reuse the POWDER_NAME_PATTERN here? Or move it to a better class and use it in both locations.

src/main/java/com/wynntils/modules/utilities/overlays/inventories/ItemSpecificationOverlay.java
```diff
@@ -150,7 +149,48 @@ private void renderOverlay(GuiContainer gui) {
...
+                            color = MinecraftChatColors.GRAY; // Dark gray is too hard to see, use normal instead
+                            break;
+                    }
+                    xOffset = -1;
+                    scale = UtilitiesConfig.Items.INSTANCE.specificationTierSize;
+                }
+            }
+
+            if (UtilitiesConfig.Items.INSTANCE.emeraldPouchSpecification) {
+                Pattern emeraldPouch = Pattern.compile("§aEmerald Pouch§2 \\[Tier (IX|X|VI{1,3}|IV|V|I{1,3})]");
```
magicus: Same here, seems unnecessary to duplicate the regex.

magicus (approved):
Overall I think this looks good. It is acceptable as-is, but I'd prefer if you could de-duplicate the regexes.

src/main/java/com/wynntils/modules/utilities/overlays/inventories/ItemLevelOverlay.java
Incompleteusern: It's a powder of non type from corkus which can be exchanged for an elemental powder, still has tiers.

src/main/java/com/wynntils/modules/utilities/overlays/inventories/ItemSpecificationOverlay.java
Incompleteusern: Maybe move it to the Powder class

src/main/java/com/wynntils/modules/utilities/overlays/inventories/ItemSpecificationOverlay.java
Incompleteusern: Shouldn't the last ] be escaped too?

src/main/java/com/wynntils/modules/utilities/overlays/inventories/ItemSpecificationOverlay.java
HeyZeer0: Yeah agreed, also compiling the pattern takes time, the appropriate behavior is to make it static

src/main/java/com/wynntils/modules/utilities/overlays/inventories/ItemSpecificationOverlay.java
```diff
@@ -150,7 +149,48 @@ private void renderOverlay(GuiContainer gui) {
...
+                        case "Fire":
+                            color = MinecraftChatColors.RED;
+                            break;
+                        case "Air":
+                            color = MinecraftChatColors.WHITE;
+                            break;
+                        case "Blank":
+                            color = MinecraftChatColors.GRAY; // Dark gray is too hard to see, use normal instead
+                            break;
+                    }
```
HeyZeer0: Pretty sure all of this is already available on the [Powder](https://github.com/Wynntils/Wynntils/blob/development/src/main/java/com/wynntils/core/framework/enums/Powder.java) class, you just need to find retrieve the enum value based on the group, better to keep it centralized so we can easily change it in the future

src/main/java/com/wynntils/modules/utilities/overlays/inventories/ItemSpecificationOverlay.java
HeyZeer0: Yup, make it static!

HeyZeer0 (changes requested):

src/main/java/com/wynntils/modules/utilities/overlays/inventories/ItemSpecificationOverlay.java
ryanzhoudev: > Shouldn't the last ] be escaped too?

Is that required? We already escaped the first [ and there are no other ['s in the regex, so the last one is interpreted as a normal character by default

src/main/java/com/wynntils/modules/utilities/overlays/inventories/ItemSpecificationOverlay.java
ryanzhoudev: Yup sounds good, though the Powder class doesn't have specifications for blank powders so that's gonna have to stay

Incompleteusern: Is it possible to add blank powder to the Powder class as it is a powder

ryanzhoudev: > Is it possible to add blank powder to the Powder class as it is a powder

No because blank powders don't have symbols

PR #355, opened 2021-10-09
chore: upgrade buildsystem to forgegradle 5

xcfrg:
Upgrade build system to forgegradle 5

- use plugins dsl instead of legacy buildscript (settings.gradle)
- upgrade gradle, forgegradle, and shadow
- use gradle.properties to store more constants
- use newer way of specifing the java compatibility
- remove inheritOutputDirs which hasn't been required since fg 2
- use newer way of adding dependencies to shadow
- disable the jar task, so it doesn't run twice
- use newer way of adding the version and description to the mcmod.info
- gradle 7.2


settings.gradle
```diff
@@ -0,0 +1,19 @@
...
+            switch (requested.id.id) {
+                case "net.minecraftforge.gradle":
+                    useModule("net.minecraftforge.gradle:ForgeGradle:${requested.version}")
+                    break
+            }
+        }
+    }
+}
+
+rootProject.name = "Wynntils"
```
magicus: missing newline

magicus:
I always feel gradle changes are a bit scary, maybe because there is too much magic going on. Have you verified that the generated output is the same? Have you verified that this works OK with the IDEs folks are using (for my part, IntelliJ; I don't know what the rest is using).

Otoh, there sure were much cruft in the old gradle file, so it's probably due for a cleanup like this.

xcfrg:
I've verified that it works on Intellij, but I don't use or have tested eclipse
Additionally, what would you think about using an access token to login instead of the username and password fields that are there currently? That's what's done in forge mdks by default (in newer versions) and it allows people with a microsoft account to login too

build.gradle
```diff
@@ -194,3 +157,5 @@ class Version {
         "${major}.${minor}.${revision}_$build-MC1.12.2"
     }
 }
+
+jar.dependsOn("shadowJar")
```
magicus: Maybe you need to configure your editor to append a newline at EOF if it is missing?

magicus:
Access token sounds good! (I have recently converted to Microsoft account, but have not tried Wynntils development since). But it also sounds like a separate change from this, right? Or is it dependent on these changes?

DevScyu (approved):

xcfrg:
> 
> 
> Access token sounds good! (I have recently converted to Microsoft account, but have not tried Wynntils development since). But it also sounds like a separate change from this, right? Or is it dependent on these changes?

There's a way I'd like to do it which is dependent on these changes

build.gradle
magicus: Please add a newline at the end of the file

magicus (changes requested):
If you fix the missing newline, I'll push this.

PR #356, opened 2021-10-09
Add github workflow caching

xcfrg:
- improves ci build times significantly

### DIFF
--- a/.github/workflows/gradle-ci.yml
+++ b/.github/workflows/gradle-ci.yml
@@ -28,4 +28,4 @@ jobs:
         uses: actions/upload-artifact@v2
         with:
           name: build
-          path: build/libs 
+          path: build/libs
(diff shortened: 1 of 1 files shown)


magicus:
Can you explain this patch a bit? You are renaming the workflow file, changing the labels, and adding a cache attribute to the java action..?

xcfrg:
oops, i pushed the wrong commit D:

xcfrg:
I switched adopt to termiun because adoptopenjdk has moved to eclipse termiun (adoptium.net), removed the chmod because in gradle 7 the file is executable by default, and the caching attribute caches the generated source by forgegradle, so it doesn't decompile the game every time it runs

magicus (changes requested):
There is no newline on the last line. But there is a trailing space. Please fix.

magicus (approved):
I think this looks good. As long as the workflow runs fine...

DevScyu (approved):

PR #358, opened 2021-10-10
Feature/path finder

Kepler-17c:
Goal: A satnav-like functionality to find the path to coords, NPC, prof-stations, bank, ...; guiding with a lootrun-path-like overlay.


magicus:
Exciting idea. But complex... I think it would be good to have proper testing of the more "stand-alone" parts with the actual pathfinding algorithm.

TrigonometricDev:
damn he's really goin for it.. impressive

Kepler-17c:
> damn he's really goin for it.. impressive

It's coming together ^^  
When I first pitched the idea in suggestions, I hugely underestimated it. Might be a bit over-engineered in some parts, but I like to have a clean solution. Especially the BVH I'm currently working on will be really cool.
(Had some more ideas for it, but they're not required and would just be a nice addition [tree cuts] or a huge generalisation and multi-purpose approach [LOD interfaces].)

src/main/java/com/wynntils/modules/map/pathfinder/data/BoundingVolumeHierarchy.java
```diff
@@ -228,73 +298,202 @@ public T findNearest (final Point3d location) {
...
+			final Pair<BvhNode<T>, Map<T, BvhNode<T>>> rebuiltStructure = buildTree(oldTree, oldPendingInserts,
+					oldPendingDeletes, oldStrategy, oldLeafCount);
+			// apply updated tree
+			synchronized (this.rebuildLock) {
+				this.rebuildInProgress = false;
+				this.tree = rebuiltStructure.a;
+				this.objectCache = rebuiltStructure.b;
+				// sort out changes to pending operations during rebuild
+				final Set<T> filteredPendingInserts = new HashSet<>();
+				final Set<T> filteredPendingDeletes = new HashSet<>();
```
Incompleteusern, line 470: I dont get the logic behind `filteredPendingInserts ` and `filteredPendingDeletes `

Kepler-17c: For data consistency each operation on the BVH has transaction behaviour: The internal tree structure and rebuilding is completely hidden.

To ensure this behaviour, the current state of contained elements is fully described by `tree`, `pendingInserts`, `pendingDeletes`. Each operation takes the current state in a synchronised block and applied modifications by replacing the state objects. That way an iterator won't be affected by changes happening while processing.

Those filtered inserts/deletes are a consequence of changes occurring during rebuilding of the tree. Elements may have been added to or removed from `pendingInserts`/`pendingDeletes`. Sorting out the changes is done by comparing the old state from when rebuilding was started with how it is now that rebuilding is finished.

Incompleteusern: ah

PR #361, opened 2021-10-12
Add animated name for items with 0% rolls

ryanzhoudev:
Not sure why there are two commits with the same description, but anyway:

- Renamed the module that provided rainbow names to perfect items
- Added animated name generation to that same file
- Obfuscation amount in animated name is available as a user setting
- Entire feature is toggleable using the same setting as the rainbow name for perfect items, which was not refactored/renamed - only the name and description were updated


P0keDev:
Neat concept, nice execution!

PR #363, opened 2021-10-22
Add shift to bulk buy

ryanzhoudev:
- Adds two settings related to this feature:
    - One to toggle the entire feature on/off
    - One to set the number of items per bulk buy

Items are checked through private boolean isBulkShopConsumable - takes an ItemStack and attempts to match with manually listed items. These only include items that are stackable, currently scrolls, potions, and the consumables from the Ava workshop merchant.


Incompleteusern:
LGTM, other than 2 things:
1. The method in which the clicking functionality is handled all conditions check if the slotIn does not equal null, it may be good to move it to the beginning and return if its null
2. The check could work for crafted items

ryanzhoudev:
> LGTM, other than 2 things:
> 
> 1. The method in which the clicking functionality is handled all conditions check if the slotIn does not equal null, it may be good to move it to the beginning and return if its null
> 2. The check could work for crafted items

Both resolved!

ryanzhoudev:
Visuals are now dynamically updated when the user presses LShift - the price will update and the text at the bottom saying "shift click to buy x" will also update.

Incompleteusern:
```java
            if (priceMatcher.matches() && !nbt.hasKey("wynntilsBulkShiftPrice") && isShiftHeld) {
                loreLine = loreLine.replace(priceMatcher.group(1), String.valueOf(Integer.parseInt(priceMatcher.group(1)) * UtilitiesConfig.INSTANCE.bulkBuyAmount));
                nbt.setBoolean("wynntilsBulkShiftPrice", true);
            } else if (priceMatcher.matches() && nbt.hasKey("wynntilsBulkShiftPrice") && !isShiftHeld) {
                loreLine = loreLine.replace(priceMatcher.group(1), String.valueOf(Integer.parseInt(priceMatcher.group(1)) / UtilitiesConfig.INSTANCE.bulkBuyAmount));
                nbt.removeTag("wynntilsBulkShiftPrice");
            }
```

Personally I dislike this code a decent bit, as it seems a bit iffy. An imo better option is to simply put the bulk price under so you can see both prices at once, it'd be helpful for at a glance.

ryanzhoudev:
> Personally I dislike this code a decent bit, as it seems a bit iffy. An imo better option is to simply put the bulk price under so you can see both prices at once, it'd be helpful for at a glance.

Actually I agree - currently, the shift visual behaviour change doesn't actually take into account how many emeralds are in the user's inventory, and we just directly sub out the price. I can add in a check for the amount of emeralds in inventory, then sub in the X mark instead of the check, though I don't know if we have the symbols for that? Though this would be significantly easier with a static bit like you said. I still want to keep the dynamic bit at the bottom though, where it changes the green text based on if you're holding shift or not.

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
```diff
@@ -1060,4 +1072,64 @@ public void clearEmptyTooltip(GuiOverlapEvent.ChestOverlap.HoveredToolTip.Pre e)
...
+        if (!UtilitiesConfig.INSTANCE.shiftBulkBuy || !isBulkShopConsumable(e.getItemStack())) return;
+
+        ItemStack is = e.getItemStack();
+        boolean isShiftHeld = Keyboard.isKeyDown(Keyboard.KEY_LSHIFT);
+        List<String> newLore = new ArrayList<>();
+        NBTTagCompound nbt = is.getTagCompound();
+
+        for (String loreLine : ItemUtils.getLore(is)) {
+            // Do not add the last bit of info text
+            if (!loreLine.contains("§aPurchasing ") && !loreLine.contains("§aShift-click to purchase ")) {
```
Incompleteusern: This could be replaced with just iterating to the before last line, as price matching can't match the last line anyways

```java
List<String> oldLore = ItemUtils.getLore(is);
for (int i = 0; i < oldLore.length - 1; i++) {
    String loreLine = oldLore.get(i);
    ...
}

File: src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java [hunk 2]
```diff
@@ -1060,4 +1072,64 @@ public void clearEmptyTooltip(GuiOverlapEvent.ChestOverlap.HoveredToolTip.Pre e)
...
+                newLore.add(loreLine);
+            }
+
+            Matcher priceMatcher = PRICE_REPLACER.matcher(loreLine);
+            if (priceMatcher.matches() && !nbt.hasKey("wynntilsBulkPrice")) {
```
Incompleteusern: Probably check whether the `wynntilsBulkPrice` key exists before applying the regex.

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
```diff
@@ -1060,4 +1072,64 @@ public void clearEmptyTooltip(GuiOverlapEvent.ChestOverlap.HoveredToolTip.Pre e)
...
+                // Determine if we have enough money to buy the bulk amount and add lore
+
+                String moneySymbol = RequirementSymbols.CHECKMARK_STRING;
```
Incompleteusern: `moneySymbol` could probably be made into a ternary.

```java
String moneySymbol = (bulkPrice > availMoney) ? TextFormatting.RED + RequirementSymbols.XMARK_STRING : TextFormatting.GREEN + RequirementSymbols.CHECKMARK_STRING;
String loreString = "§6 - "  + moneySymbol + " §f" + bulkPrice + "§7" + EmeraldSymbols.E_STRING + " (" + UtilitiesConfig.INSTANCE.bulkBuyAmount + "x)";
```

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
```diff
@@ -1060,4 +1072,64 @@ public void clearEmptyTooltip(GuiOverlapEvent.ChestOverlap.HoveredToolTip.Pre e)
...
+            }
+        }
+
+        // Only add the spacer once
+        if (!nbt.hasKey("wynntilsBulkShiftSpacer")) {
+            newLore.add(" ");
+            nbt.setBoolean("wynntilsBulkShiftSpacer", true);
+        }
+
+        String purchaseString; // If user is holding shift, just tell them they're already buying x amount
```
Incompleteusern: This could also be made a ternary, and `isShiftHeld` could be inlined

```java
String purchaseString = Keyboard.isKeyDown(Keyboard.KEY_LSHIFT) ? "§aPurchasing " : "§aShift-click to purchase " + UtilitiesConfig.INSTANCE.bulkBuyAmount;

File: src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java (hunk 1 above)
Comment (ryanzhoudev) [reply src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java]: This doesn't work, as the first iteration would ignore the price line.

Comment (Incompleteusern) [reply src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java]: yeah, you're right

File: src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java [hunk 5]
```diff
@@ -1060,4 +1072,57 @@ public void clearEmptyTooltip(GuiOverlapEvent.ChestOverlap.HoveredToolTip.Pre e)
...
+        List<String> newLore = new ArrayList<>();
+        NBTTagCompound nbt = is.getTagCompound();
+
+        for (String loreLine : ItemUtils.getLore(is)) {
+            // Do not add the last bit of info text
+            if (!loreLine.contains("§aPurchasing ") && !loreLine.contains("§aShift-click to purchase ")) {
+                newLore.add(loreLine);
+            }
+
+            Matcher priceMatcher = PRICE_REPLACER.matcher(loreLine);
```
Incompleteusern: The regex is being applied before the condition is being checked, which is probably expensive

Incompleteusern (commented):
One last thing

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
```diff
@@ -1123,12 +1119,9 @@ public void onItemHovered(ItemTooltipEvent e) {
...
-        if (isShiftHeld) {
-            purchaseString = "§aPurchasing " + UtilitiesConfig.INSTANCE.bulkBuyAmount;
-        } else {
-            purchaseString = "§aShift-click to purchase " + UtilitiesConfig.INSTANCE.bulkBuyAmount;
-        }
+        // If user is holding shift, just tell them they're already buying x amount
+        String purchaseString = Keyboard.isKeyDown(Keyboard.KEY_LSHIFT) ? "§aPurchasing " + UtilitiesConfig.INSTANCE.bulkBuyAmount : "§aShift-click to purchase " + UtilitiesConfig.INSTANCE.bulkBuyAmount;
+
         newLore.add(purchaseString);
         ItemUtils.replaceLore(is, newLore);
```
Incompleteusern, line 1124: Instead of modifying the lore it is probably better as of now to simply add it to `e.getToolTip()` assuming that works. This has a lot of plus sided imo in the ability to remove the contains in the for loop, as well as not needing the space nbt, only running the for loop till the price is found, not needing to check for the spacer.

```java
@SubscribeEvent
    public void onItemHovered(ItemTooltipEvent e) {
        // This stuff checks if the shift to bulk buy setting is on, and if it is, replace the price of items with bulk prices when shift is held
        if (!UtilitiesConfig.INSTANCE.shiftBulkBuy || !isBulkShopConsumable(e.getItemStack())) return;

        ItemStack is = e.getItemStack();
        NBTTagCompound nbt = is.getTagCompound();

        if (!nbt.hasKey("wynntilsBulkPrice")) { //Merge spacer with wynntilsBulkPrice, set up lore
            List<String> newLore = new ArrayList<>();
            for (String loreLine : ItemUtils.getLore(is)) {
                newLore.add(loreLine) //No need to check for spacing

                if (nbt.hasKey("wynntilsBulkPrice") continue;

                 Matcher priceMatcher = PRICE_REPLACER.matcher(loreLine);
                 if (!priceMatcher.matches()) continue;

                // Determine if we have enough money to buy the bulk amount and add lore

                int singularPrice = Integer.parseInt(priceMatcher.group(1));
                int bulkPrice = singularPrice * UtilitiesConfig.INSTANCE.bulkBuyAmount;
                int availMoney = PlayerInfo.get(InventoryData.class).getMoney(); // this value includes both raw emeralds and pouches

                String moneySymbol = (bulkPrice > availMoney) ? TextFormatting.RED + RequirementSymbols.XMARK_STRING : TextFormatting.GREEN + RequirementSymbols.CHECKMARK_STRING;
                String loreString = "§6 - " + moneySymbol + " §f" + bulkPrice + "§7" + EmeraldSymbols.E_STRING + " (" + UtilitiesConfig.INSTANCE.bulkBuyAmount + "x)";

                newLore.add(loreString);                    
                nbt.setBoolean("wynntilsBulkPrice", true);
            }

            // Only add the spacer once
            if (nbt.hasKey("wynntilsBulkPrice")) { //Handle odd event that it doesn't match
                newLore.add(" ");
            }
        }

        // If user is holding shift, just tell them they're already buying x amount
        String purchaseString = Keyboard.isKeyDown(Keyboard.KEY_LSHIFT) ? "§aPurchasing " + UtilitiesConfig.INSTANCE.bulkBuyAmount : "§aShift-click to purchase " + UtilitiesConfig.INSTANCE.bulkBuyAmount;

        e.getToolTip().add(purchaseString);
    }
```

Incompleteusern (changes requested):
Actual last thing

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
Incompleteusern: `e.getToolTip()` actually doesn't go directly after the lore

PR #365, opened 2021-10-29
Update health potion info variables

ryanzhoudev:
Replaced the old potion count variables with these two:
%hp_pot_charges%
%hp_pot_max%
%hp_pot_combined%

charges% returns remaining potion charges
max% returns max number of charges in that stack
combined% returns both in format "remaining/max" eg. 17/30


Incompleteusern:
LGTM. It might be possible to combine the checks for regexes for iterating through the inventory for both variables as they'd likely be used in conjunction.

ryanzhoudev:
> LGTM. It might be possible to combine the checks for regexes for iterating through the inventory for both variables as they'd likely be used in conjunction.

I think that might be possible if we return a list and just have the %var% thing use an index, not sure if that's possible in java though?

Incompleteusern:
A cache is used for other formatters.

magicus (approved):

PR #367, opened 2021-11-02
Bank lore search

ryanzhoudev:
Allows users to search item lores instead of item names in the bank search bar.
Format is 
`lore:search_term`
eg.
`lore:mana regen`
would return items on that page with mana regen.

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/overlays/inventories/BankOverlay.java
+++ b/src/main/java/com/wynntils/modules/utilities/overlays/inventories/BankOverlay.java
@@ -394,18 +394,12 @@ public class BankOverlay implements Listener {
 
         String searchText = searchField.getText().toLowerCase();
         if (searchText.startsWith("lore:")) {
-            String actualSearchText;
-            try {
-                actualSearchText = searchText.split("lore:")[1];
-            } catch (ArrayIndexOutOfBoundsException ignored) {
-                actualSearchText = "";
-            }
-
+            searchText = searchText.substring(5);
             for (int i = 0; i < bankGui.getLowerInv().getSizeInventory(); i++) {
                 if (i % 9 > 6) continue; // ignore sidebar items
 
                 ItemStack is = bankGui.getLowerInv().getStackInSlot(i);
-                if (TextFormatting.getTextWithoutFormattingCodes(ItemUtils.getStringLore(is)).toLowerCase().contains(actualSearchText))
+                if (TextFormatting.getTextWithoutFormattingCodes(ItemUtils.getStringLore(is)).toLowerCase().contains(searchText))
                     searchedItems.add(is);
             }
         } else {
--- a/src/main/java/com/wynntils/modules/utilities/overlays/inventories/BankOverlay.java
+++ b/src/main/java/com/wynntils/modules/utilities/overlays/inventories/BankOverlay.java
@@ -413,8 +413,11 @@ public class BankOverlay implements Listener {
     }
 
     private String getSearchString(ItemStack is, boolean isLore) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/wynntils/modules/utilities/overlays/inventories/BankOverlay.java
```diff
@@ -393,11 +393,24 @@ private void searchPageForItems(ChestReplacer bankGui) {
...
+                    searchedItems.add(is);
+            }
+        } else {
+            for (int i = 0; i < bankGui.getLowerInv().getSizeInventory(); i++) {
+                if (i % 9 > 6) continue; // ignore sidebar items
 
-            ItemStack is = bankGui.getLowerInv().getStackInSlot(i);
-            if (TextFormatting.getTextWithoutFormattingCodes(is.getDisplayName()).toLowerCase().contains(searchText)) searchedItems.add(is);
+                ItemStack is = bankGui.getLowerInv().getStackInSlot(i);
+                if (TextFormatting.getTextWithoutFormattingCodes(is.getDisplayName()).toLowerCase().contains(searchText))
```
magicus: Please use { }. If statements without blocks are super duper error prone! 

The simple one-line "if ... return" that are used liberally in Wynntils are kind of okay, but not complex stuff like this. (And the other case above)

magicus:
This fix screams for generalisation. :-) I won't force you, but I do think it'd be better off if you introduced a method like `getSearchString(ItemStack is, boolean isLore)` or something like that. Then read and strip "lore": and use it to set isLore. And just let this function return `ItemUtils.getStringLore(is))` or `is.getDescription()` depending on the value.

Incompleteusern:
LGTM. As magicus said, generalization is possible.

EDIT: Combination with ItemsPage is implausible

ryanzhoudev:
> This fix screams for generalisation. :-) I won't force you, but I do think it'd be better off if you introduced a method like `getSearchString(ItemStack is, boolean isLore)` or something like that. Then read and strip "lore": and use it to set isLore. And just let this function return `ItemUtils.getStringLore(is))` or `is.getDescription()` depending on the value.

Bit confused on what you mean by this, I'm gonna write something up and you can see if it's what you meant?

src/main/java/com/wynntils/modules/utilities/overlays/inventories/BankOverlay.java
```diff
@@ -393,14 +392,31 @@ private void searchPageForItems(ChestReplacer bankGui) {
...
+            if (getSearchString(is, searchLore).contains(searchText)) {
+                searchedItems.add(is);
+            }
+
         }
     }
 
+    private String getSearchString(ItemStack is, boolean isLore) {
+        if (isLore) return TextFormatting.getTextWithoutFormattingCodes(ItemUtils.getStringLore(is).toLowerCase());
+        return TextFormatting.getTextWithoutFormattingCodes(is.getDisplayName().toLowerCase());
```
magicus:
```suggestion
        if (isLore) {
          return TextFormatting.getTextWithoutFormattingCodes(ItemUtils.getStringLore(is).toLowerCase());
        } else {
          return TextFormatting.getTextWithoutFormattingCodes(is.getDisplayName().toLowerCase());
        }
```

magicus (approved):
Yes, exactly. Much better! I added a suggestion with code formatting. Apart from that, this looks good to me now.

src/main/java/com/wynntils/modules/utilities/overlays/inventories/BankOverlay.java
ryanzhoudev: Got it, thanks!

PR #368, opened 2021-11-02
Block spell hotkeys if not holding weapon

ryanzhoudev:
Works by checking if "§7 Class Req:" is in the held item lore.
Also I removed default binds for favourite trade and ingred pouch on this pr, imo it's a small enough change to be lumped in here.


Incompleteusern:
LGTM

PR #370, opened 2021-11-10
Add XP/Kills per Minute Info variable

lxhom:
I made a pretty simple variable to display how fast you're leveling because I wanted it myself. Works flawless, tested it ingame and handles level-ups perfectly. The manager is based on the AreaDPSManager, and the variables are based on the Level variables, so I hope I implemented this correctly. The only minor issue is that I'm not sure where exactly it'd be good to call update() so I put it into the getter, but it shouldn't really affect performance either way since it does basically nothing if the XP didn't change, but if there's a better place to call it I'll update it. I hope I'm doing this PR right, I'm super new to the whole Java/Forge ecosystem, so correct me if I'm doing something wrong :)


Incompleteusern:
The Kill counter would be off when dungeons/quests are finished, which is noted but would be hard to fix. For updating, there probably is a packet for when the experience bar changes that could be used instead .

lxhom:
> The Kill counter would be off when dungeons/quests are finished, which is noted but would be hard to fix. For updating, there probably is a packet for when the experience bar changes that could be used instead .

Working on it right now, I think I found a way to implement both:
- For the kills: Look for armor stands with the players name in brackets, they spawn when you kill a mob, and an implementation similar to this listening to damage labels is core.wynntils.modules.core.events.ClientEvents.checkDamageLabelFound(), so it should be pretty easy to implement
- For the XP: Theres already an implementation of an XP change listener in core.wynntils.modules.utilities.overlays.OverlayEvents.onClientTick, which works, so I can copy that code

lxhom:
So I screwed up the git repo a bit, but it should be fixed now. A new tick-based updater (like in the existing XP ticker) and accurate kill counts (with scanning kill messages) are now implemented :D

Incompleteusern:
Does kill detection work for nicknames?

lxhom:
> Does kill detection work for nicknames?

I can't test this myself, but I just asked someone on WC who was nicked whether the armor stand shows the IGN or the Nick.

> You: yo linnyflower dumb question about nicknames: does it show the real name or the nickname when you've killed a mob in the little popup?
> linnyflower: shows real name
> You: okay thanks :D

And [the code](https://github.com/Wynntils/Wynntils/blob/ea2f77955f7293cae11ba4e70f67166d08ea2fe9/src/main/java/com/wynntils/modules/core/events/ClientEvents.java#L301) just checks whether the name in the armor stand is the IGN, so it should work.

magicus (approved):

PR #377, opened 2021-11-13
Updated Fabled and Set Identification Cost

Pyromanical:
Fabled id cost formula changed from 26+12*(level) to 16+8*(level) in the last patch.


HeyZeer0:
Not only Fableds, but set items were also changed

Fabled: `level * 8 + 16`
Set: `level * 1.2 + 8`

HeyZeer0:
Nice work! Thanks for updating the values 😄

PR #381, opened 2021-11-16
Add WC Uptime info variable

ryanzhoudev:
%uptime%, just relays the one used in the server selector

oh no it also reformatted the entire file, that was not the intent

Considering %uptime_min% and %uptime_hrs% instead?

Let me know if the reformatting isn't okay or if I should just do the min/hrs implementation instead


magicus:
The reformatting was a bit annoying. It looks worse than the old as well. Try and adjust your IDE settings, if you get it right you should be able to revert the format change with a key press. And future changes are less likely to be disruptive.

ryanzhoudev:
Gonna redo

PR #388, opened 2021-11-27
Add next soul point command

ryanzhoudev:
Command is /nextsoulpoints

Lists all servers that are about to give soul points within the next 10 minutes, calculated by server uptimes.

### DIFF
--- a/src/main/java/com/wynntils/modules/core/CoreModule.java
+++ b/src/main/java/com/wynntils/modules/core/CoreModule.java
@@ -32,6 +32,7 @@ public class CoreModule extends Module {
         registerCommand(new CommandAdmin());
         registerCommand(new CommandCompass());
         registerCommand(new CommandForceUpdate());
+        registerCommand(new CommandNextSoulPoints());
         registerCommand(new CommandServer());
         registerCommand(new CommandTerritory());
         registerCommand(new CommandToken());
--- a/src/main/java/com/wynntils/modules/core/commands/CommandNextSoulPoints.java
+++ b/src/main/java/com/wynntils/modules/core/commands/CommandNextSoulPoints.java
@@ -1,0 +1,74 @@
+package com.wynntils.modules.core.commands;
+
+import com.wynntils.modules.utilities.managers.ServerListManager;
+import net.minecraft.command.CommandBase;
+import net.minecraft.command.ICommandSender;
+import net.minecraft.server.MinecraftServer;
+import net.minecraft.util.text.TextComponentString;
+import net.minecraft.util.text.TextFormatting;
+import net.minecraftforge.client.IClientCommand;
+
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class CommandNextSoulPoints extends CommandBase implements IClientCommand {
+
+
+    @Override
+    public boolean allowUsageWithoutPrefix(ICommandSender sender, String message) {
+        return false;
+    }
+
+    @Override
+    public String getName() {
+        return "nextsoulpoints";
+    }
+
+    @Override
+    public String getUsage(ICommandSender sender) {
+        return "Returns a list of servers/time pairs that regenerate soul points within the next 10 minutes";
+    }
+
+    @Override
+    public void execute(MinecraftServer server, ICommandSender sender, String[] args) {
+        Map<String, Integer> nextServers = new HashMap<>();
+
+        // Get all servers that will generate a soul point within the next 10 minutes
+        for (String availServer : ServerListManager.availableServers.keySet()) {
+            int uptimeMinutes = ServerListManager.getUptimeTotalMinutes(availServer);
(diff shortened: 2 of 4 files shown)


Incompleteusern:
Try calling `ServerListManager#updateServers` before and seeing if that fixes the need for the offset

magicus:
Interesting idea. Do you think it would be possible to enhance this to provide an easy way to find servers with a suitable uptime for loot runs?

magicus:
> 



> Do you think it would be possible to enhance this to provide an easy way to find servers with a suitable uptime for loot runs?

I've been thinking a bit more about this. I think it would be great to have a `/servers` command, that lists all servers, with uptime, number of free slots and time to next sp. (I think we used to have something like this?)

Then you could all an argument, like `[all] | sp | lootrun`, so that if you specify `sp`, you got the behavior that you've implemented here, sorting and filtering the list wrt time to next soul point. And if you specify `lootrun`, you filter and sort servers based on how suitable they are for lootrun (not too young, but not likely to reboot soon either). And `all` is just the default behavior. (Can maybe be skipped)

I realize this is quite different from your original patch. If you don't think this is a good idea, or think it is but don't want to implement it yourself, just ignore what I've said. :-)

ryanzhoudev:
I think the `/servers` command is a great idea, and the implementation idea you have is really clean too. However, I don't think I can implement this anytime soon, as I'm getting pretty busy with university applications. I'd be happy to hand this PR over to someone else if they would like to give it a shot.

magicus:
I might be able to spend time on it, at least later in december, unless someone else picks it up before that.

P0keDev:
> > 
> 
> 
> 
> > Do you think it would be possible to enhance this to provide an easy way to find servers with a suitable uptime for loot runs?
> 
> I've been thinking a bit more about this. I think it would be great to have a `/servers` command, that lists all servers, with uptime, number of free slots and time to next sp. (I think we used to have something like this?)
> 
> Then you could all an argument, like `[all] | sp | lootrun`, so that if you specify `sp`, you got the behavior that you've implemented here, sorting and filtering the list wrt time to next soul point. And if you specify `lootrun`, you filter and sort servers based on how suitable they are for lootrun (not too young, but not likely to reboot soon either). And `all` is just the default behavior. (Can maybe be skipped)
> 
> I realize this is quite different from your original patch. If you don't think this is a good idea, or think it is but don't want to implement it yourself, just ignore what I've said. :-)

Regarding sorting servers by lootrun quality: It wouldn't really work, since the main aspect of finding a good world to lootrun is for it to not have been looted by someone else recently. Beyond the world not being brand new, uptime isn't a particularly important factor of chest loot.

magicus:
@P0keDev Things might have changed since I studied the theory behind loot runs, but I learned that loot chests basically start out empty, and that the server add loots to them over time, until they are full. And the process to completely fill up the chests take >= 3 hours. Of course, if someone has looted the chests already you're out of luck, but starting your run in a too young world is also bit of pointless.

I've been doing a lot of loot runs lately, and I've started disliking the feature that Wynncraft skips the hub and just throws you into a world. Since I can't tell how old it is, I must start by going back to the hub. Then I might return to the same world, but I needed to know that it was not brand new.

Anyway, listing servers with uptime from a command while on a world seems like a good idea to me, regardless.

xcfrg:
> I've been doing a lot of loot runs lately, and I've started disliking the feature that Wynncraft skips the hub and just throws you into a world. Since I can't tell how old it is, I must start by going back to the hub. Then I might return to the same world, but I needed to know that it was not brand new.

Can't you check the Day counter in the F3 menu? Although I'm not arguing with the uptime command, it's a great idea

magicus:
@xcfrg Interesting idea, I have not even thought about it. That also means that if Wynntils want to check uptime for the current server, we can do so using Minecraft internals, and need not consult the Athena API.

ryanzhoudev:
> I've been doing a lot of loot runs lately, and I've started disliking the feature that Wynncraft skips the hub and just throws you into a world. Since I can't tell how old it is, I must start by going back to the hub. Then I might return to the same world, but I needed to know that it was not brand new.

#382, though that method does take data from the Athena cache

magicus:
@DonkeyBlaster Oh, sorry I missed that. Merged now.

ryanzhoudev:
> Try calling `ServerListManager#updateServers` before and seeing if that fixes the need for the offset

Tried, doesn't work.

ryanzhoudev:
>I've been thinking a bit more about this. I think it would be great to have a /servers command, that lists all servers, with uptime, number of free slots and time to next sp. (I think we used to have something like this?)

I just realized we already have a server command, `/s`. Argument `list` just lists all the servers online, nothing special. Argument `info <server>` just lists the player in that server.

Wynncraft also uses `/server` for some internal thing, and `/servers` just brings you to hub. I think a rework of the `/s` command is the best option here.

Incompleteusern:
Try changing the first - to a + in `ServerProfile#matchTime` and see if that works

ryanzhoudev:
Someone can either take over or we can just let this die; I don't want to work on it rn

PR #389, opened 2021-12-04
Emerald Pouch Hotkey Enhancements

ryanzhoudev:
### Emerald Pouch Manager
- EmeraldPouchManager#getPouchUsage now correctly returns 0 for empty pouches - it used to return -1

### Emerald Pouch Hotkeys
- Previously, the hotkey wouldn't work when multiple pouches were in the user's inventory
- This has been changed to allow hotkeys to work with multiple pouches in inventory, with a few conditions:
    - Player must not have multiple filled pouches in inventory
    - Player must not only have multiple empty pouches in inventory
    - Key will work when player has 1 filled pouch and any number of empty pouches in inventory

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/managers/PouchHotkeyManager.java
+++ b/src/main/java/com/wynntils/modules/utilities/managers/PouchHotkeyManager.java
@@ -35,15 +35,13 @@ public class PouchHotkeyManager {
         HashMap<Integer, Integer> emeraldPouches = new HashMap<Integer, Integer>() { // <inventory slot, usage>
         };
 
-        Integer slot = 4958;
-
         for (int i = 0; i < inventory.size(); i++) {
             ItemStack stack = inventory.get(i);
             if (!stack.isEmpty() && stack.hasDisplayName() && EmeraldPouchManager.isEmeraldPouch(stack)) {
                 emeraldPouches.put(i, EmeraldPouchManager.getPouchUsage(stack));
-                slot = i;
             }
         }
+
         pouchSwitch:
         switch (emeraldPouches.size()) {
             default:
--- a/src/main/java/com/wynntils/modules/utilities/managers/PouchHotkeyManager.java
+++ b/src/main/java/com/wynntils/modules/utilities/managers/PouchHotkeyManager.java
@@ -10,9 +10,7 @@ import net.minecraft.network.play.client.CPacketClickWindow;
 import net.minecraft.util.NonNullList;
 import net.minecraft.util.text.TextFormatting;
 
-import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.List;
 
 public class PouchHotkeyManager {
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/wynntils/modules/utilities/managers/PouchHotkeyManager.java
```diff
@@ -31,21 +32,54 @@ public static void onEmeraldHotkeyPress() {
...
+                        alreadyHasNonEmpty = true;
+                    } else if (amount > 0) { // Another pouch has a non-zero balance; notify user
+                        GameUpdateOverlay.queueMessage(TextFormatting.DARK_RED + "You have more than one filled emerald pouch in your inventory.");
+                        break pouchSwitch;
+                    }
+                }
+
+                // At this point, we have either multiple pouches with zero emeralds, or multiple pouches but only one with a non-zero balance
+                // Check to make sure we don't have a bunch of zero balances - if we do, notify user
+                System.out.println(emeraldPouches.values().stream().mapToInt(Integer::intValue).sum());
```
Incompleteusern: Seems to be left over debug, the condition is also the same as alreadyHasNonEmpty

src/main/java/com/wynntils/modules/utilities/managers/PouchHotkeyManager.java
```diff
@@ -31,21 +32,54 @@ public static void onEmeraldHotkeyPress() {
...
+                if (emeraldPouches.values().stream().mapToInt(Integer::intValue).sum() < 1) {
+                    GameUpdateOverlay.queueMessage(TextFormatting.DARK_RED + "You have more than one empty and no filled emerald pouches in your inventory.");
+                    break;
+                }
+
+                // Now, we know we have 1 used pouch and 1+ empty pouches - just open the used one
+                List<Integer> pouchSlotList = new ArrayList<>();
```
Incompleteusern: This could probably be merged with the above for loop by saving the index of the non empty pouch

Incompleteusern (changes requested):
LGTM, but a few things

src/main/java/com/wynntils/modules/utilities/managers/PouchHotkeyManager.java
```diff
@@ -46,32 +44,28 @@ public static void onEmeraldHotkeyPress() {
...
+                    } else if (emeraldPouches.get(key) > 0) { // Another pouch has a non-zero balance; notify user
                         GameUpdateOverlay.queueMessage(TextFormatting.DARK_RED + "You have more than one filled emerald pouch in your inventory.");
                         break pouchSwitch;
                     }
                 }
 
                 // At this point, we have either multiple pouches with zero emeralds, or multiple pouches but only one with a non-zero balance
                 // Check to make sure we don't have a bunch of zero balances - if we do, notify user
-                System.out.println(emeraldPouches.values().stream().mapToInt(Integer::intValue).sum());
                 if (emeraldPouches.values().stream().mapToInt(Integer::intValue).sum() < 1) {
```
Incompleteusern: This condition is the same as alreadyHasNonEmpty, probably could merge and rename the boolean to fit

PR #390, opened 2021-12-04
lootrun notes improvement / bugfix

McPlayHD:
Right now if you use "|" in lootrun notes it will create a new line (as it is intended to do?) but it doesn't reset the width counter because the method doesn't take into consideration manual newlines. This means that if the line width is met soon after there will be a very short line.
My patch splits all the text segments already splitted by the " " character again but this time at the "|" character and if it finds more than one entry it will create a new line and reset the width counter.

### DIFF
--- a/src/main/java/com/wynntils/core/utils/StringUtils.java
+++ b/src/main/java/com/wynntils/core/utils/StringUtils.java
@@ -104,12 +104,18 @@ public class StringUtils {
         int length = 0;
 
         for (String string : stringArray) {
-            if (length + string.length() >= max) {
-                result.append('|');
-                length = 0;
+            String[] lines = string.split("\\|", -1);
+            for (int i = 0; i < lines.length; i ++) {
+                String line = lines[i];
+                if (i > 0 || length + line.length() >= max) {
+                    result.append('|');
+                    length = 0;
+                }
+                if (line.length() > 0) {
+                    result.append(line).append(' ');
+                    length += line.length() + 1;  // +1 for the space following
+                }
             }
-            result.append(string).append(' ');
-            length += string.length() + 1;  // +1 for the space following
         }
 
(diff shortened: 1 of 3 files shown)


McPlayHD:
`/lr note a b c d e f | g h i j k l m n o p | q r s t u v w | x y z | This is just a sample text to show that automatic new lines still work.`
**Before:**
image
*note how the text gets cut again after `t` and before `u` without me wanting it to*

**After:**
image

src/main/java/com/wynntils/core/utils/StringUtils.java
```diff
@@ -104,12 +104,18 @@ public static String toMD5(String msg) {
         int length = 0;
 
         for (String string : stringArray) {
-            if (length + string.length() >= max) {
-                result.append('|');
-                length = 0;
+            String[] lines = string.split("\\|", -1);
+            for (int i = 0; i < lines.length; i ++) {
+                String line = lines[i];
+                if (i > 0 || length + line.length() >= max) {
```
Incompleteusern, line 110: Is the 2nd part of the condition necessary?

If i = 0 and line.length >= max, you likely don't need to append a '|' at the really beginning

McPlayHD: Yes, it's necessary because it could be that `lines.length` is equal to 1 (because there was no "|" in the text) and then we still want the newline to happen when the line gets too long. That's basically how it was before. If it would get too long it would create a newline and it did that before adding the new word.

Incompleteusern: ah

McPlayHD:
I fixed another issue where lootrun notes would grey out other overlays by simply adding `pushAttrib();` at the beginning and `popAttrib();` at the end.

PR #398, opened 2021-12-17
Abide by wynncraft's /stream

Incompleteusern:
/stream should now block almost all wynntils info about one's world


magicus (commented):
Any chance you can refactor this to only have to change one instance? Looks like it's ripe to break out a lot of code into a separate function here...

PR #399, opened 2021-12-18
feat: Implement dialogue hotbar button override

ryanzhoudev:
When in dialogue, send a CPacketHeldItemChange to trigger dialogue option if the user is already at the hotbar slot they pressed.
Works by checking if we're in dialogue which is set by a check in chat receive event, then checking keyboard button presses.

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
+++ b/src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
@@ -111,6 +111,8 @@ public class ClientEvents implements Listener {
 
     private static final Pattern CRAFTED_USES = Pattern.compile(".* \\[(\\d)/\\d\\]");
 
+    private Boolean isInInteractionDialogue = false;
+
     @SubscribeEvent
     public void onMoveEvent(InputEvent.MouseInputEvent e) {
         lastUserInput = System.currentTimeMillis();
@@ -124,6 +126,11 @@ public class ClientEvents implements Listener {
         if (currentTime <= lastAfkRequested + 500) return;
 
         lastUserInput = currentTime;
+
+        // Manually send a hotbar packet if we're in dialogue + the user presses the already selected hotbar slot
+        if (isInInteractionDialogue && Keyboard.getEventKeyState() && Keyboard.getEventKey() - 2 == McIf.player().inventory.currentItem) { // -2 because KEY_1 is 0x02, and hotbar is 0-8
+            McIf.mc().getConnection().sendPacket(new CPacketHeldItemChange(McIf.player().inventory.currentItem));
+        }
     }
 
     @SubscribeEvent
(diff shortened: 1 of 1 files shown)


src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
```diff
@@ -111,6 +111,8 @@
 
     private static final Pattern CRAFTED_USES = Pattern.compile(".* \\[(\\d)/\\d\\]");
 
+    private Boolean isInInteractionDialogue = false;
```
Incompleteusern, line 114: Maybe can be merged with `ChatManager.inDialogue`

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
```diff
@@ -319,6 +326,8 @@ public void chatHandler(ClientChatReceivedEvent e) {
         if (msg.startsWith("[Daily Rewards:")) {
             DailyReminderManager.openedDaily();
         }
+
+        isInInteractionDialogue = msg.contains("Select an option to continue");
```
Incompleteusern: Can be faked

Incompleteusern (commented):
LGTM, some minor things though

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
magicus: To avoid this, it's better to look at the message *with* formatting codes, and match from start to end, that is, equals(), or a regex enclosed in `^...$`.

magicus:
Any idea how well this will work/how hard it will break if user has remapped her hotkeys for slot selection?

ryanzhoudev:
> Any idea how well this will work/how hard it will break if user has remapped her hotkeys for slot selection?

It won't work, but will send a packet if the user has remapped that hotbar button to anything. But then again, Wynncraft itself still assumes no one has remapped their hotkeys as it still displays default hotbar numbers.

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
ryanzhoudev: I don't think a merge would be a good idea - inDialogue is currently broken(?) and most importantly doesn't differentiate between standard dialogue and options dialogue. I think we'd want to keep checks as tight as possible to not send unnecessary packets to Wynn.

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
DonkeyBlaster: How would I get this message with formatting codes? It looks like `McIf.getFormattedText(e.getMessage())` doesn't return any color codes. It looks like that area is just `\n\n                     Select an option to continue\n` though - if it's okay, I can just do a .contains() check with that string.

ryanzhoudev: [f043cda](https://github.com/Wynntils/Wynntils/pull/399/commits/f043cdab6d165bdd9c6551f3f1c1be39e24ad338) checks for the four possible status messages that Wynn can send with dialogue options; I think this is sufficient? It includes color codes checks which are illegal for players to manually send anyway.

magicus (approved):

PR #402, opened 2021-12-21
More Keybinds from Commands

TomR24:
Added Keybinds for /guild manage, /guild attack, /guild territory, /pets and /housing edit.
Haven't tested them but they should work the same as the Mob Totem Menu Keybind.
All new Keybinds are unassigned by default.


src/main/java/com/wynntils/modules/utilities/managers/KeyManager.java
```diff
@@ -79,6 +79,36 @@ public static void registerKeys() {
...
+        CoreModule.getModule().registerKeyBinding("Pet Menu", Keyboard.KEY_NONE, "Wynntils", KeyConflictContext.IN_GAME, true, () -> {
+            if (!Reference.onWorld) return;
+
+            McIf.player().sendChatMessage("/pets");
+        });
+
+        CoreModule.getModule().registerKeyBinding("Housing Edit Toggle", Keyboard.KEY_NONE, "Wynntils", KeyConflictContext.IN_GAME, true, () -> {
+            if (!Reference.onWorld) return;
+
+            McIf.player().sendChatMessage("/is edit");
```
magicus: Is `is` an alias for `housing`?

TomR24: yes

TomR24:
I have tested the keybinds and they all work, but I'm looking to change the /gu manage keybind so that it doesn't open the menu via a command, so it can be opened during wars. I don't know how to do that yet, so I might need some help with this.

src/main/java/com/wynntils/modules/utilities/managers/KeyManager.java
ryanzhoudev: IMHO it would be better if we used full names like /guild and /housing, instead of /gu and /is. This would allow anyone that wants to make any changes in the future to be able to ctrl+F more easily, as not everyone knows the shortened commands.

TomR24: I'll change this for the update that I'll put out once I got it working 👍

TomR24:
Tested the changes and everything is working. Sorry for the mess with the commits but it's ready now.

kristofbolyai (commented):
A lot could be changed to use Wynntils provided APIs and already existing code.

kristofbolyai:
Code tested, works locally. Features were implemented with (hopefully) up-to-date standards.


PS: This is my first feature I've worked on, really like working on Wynntils :) I hope my code is good enough to make it to production. (This comes from a C# guy lmao, I hate Java)

kristofbolyai:
Bug: The opened Guild manage menu interactions are not handled by the server.

magicus:
Given that https://github.com/Wynntils/Wynntils/pull/409 is now integrated, do you think this can be closed?

kristofbolyai:
> Given that #409 is now integrated, do you think this can be closed?

#409 makes all of this possible, except that in wars, you can't use /guild manage. This PR tried to implement opening the guild manage menu with the compass and then clicking, but in the end it did not work.

magicus:
@kristofbolyai @TomR24 I think the war implementation is better handled in a separate PR. I think we could get that working, using FakeInventory or such.

PR #409, opened 2021-12-28
Feature: Keybinds for user specified commands

kristofbolyai:
Adds the ability to set 6 keybinds for 6 different command, changable in Wynntils configs.

This pull request is an improvement of #402.

Some texts might require rewording.

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/managers/KeyManager.java
+++ b/src/main/java/com/wynntils/modules/utilities/managers/KeyManager.java
@@ -93,70 +93,68 @@ public class KeyManager {
 
         showLevelOverlayKey = UtilitiesModule.getModule().registerKeyBinding("Show Item Level Overlay", Keyboard.KEY_LCONTROL, "Wynntils", WynntilsConflictContext.AMBIENT, true, () -> {});
 
+        RegisterCustomCommandKeybinds();
+    }
+
+    private static void RegisterCustomCommandKeybinds() {
         CoreModule.getModule().registerKeyBinding("Command keybind 1", Keyboard.KEY_NONE, "Wynntils", KeyConflictContext.IN_GAME, true, () -> {
             String cKeyBind = UtilitiesConfig.CommandKeybinds.INSTANCE.cKeyBind1;
             if (cKeyBind.isEmpty())
                 return;
+            if (McIf.mc().currentScreen != null)
+                return;
 
-            if (cKeyBind.startsWith("/"))
-                McIf.player().sendChatMessage(cKeyBind);
-            else
-                McIf.player().sendChatMessage("/" + cKeyBind);
+            McIf.player().sendChatMessage("/" + cKeyBind);
         });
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/wynntils/modules/utilities/managers/KeyManager.java
```diff
@@ -91,6 +92,72 @@ public static void registerKeys() {
         CoreModule.getModule().registerKeyBinding("View Player's Gear", -98, "Wynntils", KeyConflictContext.IN_GAME, true, GearViewerUI::openGearViewer);
 
         showLevelOverlayKey = UtilitiesModule.getModule().registerKeyBinding("Show Item Level Overlay", Keyboard.KEY_LCONTROL, "Wynntils", WynntilsConflictContext.AMBIENT, true, () -> {});
+
+        CoreModule.getModule().registerKeyBinding("Command keybind 1", Keyboard.KEY_NONE, "Wynntils", KeyConflictContext.IN_GAME, true, () -> {
```
Incompleteusern: The keybind part is repeated for every single keybind, probably separate it with a function

src/main/java/com/wynntils/modules/utilities/managers/KeyManager.java
```diff
@@ -91,6 +92,72 @@ public static void registerKeys() {
...
+            String cKeyBind = UtilitiesConfig.CommandKeybinds.INSTANCE.cKeyBind1;
+            if (cKeyBind.isEmpty())
+                return;
+
+            if (cKeyBind.startsWith("/"))
+                McIf.player().sendChatMessage(cKeyBind);
+            else
+                McIf.player().sendChatMessage("/" + cKeyBind);
```
Incompleteusern: This should be done outside the keybind itself, perhaps in `public void onSettingChanged(String name) {` of line 503 in `UtilitiesConfig`

Incompleteusern: Also, add a check for if the player is in a gui before running the command

Incompleteusern (commented):
LGTM, although the keybinds don't seem to be implemented yet. Also a few things

kristofbolyai:
> LGTM, although the keybinds don't seem to be implemented yet.

Wdym, by that? Keybinds shouldn't be set to any default IMO, and should be decided by the user.

Incompleteusern:
oh Im dumb

src/main/java/com/wynntils/modules/utilities/managers/KeyManager.java
kristofbolyai: > Also, add a check for if the player is in a gui before running the command

Seemingly no need to do that as the keybinds dont work with open guis. Am I missing something?

kristofbolyai:
Fixed the issues, also minor change would just prefixing "/" at runtime be okay, its much cleaner and in the end it does not allow spamming chat messages so it should be alright?

magicus (approved):
This sounds like a reasonable functionality to add.

As a separate follow-up, we might consider giving hints on suitable commands, like how the %info% variables work.

PR #410, opened 2021-12-28
Add new icon Content_Raid

DevScyu:
Needs the actual icon to be added and setup in `src/main/resources/assets/wynntils/textures/map/data/map_icons_mappings.json`


magicus:
@DevScyu 
Is this intended as a draft PR? Or how should the texture and mappings json file be updated?

DevScyu:
> @DevScyu Is this intended as a draft PR? Or how should the texture and mappings json file be updated?

The texture `src/main/resources/assets/wynntils/textures/map/map_icons.png` needs updating to add the new icon and then the mappings file (`src/main/resources/assets/wynntils/textures/map/data/map_icons_mappings.json`) needs to be updated with the location of the new icon(s)

PR #413, opened 2021-12-29
Map QOL changes: If player is in war, use a generic X Z coordinate pair to open the map. Added a changeable default zoom level to maps.

kristofbolyai:
Features added:
- Maps now detect if player is in war, using the scoreboards. If so, they use configurable coordiante values to open the map. This centers the map at the desired coordinates, makes it so that the world maps can be used even in wars without the need to scroll unreasonable distances.
- Added a default zoom setting: Useful for guild activities, as the map zooms in too much often times. This way users can set a default zoom so they can view bigger regions without zooming out every time.

### DIFF
--- a/src/main/java/com/wynntils/modules/map/MapModule.java
+++ b/src/main/java/com/wynntils/modules/map/MapModule.java
@@ -23,6 +23,7 @@ import com.wynntils.modules.map.overlays.OverlayEvents;
 import com.wynntils.modules.map.overlays.ui.GuildWorldMapUI;
 import com.wynntils.modules.map.overlays.ui.MainWorldMapUI;
 import com.wynntils.modules.map.overlays.ui.WaypointCreationMenu;
+import com.wynntils.modules.map.overlays.ui.WorldMapUI;
 import com.wynntils.webapi.WebManager;
 import com.wynntils.webapi.WebReader;
 import net.minecraft.scoreboard.Score;
@@ -77,13 +78,7 @@ public class MapModule extends Module {
                 if (WebManager.getApiUrls() == null) {
                     WebManager.tryReloadApiUrls(true);
                 } else {
-                    //Check if in war using scoreboard
-                    boolean inWar = isPlayerInWar();
-
-                    if (inWar)
-                        Utils.displayGuiScreen(new MainWorldMapUI(MapConfig.WorldMap.INSTANCE.mapDefaultX, MapConfig.WorldMap.INSTANCE.mapDefaultZ));
-                    else
-                        Utils.displayGuiScreen(new MainWorldMapUI());
+                    openMap(false);
                 }
             }
         });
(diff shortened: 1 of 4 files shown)


src/main/java/com/wynntils/modules/map/MapModule.java
```diff
@@ -83,12 +93,35 @@ public void onEnable() {
...
                     WebManager.tryReloadApiUrls(true);
                 } else {
-                    Utils.displayGuiScreen(new GuildWorldMapUI());
+                    //Check if in war using scoreboard
+                    boolean inWar = isPlayerInWar();
+
+                    if (inWar)
+                        Utils.displayGuiScreen(new GuildWorldMapUI(MapConfig.WorldMap.INSTANCE.mapDefaultX, MapConfig.WorldMap.INSTANCE.mapDefaultZ));
+                    else
+                        Utils.displayGuiScreen(new GuildWorldMapUI());
```
magicus: Can you extract these two into a method which takes a main/guild map argument?

src/main/java/com/wynntils/modules/map/configs/MapConfig.java
```diff
@@ -113,6 +113,37 @@
         @Setting(displayName = "Opening Animation Length", description = "How long should be the opening animation")
         @Setting.Limitations.IntLimit(min = 50, max = 1000)
         public int animationLength = 250;
+
+        @Setting(displayName = "Default Map Zoom", description = "Default zoom value when a map is opened.")
+        @Setting.Limitations.IntLimit(min = -10, max = 300)
+        public int defaultMapZoom = 0;
+
+        @Setting(displayName = "Map Center X if in war", description = "Map center X coordinate will be this value when player is in war.")
+        public String mapDefaultXString = "-200";
```
magicus, line 122: Why String and not int?

src/main/java/com/wynntils/modules/map/overlays/ui/WorldMapUI.java
```diff
@@ -96,6 +97,10 @@ protected WorldMapUI(float startX, float startZ) {
             territories.put(territory.getFriendlyName(), new MapTerritory(territory).setRenderer(renderer));
         }
 
+        //Only set if there is no animation, otherwise it is handled by handleOpenAnimation()
+        if (!MapConfig.WorldMap.INSTANCE.openAnimation)
+            zoom = MapConfig.WorldMap.INSTANCE.defaultMapZoom;
```
magicus, line 102: I assume that if the user does not change the value, the behavior will be the same as the current behavior?

src/main/java/com/wynntils/modules/map/configs/MapConfig.java
kristofbolyai: If it's an integer, by default it has to have a min-max value and it will be rendered as a slider, not convenient for coordinates.

kristofbolyai: At least, that's what I observed, correct me if I'm wrong.

magicus: Aha, I see.

src/main/java/com/wynntils/modules/map/overlays/ui/WorldMapUI.java
kristofbolyai: No, it will center the map on a pre-defined coordinare, around middle of the map.
Do you want it changed to be an opt-in feature? If so are the default coordinates okay to stay, that way we know there is always a safe value there, no need to check them.

magicus: But only if you're in a war, right?

And really what I meant was "is the old default value for zoom 0?"

kristofbolyai: Yes, this only changes anything, if the player is in war. The default zoom value, according to code seemed to be 0 (or a fraction close to 0), but also opening the in-game map comfirms that it's 0, since max zoom value is -10, and by default when a map is opened there is only a small room to zoom.


Also, default zoom value will be still 0, if the user does not change that.

src/main/java/com/wynntils/modules/map/MapModule.java
magicus: If you fix this I'll integrate the PR

magicus (changes requested):

PR #419, opened 2021-12-31
Keybind patches

ryanzhoudev:
Removes mob totem bind as this can be done with the new custom command binds. Also unbound default ingredient pouch key to be consistent with emerald pouch.

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/managers/KeyManager.java
+++ b/src/main/java/com/wynntils/modules/utilities/managers/KeyManager.java
@@ -5,7 +5,6 @@
 package com.wynntils.modules.utilities.managers;
 
 import com.wynntils.McIf;
-import com.wynntils.Reference;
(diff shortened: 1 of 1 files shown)


magicus:
Maybe this was rushed a bit too much..? I think we need to be careful with removing functionality. 

I agree that if we had the functionality we now have, the mob totem keybind would not have been needed. However, as it now do exist, we risk breaking people's setup.

kristofbolyai:
> Maybe this was rushed a bit too much..? I think we need to be careful with removing functionality.
> 
> I agree that if we had the functionality we now have, the mob totem keybind would not have been needed. However, as it now do exist, we risk breaking people's setup.

Maybe revert for the next production release, and let people use the feature, and remove it in 2nd next production release?

ryanzhoudev:
Alternatively, set the first custom bind to be a mob totem bind, and just replace the old one with that

PR #424, opened 2021-12-31
fix: Add color check for pop off sound

ryanzhoudev:
Item pop off sound currently only checks for plaintext. Added a color check so spoof isn't possible.


magicus:
This is okay, but I believe regexps against formatted text with anchored start/end (^...$) is generally the best way.

PR #425, opened 2021-12-31
Command keybind 1 defaults to totem command and old totem keybind

kristofbolyai:
Fixes concerns over #419, where totem keybind was removed because of custom command keybinds. Totem keybind is now added back using this feature.


Incompleteusern:
Perhaps Add totem to the Presets too

magicus:
Thanks!

PR #426, opened 2022-01-01
Spellcasting by hotkeys is greatly improved

kristofbolyai:
List of changes and fixes:
- Added checks to make sure the currently held item can be used for casting spells:
  - Check if the currently held item is a weapon
  - Check if the class' level is high enough to use it
  - Check if the class type is the correct one to use the weapon
  - Check if the class has all necessary spell points assigned to use the weapon
- Added safeguards in the case when the spell cast hotkeys are spammed/run in repetition:
  - `QuickCastManager` does not allow queueing spells if there is one already queued, but yet to be executed
- If the player starts casting a spell, but does not finish it, then presses a spell cast keybinding, the spell will not try to cast, causing unintentional spell casts and attacks (and the `PacketQueue` will not try to resend the packets, causing spam).

EDIT: Added a change, if there is a spell casting in progress by the player, but then a keybind is pressed, it will try to finish the spell. For more info scroll below.

If any of these checks fail, the user will receive a chat message stating the fail condition (the first one that fails the checks).

image
image
image
image

### DIFF
--- a/src/main/java/com/wynntils/core/framework/instances/data/ActionBarData.java
+++ b/src/main/java/com/wynntils/core/framework/instances/data/ActionBarData.java
@@ -61,6 +61,8 @@ public class ActionBarData extends PlayerData {
 
                     spellData.setLastSpell(lastSpell, McIf.player().inventory.currentItem);
                 }
+                else if (spellData.getLastSpell() != SpellData.NO_SPELL)
+                    spellData.setLastSpell(SpellData.NO_SPELL, -1);
             }
         }
 
--- a/src/main/java/com/wynntils/core/framework/instances/data/ActionBarData.java
+++ b/src/main/java/com/wynntils/core/framework/instances/data/ActionBarData.java
@@ -4,6 +4,7 @@
 
 package com.wynntils.core.framework.instances.data;
 
+import com.wynntils.McIf;
 import com.wynntils.core.framework.enums.ClassType;
 import com.wynntils.core.framework.instances.containers.PlayerData;
 import com.wynntils.core.utils.StringUtils;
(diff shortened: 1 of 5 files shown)


magicus:
Yay! 🎉

magicus:
Do you think it's possible to make it so that if the user starts executing a spell "manually", and e.g. has just pressed a L (assuming a class where all spells start with L), and then presses a spell button, only the remaining part of the spell is sent? And if they've pressed like LR and want to execute a LRR spell, then only the remaining R gets sent, but if they've pressed LR and wanted a LLR spell, no spell is sent (since it can't be executed)?

magicus:
That's one of my problems, I spam the spell buttons and stuff gets messed up, and if I accidentally press the spell initial button (I switch between a mage and an archer so I always mix them up), the spell buttons all gets broken.

kristofbolyai:
> Do you think it's possible to make it so that if the user starts executing a spell "manually", and e.g. has just pressed a L (assuming a class where all spells start with L), and then presses a spell button, only the remaining part of the spell is sent? And if they've pressed like LR and want to execute a LRR spell, then only the remaining R gets sent, but if they've pressed LR and wanted a LLR spell, no spell is sent (since it can't be executed)?

Yes, it is rather easy with the already existing code. Pushing it in a few minutes.

magicus (approved):
Looks good to me.

kristofbolyai:
> Do you think it's possible to make it so that if the user starts executing a spell "manually", and e.g. has just pressed a L (assuming a class where all spells start with L), and then presses a spell button, only the remaining part of the spell is sent? And if they've pressed like LR and want to execute a LRR spell, then only the remaining R gets sent, but if they've pressed LR and wanted a LLR spell, no spell is sent (since it can't be executed)?

Added this, hopefully this does not introduce new problems. (Theoretically, it shouldn't, we will see 🙃)

magicus:
Thanks. Looks reasonable to me. Let's integrate this and try it out. (After all, the unstable branch has its name for a reason :-)). If it doesn't work, we'll tweak it. I'll seriously doubt it will be worse than the current situation, anyway.

PR #430, opened 2022-01-02
Fixed jumpy map zooming (along with other fixes) and added setting: Map auto close on movement and different tier loot chests now have configurable minimum zoom levels

kristofbolyai:
Changed and fixes: 
- Fixed jumpy zooming in and out.
- Added setting to auto close the map on movement: Only works if the user is NOT holding the map key, otherwise behavior is unchanged.
- There are now settings for all different loot chest tiers to set the minimum zoom level needed to display them on the map.
- Also fixed some really old typo (This might break compatibility with older versions when the user tries to load waypoints from copy-paste data, if it contains a CHEST (T3). Can't do much about this, users will update and it will go away.) 
- At some comments and code parts maximum zoom level was only 300, while it is 500, changed it for consistency


src/main/java/com/wynntils/modules/map/overlays/ui/GuildWorldMapUI.java
```diff
@@ -91,14 +97,30 @@ public void initGui() {
...
+
+            if (holdingDecided && !holdingMapKey && (lastPlayerX != Double.MAX_VALUE && lastPlayerZ != Double.MAX_VALUE) && (currentPosX != lastPlayerX || currentPosZ != lastPlayerZ)) {
+                McIf.mc().displayGuiScreen(null);
+                return;
+            }
+
+            lastPlayerX = currentPosX;
+            lastPlayerZ = currentPosZ;
+        }
+
```
HeyZeer0: The player can be moved by external actions, specifically check for the movement keybinds for better precision, and it will also remove the required variables. Just as a future tip for checking if the player moved this way, check if the player movement (movX, movY, movZ) != 0

src/main/java/com/wynntils/modules/map/overlays/ui/MainWorldMapUI.java
```diff
@@ -134,14 +140,30 @@ private static boolean isHoldingMapKey() {
...
+
+            if (holdingDecided && !holdingMapKey && (lastPlayerX != Double.MAX_VALUE && lastPlayerZ != Double.MAX_VALUE) && (currentPosX != lastPlayerX || currentPosZ != lastPlayerZ)) {
+                McIf.mc().displayGuiScreen(null);
+                return;
+            }
+
+            lastPlayerX = currentPosX;
+            lastPlayerZ = currentPosZ;
+        }
+
```
HeyZeer0: Same as above, also to not duplicate code please move this to a method and call it on both instances

src/main/java/com/wynntils/modules/map/overlays/ui/MainWorldMapUI.java
```diff
@@ -30,10 +31,15 @@
 public class MainWorldMapUI extends WorldMapUI {
 
     private boolean holdingMapKey = false;
+    private boolean holdingDecided = false;
     private final long creationTime = System.currentTimeMillis();
     private long lastClickTime = Integer.MAX_VALUE;
     private static final long doubleClickTime = Utils.getDoubleClickTime();
 
+    //Only updated if MapConfig.WorldMap.INSTANCE.autoCloseMapOnMovement is true, don't rely on this for anything else
```
HeyZeer0: Space after the //
```suggestion
    // Only updated if MapConfig.WorldMap.INSTANCE.autoCloseMapOnMovement is true, don't rely on this for anything else
```

src/main/java/com/wynntils/modules/map/instances/WaypointProfile.java
```diff
@@ -55,6 +55,11 @@ public int getZoomNeeded() {
         return zoomNeeded;
     }
 
+    public void setZoomNeeded(int zoomNeeded)
+    {
+        this.zoomNeeded = zoomNeeded;
+    }
```
HeyZeer0: Please follow our code patterns!
```suggestion
    public void setZoomNeeded(int zoomNeeded) {
        this.zoomNeeded = zoomNeeded;
    }
```

src/main/java/com/wynntils/modules/map/configs/MapConfig.java
```diff
@@ -88,6 +88,9 @@
     public static class WorldMap extends SettingsClass {
         public static WorldMap INSTANCE;
 
+        @Setting(displayName = "Auto Close Map On Movement", description = "Should the map close when the user starts moving? If the user opened the map by holding the keybind, the map will stay open until the key is released.")
+        public boolean autoCloseMapOnMovement = true;
```
HeyZeer0: Disabled by default
```suggestion
        public boolean autoCloseMapOnMovement = false;
```

src/main/java/com/wynntils/modules/map/overlays/ui/GuildWorldMapUI.java
```diff
@@ -91,14 +97,30 @@ public void initGui() {
     @Override
     public void drawScreen(int mouseX, int mouseY, float partialTicks) {
         // HeyZeer0: This detects if the user is holding the map key;
-        if (!holdingMapKey && (System.currentTimeMillis() - creationTime >= 150) && isHoldingMapKey()) holdingMapKey = true;
-
+        if (!holdingDecided && (System.currentTimeMillis() - creationTime >= 150)) {
+            holdingDecided = true;
+            if (isHoldingMapKey())
+                holdingMapKey = true;
+        }
```
HeyZeer0, line 100: I don't get the purpose of the `holdingDecided` variable? Aren't you able to just use the holdingMapKey? 🤔

src/main/java/com/wynntils/modules/map/overlays/ui/WorldMapUI.java
```diff
@@ -161,6 +162,26 @@ protected void createIcons() {
...
+                            case "Chest (T2)":
+                                waypointProfile.setZoomNeeded(MapConfig.WorldMap.INSTANCE.lootChestTier2MinZoom);
+                                break;
+                            case "Chest (T3)":
+                                waypointProfile.setZoomNeeded(MapConfig.WorldMap.INSTANCE.lootChestTier3MinZoom);
+                                break;
+                            case "Chest (T4)":
+                                waypointProfile.setZoomNeeded(MapConfig.WorldMap.INSTANCE.lootChestTier4MinZoom);
+                                break;
+                        }
```
HeyZeer0: Don't check for the display name, check literally for the enum values!
```suggestion
                        switch (waypointProfile.getType()) {
                            case LOOTCHEST_T1:
                                waypointProfile.setZoomNeeded(MapConfig.WorldMap.INSTANCE.lootChestTier1MinZoom);
                                break;
                            case LOOTCHEST_T2:
                                waypointProfile.setZoomNeeded(MapConfig.WorldMap.INSTANCE.lootChestTier2MinZoom);
                                break;
                            case LOOTCHEST_T3:
                                waypointProfile.setZoomNeeded(MapConfig.WorldMap.INSTANCE.lootChestTier3MinZoom);
                                break;
                            case LOOTCHEST_T4:
                                waypointProfile.setZoomNeeded(MapConfig.WorldMap.INSTANCE.lootChestTier4MinZoom);
                                break;
                        }
```

src/main/java/com/wynntils/modules/map/overlays/ui/WorldMapUI.java
```diff
@@ -406,7 +427,7 @@ protected void handleOpenAnimation() {
     protected void handleZoomAcceleration(float partialTicks) {
         if (McIf.getSystemTime() > zoomEnd) return;
 
-        float percentage = Math.min(1f, 1f - (zoomEnd - McIf.getSystemTime()) / ZOOM_RESISTANCE);
+        float percentage = MathHelper.clamp(1f - (zoomEnd - McIf.getSystemTime()) / ZOOM_RESISTANCE, 0f, 1f);
```
HeyZeer0: It is probably wise as well to avoid these issues in the future to replace the system time with nanoseconds (`System.nanoTime()`) since this one is calculated by the JVM and has higher precision. Tho some extra stuff will need to be changed if that's a lot of work, don't mind it 😄

HeyZeer0 (commented):
Thanks for your contribution. Just some changes need to be made and it will look really good!
On the other side, the icon zoom config seems a bit hacky, personally, I would prefer it removed but I'll let the others decide 😄

Thanks again!

HeyZeer0 (changes requested):
Forgot to request the changes lol

src/main/java/com/wynntils/modules/map/overlays/ui/GuildWorldMapUI.java
kristofbolyai: It is used in the code part below. We can't check for `!holdingMapKey` as it is defaulted to false, the condition will not work properly. That is why `holdingDecided` is introduced.

```java
            if (holdingDecided && !holdingMapKey && (lastPlayerX != Double.MAX_VALUE && lastPlayerZ != Double.MAX_VALUE) && (currentPosX != lastPlayerX || currentPosZ != lastPlayerZ)) {
                McIf.mc().displayGuiScreen(null);
                return;
            }
```

TomR24:
Another cool thing would be an option to have the map become partially transparent when moving

kristofbolyai:
> Another cool thing would be an option to have the map become partially transparent when moving

Perhaps, may do it sometime later. (Definitely not this PR tho.)

src/main/java/com/wynntils/modules/map/overlays/ui/WorldMapUI.java
kristofbolyai: Changed in https://github.com/Wynntils/Wynntils/pull/430/commits/587d8537fff526d69b3a9eb254e65e39d7e4b504.

Was it that simple or did I miss something? 🤔

src/main/java/com/wynntils/modules/map/overlays/ui/GuildWorldMapUI.java
kristofbolyai: This is actually outdated now, as I extracted this into a new method, but same idea applies, we need to know 
if holding was decided or otherwise we might use `holdingMapKey`  before it was set, hence why `holdingDecided` was added.

See `WorldMapUI.java` line 542, the reason this was a needed change.

src/main/java/com/wynntils/modules/map/overlays/ui/GuildWorldMapUI.java
```diff
@@ -91,14 +93,26 @@ public void initGui() {
...
 
+        if (MapConfig.WorldMap.INSTANCE.autoCloseMapOnMovement)
+        {
+            if (checkForPlayerMovement(holdingDecided, holdingMapKey))
+            {
+                McIf.mc().displayGuiScreen(null);
+                return;
+            }
+        }
+
```
HeyZeer0: Please follow our code standards! 
```suggestion
        if (MapConfig.WorldMap.INSTANCE.autoCloseMapOnMovement && checkForPlayerMovement(holdingDecided, holdingMapKey))  {
            McIf.mc().displayGuiScreen(null);
            return;
        }

```

src/main/java/com/wynntils/modules/map/overlays/ui/MainWorldMapUI.java
```diff
@@ -134,14 +140,26 @@ private static boolean isHoldingMapKey() {
...
 
+        if (MapConfig.WorldMap.INSTANCE.autoCloseMapOnMovement)
+        {
+            if (checkForPlayerMovement(holdingDecided, holdingMapKey))
+            {
+                McIf.mc().displayGuiScreen(null);
+                return;
+            }
+        }
+
```
HeyZeer0: Please follow our code standards!
```suggestion
        if (MapConfig.WorldMap.INSTANCE.autoCloseMapOnMovement && checkForPlayerMovement(holdingDecided, holdingMapKey))  {
            McIf.mc().displayGuiScreen(null);
            return;
        }

```

HeyZeer0 (changes requested):
Awesome!! Feel free to merge it yourself whenever the last changes were made!

Thanks! 😄

PR #435, opened 2022-01-05
Dialogue and chat improvements

P0keDev:
Most importantly, these changes finally fix the issue where dialogue would cause the chat history to be cleared. This was actually a one line fix, and the cause of the issue was code that was changed and pushed in #375 without being tested or reviewed.

Besides that fix, this PR fixes a small issue with chat timestamps, removes clickable message options made redundant by vanilla Wynncraft, fixes clickable duel messages, and adds an option to progress dialogue by right clicking the NPC (disabled by default). I also added a change to make the check for dialogue completion ignore color, which I hope will fix a sporadic issue with dialogue never being registered as over, which makes a complete mess of chat. However, I found no way to reliably reproduce this issue, nor could I pinpoint specifically what caused it, so this fix is mostly speculative (but does not break anything that currently works). Any further insight into this bug would be very useful.

These changes probably warrant a push to stable, since fixing dialogue issues is a big deal.

### DIFF
--- a/src/main/java/com/wynntils/modules/chat/configs/ChatConfig.java
+++ b/src/main/java/com/wynntils/modules/chat/configs/ChatConfig.java
@@ -83,6 +83,9 @@ public class ChatConfig extends SettingsClass {
     @Setting(displayName = "Clickable Duel Requests", description = "Should duel requests provide a clickable command?", order = 19)
     public boolean clickableDuelMessage = true;
 
+    @Setting(displayName = "Right Click for Dialogue", description = "Should dialogue progress when you right click the NPC?\n\n§8This may not work on every NPC.", order = 20)
+    public boolean rightClickDialogue = false;
+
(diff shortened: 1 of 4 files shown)


Incompleteusern:
Sorry about that. How did this clear dialogue history?
image

EDIT: ah

Incompleteusern:
Also how does the timestamp fix work?

P0keDev:
> Also how does the timestamp fix work?

The original implementation assumed that the given text component had an empty root component and that all the actual text existed in its siblings, so it simply put the timestamp components at the front of the sibling list. However, sometimes the root component does have some text - notably, in the greyed out messages you get while in dialogue - and in this case, the timestamp would end up within the overall message.

Now, an entirely new component is created and the original message is appended to it after adding the timestamp, so it's not possible for that to happen

PR #436, opened 2022-01-05
Added ingredient guide and ingredient favoriting

kristofbolyai:
Finally got around doing this somewhat long awaited feature. The main focus was implementing the favoriting of ingredients. This means that while the guide is fully working, the search functions are very limited right now. In an upcoming PR, I am planning to address this and add the Advanced and Basic search modes the item guide has.

Some "bugs" or things that may need change, but I need input/help with:
- Spawner eggs don't have the correct "color" (type)
- Quest book menu for the ingredient guide has the same icon (compass) as the item guide, not sure what to change it to nor how to do it

Few screenshots can be found here: https://imgur.com/a/OMSbaRd


P0keDev:
For skull textures, you can copy over [this method](https://github.com/P0keDev/Wynntils/blob/7fe9a2bd30e5a2d16245590a48313fdae1e1ff25/src/main/java/com/wynntils/core/utils/ItemUtils.java#L92). Should be noted that the textures are hard-coded into Athena since they're not in the API at all, and they're probably not up to date since I did them a year ago

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientProfile.java
```diff
@@ -0,0 +1,233 @@
...
+            itemLore.add(itemModifier.getFormattedModifierText());
+        }
+
+        if (durability != null || duration != null || itemModifiersList.size() > 0)
+            itemLore.add("");
+
+        // Bolyai, 2022.01.04: As of the date, untradeable ingredients don't exist, adding this for future
+        // untradeable
+        if (untradeable)
+            itemLore.add(TextFormatting.RED + "Untradeable Item");
```
P0keDev: Should be "Untradable" in item lore

kristofbolyai:
> For skull textures, you can copy over [this method](https://github.com/P0keDev/Wynntils/blob/7fe9a2bd30e5a2d16245590a48313fdae1e1ff25/src/main/java/com/wynntils/core/utils/ItemUtils.java#L92). Should be noted that the textures are hard-coded into Athena since they're not in the API at all, and they're probably not up to date since I did them a year ago

Already doing most of that, except the random uuid. Kinda strange how that is needed, but thanks, add this next commit.

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientProfile.java
```diff
@@ -0,0 +1,233 @@
...
+    String material;
+    List<ProfessionType> professions = new ArrayList<>();
+    Map<String, IngredientIdentificationContainer> statuses = new HashMap<>();
+    Map<String, Integer> itemModifiers = new HashMap<>();
+    Map<String, Integer> ingredientModifiers = new HashMap<>();
+
+    transient IngredientTier ingredientTier;
+    transient List<ItemModifier> itemModifiersList = new ArrayList<>();
+    transient List<IngredientModifier> ingredientModifiersList = new ArrayList<>();
+    transient Map<ItemModifierType, ItemModifier> itemModifiersTypeMap = new HashMap<>();
```
Incompleteusern: `itemModifiersList` and `itemModifiersTypeMap` should be merged imo, maybe `itemModifiers` and `itemModifiersList` plus `ingredientModifiers` and `ingredientModifiersList`

kristofbolyai: Merged `itemModifiersList` into `itemModifiersTypeMap`, but would rather keep `itemModifiersTypeMap`, may save lot of processing later on when search features are added. Also removed `ingredientModifiersList` but I may add them back later on, when search features are added, if they are worth caching instead of more processing.

kristofbolyai:
> For skull textures, you can copy over [this method](https://github.com/P0keDev/Wynntils/blob/7fe9a2bd30e5a2d16245590a48313fdae1e1ff25/src/main/java/com/wynntils/core/utils/ItemUtils.java#L92). Should be noted that the textures are hard-coded into Athena since they're not in the API at all, and they're probably not up to date since I did them a year ago

Also fixed this, did not mention it in commits.

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientProfile.java
```diff
@@ -0,0 +1,213 @@
...
+            builder.append("✫");
+        }
+
+        builder.append(TextFormatting.GRAY);
+
+        for (int i = 0; i < 3 - tier; i++) {
+            builder.append("✫");
+        }
+
+        builder.append(ingredientTier.getBracketColor()).append("]");
```
P0keDev: The entire tier string should be defined in the IngredientTier enum, instead of building it in this messy way

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientProfile.java
```diff
@@ -0,0 +1,213 @@
...
+
+        if (material.matches("(.*\\d.*)")) {
+            String[] split = material.split(":");
+
+            stack = new ItemStack(Item.getItemById(Integer.parseInt(split[0])));
+            if (split.length <= 1) return guideStack = stack;
+
+            stack.setItemDamage(Integer.parseInt(split[1]));
+        } else
+            stack = new ItemStack(Item.getByNameOrId(material));
```
P0keDev: Brackets should be consistent

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientProfile.java
```diff
@@ -0,0 +1,213 @@
...
+    int level;
+    String material;
+    List<ProfessionType> professions = new ArrayList<>();
+    Map<String, IngredientIdentificationContainer> statuses = new HashMap<>();
+    Map<String, Integer> itemModifiers = new HashMap<>();
+    Map<String, Integer> ingredientModifiers = new HashMap<>();
+
+    transient IngredientTier ingredientTier;
+    //transient List<ItemModifier> itemModifiersList = new ArrayList<>();
+    //transient List<IngredientModifier> ingredientModifiersList = new ArrayList<>();
```
P0keDev: Don't include commented out code

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientProfile.java
```diff
@@ -0,0 +1,213 @@
...
+    public void postParse() {
+        ingredientTier = IngredientTier.fromInteger(tier);
+
+        itemModifiersTypeMap = new HashMap<>();
+        for (String key : itemModifiers.keySet()) {
+            ItemModifierType type = ItemModifierType.valueOf(key.toUpperCase(Locale.ROOT));
+            ItemModifier itemModifier = new ItemModifier(type, itemModifiers.get(key));
+            itemModifiersTypeMap.put(type, itemModifier);
+        }
+    }
```
P0keDev: This implementation for modifiers (both item and ingredient) is quite messy - a better approach would be to define container objects like is done [here](https://github.com/Wynntils/Wynntils/blob/7d6e7fc65fe981542ed5684078afc8f405a1cadf/src/main/java/com/wynntils/webapi/profiles/item/objects/ItemRequirementsContainer.java), which is what I intended when designing the format of the Athena ingredient list

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientProfile.java
```diff
@@ -0,0 +1,213 @@
...
+                type = IngredientModifierType.valueOf(key.toUpperCase(Locale.ROOT));
+
+            itemLore.addAll(Arrays.asList(new IngredientModifier(type, ingredientModifiers.get(key)).getLoreLines()));
+        }
+
+        if (ingredientModifiers.size() > 0)
+            itemLore.add("");
+
+        ItemModifier durability = itemModifiersTypeMap.get(ItemModifierType.DURABILITY);
+        ItemModifier duration = itemModifiersTypeMap.get(ItemModifierType.DURATION);
```
P0keDev: See other comment on modifiers

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientProfile.java
```diff
@@ -0,0 +1,213 @@
...
+            itemLore.add(itemModifier.getFormattedModifierText());
+        }
+
+        if (durability != null || duration != null || itemModifiers.size() > 0)
+            itemLore.add("");
+
+        // Bolyai, 2022.01.04: As of the date, untradeable ingredients don't exist, adding this for future
+        // untradeable
+        if (untradeable)
+            itemLore.add(TextFormatting.RED + "Untradeable item");
```
P0keDev: Already mentioned this, should be "Untradable Item"

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientProfile.java
```diff
@@ -0,0 +1,213 @@
...
+
+        itemLore.add(TextFormatting.GRAY + "Crafting Lv. Min: " + level);
+
+        for (ProfessionType profession : professions) {
+            itemLore.add("  " + profession.getProfessionIconChar() + " " + TextFormatting.GRAY + profession.getDisplayName());
+        }
+
+        // updating lore
+        {
```
P0keDev: Unless the entire method is formatted into these blocks (which I don't recommend), this shouldn't be its own block

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientProfile.java
```diff
@@ -0,0 +1,213 @@
...
+            display.setString("Name", getIngredientStringFormatted());
+
+            tag.setTag("display", display);
+            tag.setBoolean("Unbreakable", true);  // this allow items to have damage
+
+            if (stack.getItem() == Items.SKULL) {
+                HashMap<String, String> ingredientHeadTextures = WebManager.getIngredientHeadTextures();
+
+                if (ingredientHeadTextures.containsKey(name))
+                {
```
P0keDev: Bracket should be on same line as the if statement

src/main/java/com/wynntils/webapi/profiles/item/objects/IdentificationContainer.java
```diff
@@ -69,6 +72,9 @@ public boolean hasConstantValue() {
     }
 
     public static String getAsLongName(String shortName) {
+        if (shortName.equals("gatherXPBonus"))
+            return "Gather XP Bonus";
+
```
P0keDev: This should not be hard-coded like this, if it's not getting parsed correctly then something may need to change on Athena's end

src/main/java/com/wynntils/webapi/profiles/ingredient/enums/IngredientTier.java
```diff
@@ -0,0 +1,47 @@
...
+            case 1:
+                return TIER_1;
+            case 2:
+                return TIER_2;
+            case 3:
+                return TIER_3;
+        }
+        return null;
+    }
+}
```
P0keDev: Use the `@SerializedName` annotation on the enum definitions instead, so you can directly serialize the numbers into the corresponding tier enum

src/main/java/com/wynntils/webapi/profiles/item/objects/IdentificationContainer.java
```diff
@@ -29,6 +29,9 @@ public IdentificationContainer(IdentificationModifier type, int baseValue, boole
         calculateMinMax();
     }
 
+    public IdentificationContainer() {
+    }
+
```
P0keDev: Don't think this should be here

P0keDev (changes requested):
Went over things in more detail, feel free to ask for clarification on anything

src/main/java/com/wynntils/webapi/profiles/item/objects/IdentificationContainer.java
kristofbolyai: The constructor below needs it.

```java
    public IngredientIdentificationContainer(IdentificationModifier type, int minimum, int maximum) {
        this.type = type;
        this.minimum = minimum;
        this.maximum = maximum;
        this.isFixed = minimum == maximum;
    }
```

kristofbolyai: Actually that whole inheritance is not needed, fixed.

src/main/java/com/wynntils/webapi/profiles/item/objects/IdentificationContainer.java
kristofbolyai: This is getting parsed as `Gather X P Bonus`, needs to be changed to `gatherXpBonus` in Athena?

kristofbolyai: Removed the hardcoded part.

kristofbolyai: Also, when/if we change that in Artemis, we should get the skull textures updated.

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientProfile.java
```diff
@@ -0,0 +1,203 @@
...
+
+        if (statuses.size() > 0)
+            itemLore.add("");
+
+        int left = ingredientModifiers.getLeft();
+        int right = ingredientModifiers.getRight();
+        int above = ingredientModifiers.getAbove();
+        int under = ingredientModifiers.getUnder();
+        int touching = ingredientModifiers.getTouching();
+        int notTouching = ingredientModifiers.getNotTouching();
```
Incompleteusern: Please most most of this stack lore generation to the relevant classes. 

It might be good to remove the enums and inline the lore line generation, but that's up to you.

Incompleteusern (changes requested):
Very cool! 

Also the icon is the same as item page since its passed the same to the constructor. If you want to change it, either make an icon yourself and create a reference to it, or I can put in the creative-team channel to ask for one to be made if you want

kristofbolyai:
> Also the icon is the same as item page since its passed the same to the constructor. If you want to change it, either make an icon yourself and create a reference to it, or I can put in the creative-team channel to ask for one to be made if you want.

Yes, I was aware why it was the same, but yes, asking the creative people would be helpful, I'm not an artist myself.

Incompleteusern:
Okay, Ill add a message in the channel later

src/main/java/com/wynntils/webapi/profiles/ingredient/enums/IngredientTier.java
```diff
@@ -0,0 +1,30 @@
...
+
+public enum IngredientTier {
+    @SerializedName("0")
+    TIER_0(0, "§7[§8✫✫✫§7]"),
+    @SerializedName("1")
+    TIER_1(1, "§6[§e✫§8✫✫§6]"),
+    @SerializedName("2")
+    TIER_2(2, "§5[§d✫✫§8✫§5]"),
+    @SerializedName("3")
+    TIER_3(3, "§3[§b✫✫✫§3]");
```
P0keDev: It's usually better to use `TextFormatting` than to directly include the color codes like this

P0keDev (commented):
Apologies for taking so long to look over this

Overall looks good, although the modifier enums seem a bit unnecessary - they could probably be replaced with helper functions within the container classes, if not directly coded into the lore generation. Someone else may want to weigh in on that, though

src/main/java/com/wynntils/webapi/profiles/ingredient/enums/IngredientTier.java
magicus: @P0keDev You think that? I typically prefer to use the literal strings, since these can be copied/compared directly with the output from Wynncraft. (Unless it's a string that *we* create to display to the user, then TextFormatting.RED etc is better)

kristofbolyai: @magicus What do you think about removing the enums? Should we keep it or move to helper function or add it into lore generation?

P0keDev: @magicus I prefer it because it's easier to understand and less prone to mistakes. Raw color codes are fine for checking Wynn messages in my opinion, since those messages are often a jumble of colors and what colors they are isn't super relevant, but for something like this I believe using `TextFormatting` is better - ultimately it's fine either way

magicus: I needed to check the usage. Yes, I agree with you now. If it's used this way (to create strings of our own), TextFormatting is better.

magicus: @kristofbolyai I have no strong opinion on that. The code looks fine to me (but I have not spent too much time understanding all details).

Incompleteusern (approved):

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientModifiers.java
```diff
@@ -0,0 +1,86 @@
...
+            return new String[] {
+                    TextFormatting.GREEN + "+" + modifierValue + "%" + TextFormatting.GRAY + " Ingredient Effectiveness",
+                    TextFormatting.GRAY + "(To ingredients " + type.getDisplayValue() + " this one)"
+            };
+        else
+            return new String[] {
+                    TextFormatting.RED.toString() + modifierValue + "%" + TextFormatting.GRAY + " Ingredient Effectiveness",
+                    TextFormatting.GRAY + "(To ingredients " + type.getDisplayValue() + " this one)"
+            };
+    }
```
P0keDev: You could simply pass the appropriate lore string into this function and use it, instead of passing it a string that has to be parsed into an enum which stores the lore string in question.

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientItemModifiers.java
```diff
@@ -0,0 +1,102 @@
...
+        if (type == ItemModifierType.DURATION || type == ItemModifierType.CHARGES || type == ItemModifierType.DURABILITY)
+            if (modifierValue > 0)
+                return TextFormatting.GREEN + "+" + modifierValue + " " + type.getEffectString();
+            else
+                return TextFormatting.RED.toString() + modifierValue + " " + type.getEffectString();
+        else if (modifierValue > 0)
+            return TextFormatting.RED + "+" + modifierValue + " " + type.getEffectString();
+
+        return TextFormatting.GREEN.toString() + modifierValue + " " + type.getEffectString();
+    }
```
P0keDev: The same applies here. I also recommend using braces for the outer if statements here.

P0keDev (changes requested):
My bad, I should've been more explicit - my comment on enums was referring to `IngredientModifierType` and `ItemModifierType`, and what I said about the tier enum was unrelated to that (which, for the record, is a better implementation than simply using an integer in my opinion).

Also, after taking a closer look, I can confidently say that those two enums should be removed; they contribute very little besides making everything more verbose.

magicus (approved):
LGTM!

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientItemModifiers.java
```diff
@@ -0,0 +1,114 @@
...
+            case "strength":
+            case "intelligence":
+            case "dexterity":
+            case "agility":
+            case "defense":
+                return itemModifierName.substring(0, 1).toUpperCase(Locale.ROOT) + itemModifierName.substring(1) + " Min.";
+        }
+
+        return null;
+    }
```
P0keDev: This method is completely unnecessary -- you pass a string to `getFormattedModifierText` only to use a switch statement that turns it into a different string (or even the same string). Remove this and pass the formatted string into `getFormattedModifierText`.

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientModifiers.java
```diff
@@ -0,0 +1,97 @@
...
+                return "above";
+            case "under":
+                return "under";
+            case "touching":
+                return "touching";
+            case "notTouching":
+                return "not touching";
+        }
+        return null;
+    }
```
P0keDev: Same issue here - remove this method

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientItemModifiers.java
```diff
@@ -0,0 +1,114 @@
...
+
+    public static String getFormattedModifierText(String itemModifierName, int modifierValue) {
+        String requirementString = getRequirementStringFromModifier(itemModifierName);
+        if (itemModifierName.equals("duration") || itemModifierName.equals("charges") || itemModifierName.equals("durability")) {
+            if (modifierValue > 0)
+                return TextFormatting.GREEN + "+" + modifierValue + " " + requirementString;
+            else
+                return TextFormatting.RED.toString() + modifierValue + " " + requirementString;
+        }
+        else if (modifierValue > 0) {
```
P0keDev: Should be all one line

P0keDev (changes requested):

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientItemModifiers.java
kristofbolyai: I mean, in the end it has to be parsed somewhere, so its just a matter of moving that code? Also, I made that switch have "useless" cases so all modifiers are defined in the code aswell, but I can see why we could remove it.

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientModifiers.java
magicus: Is this not about converting one "type" of string to another? More precisely, mapping from how values are stored in the source json, to how information should be presented to the user. If this is the case, I do indeed see an argument for keeping both this and the other string conversion method.

kristofbolyai: > Is this not about converting one "type" of string to another? More precisely, mapping from how values are stored in the source json, to how information should be presented to the user. If this is the case, I do indeed see an argument for keeping both this and the other string conversion method.

That is the case. See https://github.com/Wynntils/Wynntils/pull/436#discussion_r792335077

magicus: Maybe the original enum did have a purpose after all, then. I think I disagree with @P0keDev here, and would argue that the enum did have a code clarity purpose, but that it -- if anything -- did not go far enough. If it had, I don't think Poke would have suggested that it should be removed. 

Let's make sure we are all aboard for this, before doing further changes and then just reverting them. :-) What if you restored a enum type, with the following features:
 * a factory method that makes it clear they are created from json, e.g. `IngretientPlacement::fromJsonMarker(String s)`
 * a string output method that makes it clear it is for the GUI, e.g. `getFriendlyName()`

@P0keDev Would you find that reasonable? (My reasoning is about both enum types)

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientItemModifiers.java
magicus: This I agree with fully ;-)

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientModifiers.java
kristofbolyai: > a factory method that makes it clear they are created from json, e.g. `IngretientPlacement::fromJsonMarker(String s)`

Not quite sure what that would change except marking that it is created from JSON which is already in the `webapi/profiles` folder, wouldn't that be enough?

P0keDev: Am I missing something? The only time this method is used is to convert a string literal passed into `getLoreLines`, as seen in `generateAllLoreLines`. I don't see it used anywhere to convert api values.

Moreover, there's no need to do any additional parsing of the api - the deserializer will do it all automatically using the appropriately named variables in both these container classes. There's no extra information that will be missing.

kristofbolyai: @P0keDev , where do you suggest to do the parsing of the api then? It has to be done as API returns the modifier values but only a short name. For example: you have to convert `left` into `to the left of` when making the ingredient lores. Same with some other modifiers.

magicus: @P0keDev The enums represent an abstract concept. This concept is realized in two different ways in the code: as a string in the json representation, and as a string for the user interface. 

So while it might seem like this is "converting" one string to another, that's mostly due to circumstances. What if the json representation had used integers instead, like 1 for left, 2 for right, etc. Would you have agreed in that case that the enum makes sense?

P0keDev: @magicus I'm afraid I don't understand your point, because the format of the json representation doesn't need to be kept track of or parsed - it's deserialized for us. All of the information about the modifiers is extracted and stored using the variables `left`, `right`, `under`, etc. Every ingredient follows the same format, so all you need is the raw number values. With those, you already know exactly how to construct the ingredient lore. There is no reason to convert from anything to anything, because the raw values are the only information that varies between ingredients. If the issue is about avoiding string literals within the code, constants could be used and named accordingly - I don't see what an enum provides besides unnecessary bloat.

kristofbolyai: Oh, I think I see what you mean now, the IngredientProfile lore generation could just get the value and use the correct strings there.

P0keDev: @kristofbolyai The only thing being parsed from the api is the numbers. The variables have names, and we know exactly what corresponds to what. The leftward ingredient modifier is stored in `left` - to format the string, you just need to return `left + "% Ingredient Effectiveness (Ingredients to the left of this one)"`

kristofbolyai: > @kristofbolyai The only thing being parsed from the api is the numbers. The variables have names, and we know exactly what corresponds to what. The leftward ingredient modifier is stored in `left` - to format the string, you just need to return `left + "% Ingredient Effectiveness (Ingredients to the left of this one)`

Yep, got it.

kristofbolyai: Waiting for @magicus before changing this.

P0keDev:
For the record, when I was designing the format of the Athena ingredient api, I had the implementation of [ItemRequirementsContainer](https://github.com/Wynntils/Wynntils/blob/development/src/main/java/com/wynntils/webapi/profiles/item/objects/ItemRequirementsContainer.java) in mind for these specific stats.

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientItemModifiers.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    public int getCharges() {
+        return charges;
+    }
+
+    public static String getFormattedModifierText(String itemModifierName, int modifierValue) {
+        String requirementString = getRequirementStringFromModifier(itemModifierName);
+        if (itemModifierName.equals("duration") || itemModifierName.equals("charges") || itemModifierName.equals("durability")) {
+            if (modifierValue > 0)
```
magicus: Also, multi-line if statements without braces is a recipe for problems. Please use braces.

magicus: See e.g. https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/ for a cautionary tale.

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientModifiers.java
magicus: Ok, now I have dived much deeper into the change than I did before. :-)

I think I see @P0keDev's point now. You have a single method with a switch, but this is only called with hard-coded string literals, so it can essentially be "inlined" at "compile time", and thus "inlined" at the source level.

I can agree with that. 

However. I still think we're missing some opporunity for code improvement. Let me see if I can be more concrete in what I mean. TBC.

magicus: Normally, I'd say that the start would be to convert the long list of integer members to an array. But that will break the json deserializing. :-( Right now I can't think of a good way to solve this (more than the realization that automagic deserialization of json can be anathema to good design). What I'd like to see, basically, is generateAllLoreLines turned into a loop, instead of a set of hard-coded duplicated pieces of code. 

Maybe the best solution would be to have a really, really simple holder class for json deserialization, and then move the results from that class, into a more general class that can associate each json field with an array `int[] modifiers`, and a corresponding `String[] modifiersDesc = { "to the left of", ... }`

Then we could have just one instance of "parsing" the json, when assigning `modifiers[0] = jsonObj.left` etc.

magicus: If it's worth the trouble, I don't know. Otherwise just go with the code duplication and Pokes suggestions.

kristofbolyai: Reading through all of this, I think, this PR we should use P0ke's implementation idea, fix obvious code format issues and create a new draft PR and discuss the refactoring of the deserialization (maybe including Item Guide aswell). @magicus @P0keDev, what do you think?

magicus: I'm okay with that. I'm also okay with not doing a follow-up PR. It's not the most pressing issue in Wynntils.

P0keDev: Sounds good to me.

The more hardcoded, redundant implementation that's already used for items certainly isn't the cleanest, but I do think it has benefits in being easier to understand and not quite as verbose. Of course, if you really wanted to, you could have the best of both worlds by reformatting the api to use arrays, so we could deserialize those directly. Ultimately, I think it would be more trouble than it's worth.

magicus: My rule of thumb is that copying code once or twice is better than abstracting, but after that, the DRY principle wins. But, let's keep it this way now. Maybe we can rethink things for Artemis.

kristofbolyai:
Oops, I did not mean to request a review from Incompleteusern, sorry about that.

A fourth check might not be unnecessary on this PR though, just to make sure everybody agrees.

src/main/java/com/wynntils/webapi/profiles/ingredient/IngredientModifiers.java
```diff
@@ -0,0 +1,86 @@
...
+
+
+    public boolean anyExists() {
+        return left != 0 || right != 0 || under != 0 || above != 0 || touching != 0 || notTouching != 0;
+    }
+
+    public static String[] getLoreLines(String modifierName, int modifierValue) {
+        String readableModifier = modifierName;
+
+        switch (modifierName) {
```
magicus: I still think you're missing the point..? The idea was that you shouldn't do any string replacement here, but call it with "to the left of" instead of "left" as the single place in the code where you're calling it with a string literal "left".

magicus: At least that's what I thought that Poke and I agreed on. :-)

kristofbolyai: Oh, right, fixed it now. 😅

I totally got sidetracked by those minor refactorings.

magicus (approved):
Getting kind of lost now on what to look for, but I think this looks ok and has a promise to be the final iteration. :) Let @P0keDev have a look as well first, then you can push.

P0keDev (approved):

PR #440, opened 2022-01-11
feat: Add favourite item confirmation for items in identifier slots

ryanzhoudev:


ryanzhoudev:
Abandoning for now, might come back to it in the future

PR #442, opened 2022-01-13
Properly order spell cast keybind settings

ryanzhoudev:
Yes, this will reset people's settings.


magicus:
I like it. The order always confused me. We might need to give a heads-up tho.

PR #448, opened 2022-02-10
feat: Add consumable charge count in hotbar

ryanzhoudev:
Didn't use any existing method of getting charges as they wouldn't work for healing, skill, and crafted all together. Also not sure if string manipulation is the best way to do this but I didn't think of another so...

Edit: now also works with scrolls and food
Edit 2: This also happens to overlay the charges in other inventories, which we can't really check for as that event doesn't return any inventory data. (Can't check for inventory slot num, if in bank, etc)

Looks like this (without the red background)
image

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/overlays/inventories/ItemLevelOverlay.java
+++ b/src/main/java/com/wynntils/modules/utilities/overlays/inventories/ItemLevelOverlay.java
@@ -18,6 +18,7 @@ import com.wynntils.modules.utilities.managers.KeyManager;
 import net.minecraft.init.Items;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
+import net.minecraft.util.text.TextFormatting;
 import net.minecraftforge.fml.common.eventhandler.SubscribeEvent;
 
 import java.util.regex.Matcher;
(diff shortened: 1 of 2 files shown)


src/main/java/com/wynntils/modules/utilities/overlays/inventories/ItemLevelOverlay.java
```diff
@@ -63,55 +64,64 @@ public static String romanToArabic(String romanNumeral) {
...
-        // item level
-        IntRange level = ItemUtils.getLevel(lore);
-        if (level != null) {
-            event.setOverlayText(UtilitiesConfig.Items.INSTANCE.averageUnidentifiedLevel ? level.toString() : Integer.toString(level.getAverage()));
-            return;
+        // Hotbar charges
+        if (UtilitiesConfig.INSTANCE.showConsumableChargesHotbar && (item == Items.POTIONITEM || item == Items.DIAMOND_AXE)) { // potion charge count in hotbar
+            if (!name.contains("[") || !name.contains("/")) return; // Make sure it's actually some consumable
+            String[] consumable = name.split(" ");
+            String[] charges = consumable[consumable.length - 1].split("/");
```
kristofbolyai: Maybe add some checks so that the array indexes will not go out of bounds? Maybe this could be spoofed with something?

ryanzhoudev: I'm not sure where this could be spoofed, we check for actual item matches and if they have [ and / in their names
Even if someone were to rename a crafted, the `.split(" ")` gets the `.length() - 1` so it's always the last bit of text in the name, which is guaranteed to be charges

kristofbolyai: @DonkeyBlaster Could you just add 2 simple checks to make sure the array is in bounds (just as a safeguard later on for Wynncraft updates and to make sure the player does not get stuck into a state where a specific inventory always crashes just because somehow the code thinks an item is consumable but it is not actually so it will produce an exception.) Once done, we can merge.

ryanzhoudev: Done, at least I think that's what you were talking about

kristofbolyai (approved):

PR #449, opened 2022-02-11
Fixing hotbar highlight colours and settings

McPlayHD:
- Disabling highlights of specific tiers would not work on your hotbar.
- Hotbar colours would also not be customizable.
- Also added a setting for disabling crafted item highlights.

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/overlays/inventories/RarityColorOverlay.java
+++ b/src/main/java/com/wynntils/modules/utilities/overlays/inventories/RarityColorOverlay.java
@@ -166,7 +166,7 @@ public class RarityColorOverlay implements Listener {
         if (lore.contains(ItemTier.MYTHIC.asFormattedName()) && UtilitiesConfig.Items.INSTANCE.mythicHighlight) {
             return ItemTier.MYTHIC.getCustomizedHighlightColor();
         }
-        if (name.matches("^(" + TextFormatting.DARK_AQUA + ".*%.*)$")) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/wynntils/modules/utilities/overlays/inventories/RarityColorOverlay.java
```diff
@@ -166,7 +166,7 @@ private static CustomColor getHighlightColor(Slot s, ItemStack is, String lore,
         if (lore.contains(ItemTier.MYTHIC.asFormattedName()) && UtilitiesConfig.Items.INSTANCE.mythicHighlight) {
             return ItemTier.MYTHIC.getCustomizedHighlightColor();
         }
-        if (name.matches("^(" + TextFormatting.DARK_AQUA + ".*%.*)$")) {
+        if (name.matches("^(" + TextFormatting.DARK_AQUA + ".*%.*)$") && UtilitiesConfig.Items.INSTANCE.craftedHighlight) {
```
kristofbolyai: Really small thing but maybe check  if crafted item highlights are on before doing regex?

McPlayHD: oh true we could actially change that everywhere since "contains" takes more CPU cycles than checking a simple boolean.

McPlayHD (commented):
I will change the order of boolean and "contains" and "match" checks for better performance.

McPlayHD:
I added a selector for the minimum tier ingredient to be highlighted.
image

HeyZeer0 (approved):
Awesome, LGTM!! 😄 
The reward menu is generating an INSANE amount of garbage we should definitely fix that in the future

PR #451, opened 2022-02-12
Adding soul points to /s command

PhoenixOrigin:
Made it so /s sp and /s nextsoulpoint will return 10 upcoming sp worlds


src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -38,7 +40,7 @@ public String getName() {
 
     @Override
     public String getUsage(ICommandSender sender) {
-        return "/s <command> [options]\n\ncommands:\nl,ls,list | list available servers\ni,info | get info about a server\n\nmore detailed help:\n/s <command> help";
+        return "/s <command> [options]\n\ncommands:\nl,ls,list | list available servers\ni,info | get info about a server\n\nmore detailed help:\n/s <command> help \nSkill Points:\nnextsoulpoint";
```
kristofbolyai: Please make this so it follows the  help menu standards

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -56,6 +58,12 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
                     case "i":
                         serverInfo(server, sender, Arrays.copyOfRange(args, 1, args.length));
                         break;
+                    case "nextsoulpoint":
+                        nextSoulPoint(server, sender, args);
+                        break;
+                    case "sp":
+                        nextSoulPoint(server, sender, args);
+                        break;
```
kristofbolyai: You can stack cases, no need to repeat code

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
kristofbolyai: Also make sure the help command includes the aliases as well.

src/main/java/com/wynntils/modules/utilities/managers/ServerListManager.java
```diff
@@ -14,10 +14,20 @@
 
     private static Map<String, ServerProfile> availableServers = new HashMap<>();
 
+    public static Map<String, ServerProfile> getAvailableServers(){
```
kristofbolyai: Please add space before `{`.

kristofbolyai:
Also please close #450 if this is the working PR.

PhoenixOrigin:
fixed everything i was told to do now complete i think

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -104,8 +104,11 @@ private void nextSoulPoint(MinecraftServer server, ICommandSender sender, String
             } else {
                 minuteColor = TextFormatting.RED;
             }
-
-            toSend.appendText("\n§b- §6" + wynnServer + " §b in " + minuteColor + uptimeMinutes + " minutes");
+            if(uptimeMinutes==1){
+                toSend.appendText("\n§b- §6" + wynnServer + " §b in " + minuteColor + uptimeMinutes + " minute");
+            }else{
+                toSend.appendText("\n§b- §6" + wynnServer + " §b in " + minuteColor + uptimeMinutes + " minutes");
+            }
```
kristofbolyai: Space after `if` and before and after brackets and `==` please. Just in general try to follow code standards.

kristofbolyai:
> fixed everything i was told to do now complete i think

I think you missed this one.
https://github.com/Wynntils/Wynntils/pull/451#discussion_r805192306

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +73,82 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                style.setHoverEvent(hoverEvent);
+                style.setInsertion("");
+                wc.setStyle(style);
+                sent.appendSibling(wc);
+            } else {
+                TextComponentString wc = new TextComponentString(TextFormatting.BLUE + wynnServer);
+                Style style = sent.getStyle();
+                style.setBold(false);
+                style.setItalic(false);
+                style.setStrikethrough(false);
```
kristofbolyai: No need to set these.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +73,82 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                style.setColor(TextFormatting.BLUE);
+                style.setUnderlined(false);
+                style.setObfuscated(false);
+                style.setInsertion("");
```
kristofbolyai: Same as above, no need to set these.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +73,82 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                HoverEvent hoverEvent = new HoverEvent(HoverEvent.Action.SHOW_TEXT, new TextComponentString(TextFormatting.RED + "You require Champion or Hero to access this feature."));
```
kristofbolyai: Change text to "HERO or higher rank is required to use /switch"

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +73,82 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                style.setHoverEvent(hoverEvent);
+                ClickEvent clickEvent = new ClickEvent(ClickEvent.Action.SUGGEST_COMMAND, "");
+                style.setClickEvent(clickEvent);
```
kristofbolyai: No need for these.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
kristofbolyai:
```suggestion
        return "/s <command> [options]\n\ncommands:\nl,ls,list | list available servers\ni,info | get info about a server\nsp,nextsoulpoint | list servers with soul point timers\n\nmore detailed help:\n/s <command> help";
```

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +73,82 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+            }
+            TextComponentString sent = new TextComponentString("§b- §6");
+            if(WebManager.getPlayerProfile() != null && (WebManager.getPlayerProfile().getTag() == PlayerStatsProfile.PlayerTag.HERO || WebManager.getPlayerProfile().getTag() == PlayerStatsProfile.PlayerTag.CHAMPION)) {
+                TextComponentString wc = new TextComponentString(TextFormatting.BLUE + wynnServer);
+                Style style = sent.getStyle();
+                style.setBold(false);
+                style.setItalic(false);
+                style.setStrikethrough(false);
+                style.setUnderlined(false);
+                style.setObfuscated(false);
```
kristofbolyai: Not needed.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +73,82 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+            int uptimeMinutes = sortedServers.get(wynnServer);
+            TextFormatting minuteColor;
+            if (uptimeMinutes <= 2) {
+                minuteColor = TextFormatting.GREEN;
+            } else if (uptimeMinutes <= 5) {
+                minuteColor = TextFormatting.YELLOW;
+            } else {
+                minuteColor = TextFormatting.RED;
+            }
+            TextComponentString sent = new TextComponentString("§b- §6");
```
kristofbolyai: Use `TextFormatting.BOLD` and `TextFormatting.GOLD`.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +73,82 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                style.setItalic(false);
+                style.setStrikethrough(false);
+                style.setUnderlined(false);
+                style.setObfuscated(false);
+                style.setColor(TextFormatting.BLUE);
+                ClickEvent clickEvent = new ClickEvent(ClickEvent.Action.RUN_COMMAND, "/switch " + wynnServer);
+                style.setClickEvent(clickEvent);
+                HoverEvent hoverEvent = new HoverEvent(HoverEvent.Action.SHOW_TEXT, new TextComponentString("Switch To " + TextFormatting.BLUE + wynnServer));
+                style.setHoverEvent(hoverEvent);
+                style.setInsertion("");
```
kristofbolyai: Not needed.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -38,8 +43,7 @@ public String getName() {
 
     @Override
     public String getUsage(ICommandSender sender) {
-        return "/s <command> [options]\n\ncommands:\nl,ls,list | list available servers\ni,info | get info about a server\n\nmore detailed help:\n/s <command> help";
-    }
+        return "/s <command> [options]\n\ncommands:\nl,ls,list | list available servers\ni,info | get info about a server\nsp,nextsoulpoint | list servers with soul point timers\n\nmore detailed help:\n/s <command> help";    }
```
kristofbolyai:
```suggestion
        return "/s <command> [options]\n\ncommands:\nl,ls,list | list available servers\ni,info | get info about a server\nsp,nextsoulpoint | list servers with soul point timers\n\nmore detailed help:\n/s <command> help";
        }
```

Sorry about this, my bad I think.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +73,78 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+            }
+        }
+
+        // Sort servers by time until soul point
+        Map<String, Integer> sortedServers = nextServers.entrySet().stream()
+                .sorted(Map.Entry.comparingByValue())
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (e1, e2) -> e1, LinkedHashMap::new));
+
+
+        TextComponentString toSend = new TextComponentString("\n" + "§bApproximate soul point times:" + "\n");
```
kristofbolyai: No need for a newline before the message. Also no need to set it to bold, you do it with a style later on.

```suggestion
        TextComponentString toSend = new TextComponentString("Approximate soul point times:" + "\n");
```

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +73,78 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+            if (uptimeMinutes <= 2) {
+                minuteColor = TextFormatting.GREEN;
+            } else if (uptimeMinutes <= 5) {
+                minuteColor = TextFormatting.YELLOW;
+            } else {
+                minuteColor = TextFormatting.RED;
+            }
+            TextComponentString sent = new TextComponentString(TextFormatting.BOLD + "-" + TextFormatting.GOLD);
+            if(WebManager.getPlayerProfile() != null && (WebManager.getPlayerProfile().getTag() == PlayerStatsProfile.PlayerTag.HERO || WebManager.getPlayerProfile().getTag() == PlayerStatsProfile.PlayerTag.CHAMPION)) {
+                TextComponentString wc = new TextComponentString(TextFormatting.BLUE + wynnServer);
```
kristofbolyai: Rename `wc` to `serverLine`.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +73,78 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                wc.getStyle()
+                .setBold(false)
+                .setItalic(false)
+                .setStrikethrough(false)
+                .setUnderlined(false)
+                .setObfuscated(false)
```
kristofbolyai: Try removing these, now with the refactoring it should not crash.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +73,78 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                .setColor(TextFormatting.BLUE)
+                .setClickEvent(new ClickEvent(ClickEvent.Action.RUN_COMMAND, "/switch " + wynnServer))
+                .setHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, new TextComponentString("Switch To " + TextFormatting.BLUE + wynnServer)))
+                .setInsertion("");
```
kristofbolyai: Remove this, same as above.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +73,78 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                sent.appendSibling(wc);
+            } else {
+                TextComponentString wc = new TextComponentString(TextFormatting.BLUE + wynnServer);
+                wc.getStyle()
+                .setBold(false)
+                .setItalic(false)
+                .setStrikethrough(false)
```
kristofbolyai: Also remove this.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +73,78 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                .setColor(TextFormatting.BLUE)
+                .setUnderlined(false)
+                .setObfuscated(false)
+                .setInsertion("")
```
kristofbolyai: Same as before.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +74,64 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                serverLine.getStyle()
+                .setColor(TextFormatting.BLUE)
+                .setClickEvent(new ClickEvent(ClickEvent.Action.RUN_COMMAND, "/switch " + wynnServer))
+                .setHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, new TextComponentString("Switch To " + TextFormatting.BLUE + wynnServer)));
+                sent.appendSibling(serverLine);
+            } else {
+                TextComponentString wc = new TextComponentString(TextFormatting.BLUE + wynnServer);
+                wc.getStyle()
+                .setHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, new TextComponentString(TextFormatting.RED + "HERO or higher rank is required to use /switch")));
+                sent.appendSibling(wc);
```
kristofbolyai: Rename `wc` here aswell.

kristofbolyai:
Rename that variable, otherwise looks good to me.

src/main/java/com/wynntils/modules/utilities/managers/ServerListManager.java
```diff
@@ -14,10 +14,20 @@
...
     public static synchronized void updateServers() {
         WebManager.getServerList((list) -> availableServers = list);
     }
 
+    public static int getUptimeTotalMinutes(String id) {
+        long milliseconds = System.currentTimeMillis() - availableServers.get(id).getFirstSeen();
+        return (int) (milliseconds / (1000 * 60));
+    }
+
+
```
HeyZeer0: Remove this empty white space
```suggestion
```

src/main/java/com/wynntils/modules/utilities/managers/ServerListManager.java
```diff
@@ -14,10 +14,20 @@
...
+    }
+
     public static synchronized void updateServers() {
         WebManager.getServerList((list) -> availableServers = list);
     }
 
+    public static int getUptimeTotalMinutes(String id) {
+        long milliseconds = System.currentTimeMillis() - availableServers.get(id).getFirstSeen();
+        return (int) (milliseconds / (1000 * 60));
+    }
```
HeyZeer0: This code should be inside the ServerProfile and use the methods already available in there. The name of the method is also inverted, something like this would work:
```java
    public static int getUptimeMinutes() {
        return (int) (getUptime() / TimeUnit.MINUTES.toMillis(1));
    }
```

src/main/java/com/wynntils/modules/utilities/managers/ServerListManager.java
```diff
@@ -14,10 +14,20 @@
 
     private static Map<String, ServerProfile> availableServers = new HashMap<>();
 
+    public static Map<String, ServerProfile> getAvailableServers() {
+        return availableServers;
+    }
+
```
HeyZeer0: This should be located below the ``getServer`` method.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +74,64 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (e1, e2) -> e1, LinkedHashMap::new));
+
+        TextComponentString toSend = new TextComponentString("Approximate soul point times:" + "\n");
+        toSend.getStyle()
+                .setBold(true)
+                .setColor(TextFormatting.AQUA);
+        sender.sendMessage(toSend);
+
+        List<String> keys = sortedServers.keySet().stream().limit(10).collect(Collectors.toList());
+
```
HeyZeer0, line 127: Remove this useless white space
```suggestion
```

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +74,64 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+        for (String wynnServer : keys) {
+            int uptimeMinutes = sortedServers.get(wynnServer);
+            TextFormatting minuteColor;
+            if (uptimeMinutes <= 2) {
+                minuteColor = TextFormatting.GREEN;
+            } else if (uptimeMinutes <= 5) {
+                minuteColor = TextFormatting.YELLOW;
+            } else {
+                minuteColor = TextFormatting.RED;
+            }
```
HeyZeer0: Space this
```suggestion

            TextFormatting minuteColor;
            if (uptimeMinutes <= 2) {
                minuteColor = TextFormatting.GREEN;
            } else if (uptimeMinutes <= 5) {
                minuteColor = TextFormatting.YELLOW;
            } else {
                minuteColor = TextFormatting.RED;
            }
            
```

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +74,64 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+            TextComponentString sent = new TextComponentString(TextFormatting.BOLD + "-" + TextFormatting.GOLD);
+            if(WebManager.getPlayerProfile() != null && (WebManager.getPlayerProfile().getTag() == PlayerStatsProfile.PlayerTag.HERO || WebManager.getPlayerProfile().getTag() == PlayerStatsProfile.PlayerTag.CHAMPION)) {
```
HeyZeer0: Space between the if and the condition
```suggestion
            if (WebManager.getPlayerProfile() != null && (WebManager.getPlayerProfile().getTag() == PlayerStatsProfile.PlayerTag.HERO || WebManager.getPlayerProfile().getTag() == PlayerStatsProfile.PlayerTag.CHAMPION)) {
```

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +74,64 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                .setHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, new TextComponentString(TextFormatting.RED + "HERO or higher rank is required to use /switch")));
+                sent.appendSibling(wc);
+            }
+
+            if (uptimeMinutes == 1) {
+                sent.appendText(" §b in " + minuteColor + uptimeMinutes + " minute");
+            } else {
+                sent.appendText(" §b in " + minuteColor + uptimeMinutes + " minutes");
+            }
+            sender.sendMessage(sent);
```
HeyZeer0: Space between the condition and the last action
```suggestion
            }
            
            sender.sendMessage(sent);
```

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +74,64 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
 
+    private void nextSoulPoint(MinecraftServer server, ICommandSender sender, String[] args){
+        Map<String, Integer> nextServers = new HashMap<>();
+
+        for (String availServer : ServerListManager.getAvailableServers().keySet()) {
+            int uptimeMinutes = ServerListManager.getUptimeTotalMinutes(availServer);
+            if ((uptimeMinutes % 20) >= 10) { // >= to 10min because we're looking for time UNTIL 20min, not after
+                nextServers.put(availServer, 20 - (uptimeMinutes % 20));
+            }
+        }
```
HeyZeer0: I feel like this could be done in a cleaner way, you don't exactly need the minutes, you can always just use the MS, but it is fair!

HeyZeer0 (changes requested):
Awesome!! Thanks for contributing to Wynntils. 
We appreciate the help, just some changes and this should be ready!

kristofbolyai:
@PhoenixImpulse Check all the open requests, I closed those which were fixed.

kristofbolyai:
Looks ok to me now.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +74,65 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+            if (uptimeMinutes <= 2) {
+                minuteColor = TextFormatting.GREEN;
+            } else if (uptimeMinutes <= 5) {
+                minuteColor = TextFormatting.YELLOW;
+            } else {
+                minuteColor = TextFormatting.RED;
+            }
+
+            TextComponentString world = new TextComponentString(TextFormatting.BOLD + "-" + TextFormatting.GOLD);
+            if (WebManager.getPlayerProfile() != null && (WebManager.getPlayerProfile().getTag() == PlayerStatsProfile.PlayerTag.HERO || WebManager.getPlayerProfile().getTag() == PlayerStatsProfile.PlayerTag.CHAMPION)) {
```
magicus: Personally I'd have extract this into a `boolean canUseSwitch` outside the for loop. Partly it's due to old school premature optimization thinking, but I also think it makes it clearer. You might want to get opinions from someone else than me on this, though...

magicus (commented):
Overall looks very good!

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
kristofbolyai: Maybe add it as a method in `PlayerTag`?

kristofbolyai:
Also small thing, could you rename the pull request? I think you accidentally misstyped it. Maybe something like "Adding soul point timers to /s" or something along those lines would be better. 👍

magicus:
@kristofbolyai Funny how you can be so blind to the obvious... I did not even see that. :) And yet "Do the patch do what it claims to do" is like the fundamental code reviewing rule.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -43,7 +44,7 @@ public String getName() {
 
     @Override
     public String getUsage(ICommandSender sender) {
-        return "/s <command> [options]\n\ncommands:\nl,ls,list | list available servers\ni,info | get info about a server\nsp,nextsoulpoint | list servers with soul point timers\n\nmore detailed help:\n/s <command> help";
+        return "/s <command> [options]\n\ncommands:\nl,ls,list | list available servers\ni,info | get info about a server\nsp,nextsoulpoint | list servers with soul point timers and if you add switch at the end you automatically switch to the world with the lowest soul point timer\n\nmore detailed help:\n/s <command> help";
```
kristofbolyai:
```suggestion
        return "/s <command> [options]\n\ncommands:\nl,ls,list | list available servers\ni,info | get info about a server\nsp,nextsoulpoint | list servers with soul point timers\n\nmore detailed help:\n/s <command> help";
```

kristofbolyai: Instead of putting that description there, maybe add a check if `args[1] == "help"` and make it display the help for the subcommand and talk about the "switch" argument there.

Example: 
```java
        TextComponentString text;
        if (options.contains("help")) {
            text = new TextComponentString(
                    "Usage: /s list [type] [options]\n order of types and options does not matter\nDefault: print all servers oldest to new\n\ntypes:\n");
            for (String type : serverTypes) {
                text.appendText(String.format("  %s\n", type));
            }
            text.appendText("options:\n");
            text.appendText("  g, group : group servers by type\n");
            text.appendText("  s, sort : sort servers alphabetically, sets group flag\n");
            text.appendText("  c, count : print amount of online servers\n");
            text.appendText("  h, help : this help\n");
            sender.sendMessage(text);
            return;
        }
```

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -89,48 +89,65 @@ private void nextSoulPoint(MinecraftServer server, ICommandSender sender, String
...
+                } catch (NullPointerException e) {
+                    server.createCommandManager().executeCommand(sender, "switch " + world);
+                }
             }
-
-            if (uptimeMinutes == 1) {
-                world.appendText(" §b in " + minuteColor + uptimeMinutes + " minute");
-            } else {
-                world.appendText(" §b in " + minuteColor + uptimeMinutes + " minutes");
+            else{
```
kristofbolyai:
```suggestion
            else {
```

kristofbolyai (changes requested):

kristofbolyai:
Please implement the requested changed, let me know if you have questions regarding them. Also, good job implementing `PlayerTag.canUseSwitch()`.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +75,86 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                    sender.getServer().getCommandManager().executeCommand(sender, "switch " + world);
+                } catch (NullPointerException e) {
+                    server.createCommandManager().executeCommand(sender, "switch " + world);
+                }
+            }
+            else {
+                sender.sendMessage(new TextComponentString("HERO or higher rank is required to use /switch"));
+            }
+        }
+
```
kristofbolyai, line 104: Remove this useless line.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +75,86 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+        else if (args[1].equalsIgnoreCase("help")) {
+            TextComponentString text = new TextComponentString("Usage: /s sp \nDefault: Prints the 10 worlds with the lowest soul point timers");
+            text.appendText("\nOptions");
+            text.appendText("\n  s, switch : automatically switch to the server with the lowest soul point timer");
+        }
+
```
kristofbolyai, line 115: Remove this useless line.

kristofbolyai (commented):
Those 2 small things, other than that it looks good!

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
kristofbolyai: Basically make it so the `}` and the `else` is in the same line.
Something like: `}  else if ...`

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +75,82 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+        if (args[1].equalsIgnoreCase("switch") || args[1].equalsIgnoreCase("s")) {
+            if (WebManager.getPlayerProfile().canUseSwitch()) {
+                String world = sortedServers.keySet().stream().limit(1).toString();
+                try {
+                    sender.getServer().getCommandManager().executeCommand(sender, "switch " + world);
+                } catch (NullPointerException e) {
+                    server.createCommandManager().executeCommand(sender, "switch " + world);
+                }
+            }
+            else {
```
kristofbolyai: My bad on this, should have explained this better.

```suggestion
} else {
```

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +75,81 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+
+        if (args[1].equalsIgnoreCase("switch") || args[1].equalsIgnoreCase("s")) {
+            if (WebManager.getPlayerProfile().canUseSwitch()) {
+                String world = sortedServers.keySet().stream().limit(1).toString();
+                try {
+                    sender.getServer().getCommandManager().executeCommand(sender, "switch " + world);
+                } catch (NullPointerException e) {
+                    server.createCommandManager().executeCommand(sender, "switch " + world);
+                }
+} else {
```
kristofbolyai:
```suggestion
            } else {
```

kristofbolyai:
Looks very good now!

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -88,21 +88,9 @@ private void nextSoulPoint(MinecraftServer server, ICommandSender sender, String
...
-                    sender.getServer().getCommandManager().executeCommand(sender, "switch " + world);
-                } catch (NullPointerException e) {
-                    server.createCommandManager().executeCommand(sender, "switch " + world);
-                }
-            } else {
-                sender.sendMessage(new TextComponentString("HERO or higher rank is required to use /switch"));
-            }
-        } else if (args[1].equalsIgnoreCase("help")) {
+        if (args[1].equalsIgnoreCase("help")) {
             TextComponentString text = new TextComponentString("Usage: /s sp \nDefault: Prints the 10 worlds with the lowest soul point timers");
```
kristofbolyai:
```suggestion
            TextComponentString text = new TextComponentString("Usage: /s sp \nDefault: Prints 10 worlds with increasing soul point timers");
```

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -17,11 +21,13 @@
 import net.minecraft.util.text.Style;
 import net.minecraft.util.text.TextComponentString;
 import net.minecraft.util.text.TextFormatting;
+import net.minecraft.util.text.event.ClickEvent;
 import net.minecraft.util.text.event.HoverEvent;
 import net.minecraftforge.client.IClientCommand;
+import scala.tools.nsc.doc.base.comment.Text;
```
kristofbolyai: Assume this was just left there unintentionally, could you remove it?

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +71,68 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+        }
+
+        // Sort servers by time until soul point
+        Map<String, Integer> sortedServers = nextServers.entrySet().stream()
+                .sorted(Map.Entry.comparingByValue())
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (e1, e2) -> e1, LinkedHashMap::new));
+
+        if (args[1].equalsIgnoreCase("help")) {
+            TextComponentString text = new TextComponentString("Usage: /s sp \nDefault: Prints 10 worlds with increasing lowest soul point timers");
+            sender.sendMessage(text);
```
HighCrit: Why do the work above, if you're not gonna use it?

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +71,68 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+        } else {
```
HighCrit: You can flatten the if-statement by using a return in the if block and remove the else. But this is just a personal preference.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +71,68 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+            TextComponentString soulPointInfo = new TextComponentString("Approximate soul point times:" + "\n");
```
HighCrit: Why have two separate strings?

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -56,6 +58,10 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
                     case "i":
                         serverInfo(server, sender, Arrays.copyOfRange(args, 1, args.length));
                         break;
+                    case "nextsoulpoint":
```
HighCrit: I feel like no one will use this long arg.

src/main/java/com/wynntils/webapi/profiles/ServerProfile.java
```diff
@@ -7,6 +7,7 @@
 import com.wynntils.core.utils.StringUtils;
 
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
```
HighCrit: This doesn't seem to be used.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +71,68 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                    minuteColor = TextFormatting.YELLOW;
+                } else {
+                    minuteColor = TextFormatting.RED;
+                }
+
+                TextComponentString world = new TextComponentString(TextFormatting.BOLD + "-" + TextFormatting.GOLD);
+                if (WebManager.getPlayerProfile().canUseSwitch()) {
+                    TextComponentString serverLine = new TextComponentString(TextFormatting.BLUE + wynnServer);
+                    serverLine.getStyle()
+                            .setColor(TextFormatting.BLUE)
```
HighCrit: Didn't you already set the colour in the constructor?

HighCrit (changes requested):
I'll be testing this somewhere later this week. I also strongly believe the for-loops and the use of various data structures can be lessened, which I will play with during testing.

kristofbolyai:
The request wanted you to move the help to the top of the function so you don't do useless computations, not remove the help.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +70,57 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
         }
     }
 
+    private void nextSoulPoint(MinecraftServer server, ICommandSender sender, String[] args){
```
HighCrit: `server` argument is unused and could therefore be removed

kristofbolyai:
What even is this feature exactly? https://github.com/Wynntils/Wynntils/pull/451/commits/e6ddb4371387896512b268dfaa44b539cf6429bf

Also, add some pull request description about it and change the PR name althought I don't really like that you kinda change up your mind mid PR.

kristofbolyai:
> What even is this feature exactly? [e6ddb43](https://github.com/Wynntils/Wynntils/commit/e6ddb4371387896512b268dfaa44b539cf6429bf)

Oh, I kinda see now. I don't see how it could be used tho.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -70,6 +73,71 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+            text.getStyle()
+                    .setColor(TextFormatting.RED);
+            sender.sendMessage(text);
+            return;
+        } else{
+            int interval = Integer.parseInt(args[1]);
+            for (String availableServer : ServerListManager.getAvailableServers().keySet()) {
+                nextServers.put(availableServer, interval - (ServerListManager.getServer(availableServer).getUptimeMinutes() % interval));
+            }
+        }
```
HighCrit: Or make the interval an optional variable with a default value of 20 minutes. Then this would be the exact same as the nextSoulPoint function

HighCrit (changes requested):

kristofbolyai:
No mobs spawn every x minutes tho? Rare spawns don't really have a timer but also like you can't know if someone spawned that rare mob.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +74,127 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+        if(args[1].equals(null)){
+            for (String availableServer : ServerListManager.getAvailableServers().keySet()) {
+                nextServers.put(availableServer, 20 - (ServerListManager.getServer(availableServer).getUptimeMinutes() % 20));
+            }
+        } else{
+            int interval = Integer.parseInt(args[1]);
+            for (String availableServer : ServerListManager.getAvailableServers().keySet()) {
+                nextServers.put(availableServer, interval - (ServerListManager.getServer(availableServer).getUptimeMinutes() % interval));
+            }
+        }
```
HighCrit: You can generify this if statement by moving the loop outside the else-block then in the if-block replace everything by `interval=20` which you'd need to declare above the if due to scopes.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +70,75 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
         }
     }
 
+    private void nextSoulPoint(ICommandSender sender, String[] args){
+
+        if (args[1].equalsIgnoreCase("help")) {
+            TextComponentString text = new TextComponentString("Usage: /s sp \nDefault: Prints 10 worlds with increasing lowest soul point timers");
+            sender.sendMessage(text);
```
Incompleteusern: Inline `text`

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +70,75 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+            return;
+        }
+
+        Map<String, Integer> nextServers = new HashMap<>();
+
+        int offset = 0;
+
+        if(args[1]!=null){
```
Incompleteusern: Please add spacing to this and the other `!=`

Incompleteusern: Another optional thing is to make this a ternary,  along with the other one

```java
offset = args[1] != null ? Integer.parseInt(args[1]) : 0;
```

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +70,75 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                        .setClickEvent(new ClickEvent(ClickEvent.Action.RUN_COMMAND, "/switch " + wynnServer))
+                        .setHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, new TextComponentString("Switch To " + TextFormatting.BLUE + wynnServer)));
+                world.appendSibling(serverLine);
+            } else {
+                TextComponentString serverLine = new TextComponentString(TextFormatting.BLUE + wynnServer);
+                serverLine.getStyle()
+                        .setHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, new TextComponentString(TextFormatting.RED + "HERO or higher rank is required to use /switch")));
+                world.appendSibling(serverLine);
+            }
+            if (uptimeMinutes == 1) {
```
Incompleteusern: This can also be a ternary

```java
world.appendText(" §b in " + minuteColor + uptimeMinutes + (uptimeMinutes == 1 ? " minute" : "minutes"));
```

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +70,75 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                world.appendText(" §b in " + minuteColor + uptimeMinutes + " minute");
+            } else {
+                world.appendText(" §b in " + minuteColor + uptimeMinutes + " minutes");
+            }
+            sender.sendMessage(world);
+            }
```
Incompleteusern: Untab this once

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +70,75 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+        int offset = 0;
+
+        if(args[1]!=null){
+            offset = Integer.parseInt(args[1]);
+        }
+
+        int interval = 20;
+
+        if(args[2]!=null){
+            interval = Integer.parseInt(args[2]);
```
Incompleteusern: I don't see the benefit of a non-20 interval, or an offset. Feel free to resolve this though

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +70,75 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+        }
+
+        for (String availableServer : ServerListManager.getAvailableServers().keySet()) {
+            nextServers.put(availableServer, (interval - (ServerListManager.getServer(availableServer).getUptimeMinutes() % interval)) + offset);
+        }
+
+        // Sort servers by time until soul point
+        Map<String, Integer> sortedServers = nextServers.entrySet().stream()
```
Incompleteusern, line 120: Perhaps use a `TreeMap` instead

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +70,75 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (e1, e2) -> e1, LinkedHashMap::new));
+
+        TextComponentString soulPointInfo = new TextComponentString("Approximate soul point times:" + "\n");
+        soulPointInfo.getStyle()
+                .setBold(true)
+                .setColor(TextFormatting.AQUA);
+        sender.sendMessage(soulPointInfo);
+        List<String> keys = sortedServers.keySet().stream().limit(10).collect(Collectors.toList());
+        for (String wynnServer : keys) {
+            int uptimeMinutes = sortedServers.get(wynnServer);
```
Incompleteusern: Use `entryset` instead of using keys. Another suggestion is to add an arg for the number of keys to limit to

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +70,75 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+            TextFormatting minuteColor;
+            if (uptimeMinutes <= 2) {
+                minuteColor = TextFormatting.GREEN;
+            } else if (uptimeMinutes <= 5) {
+                minuteColor = TextFormatting.YELLOW;
+            } else {
+                minuteColor = TextFormatting.RED;
+            }
+            TextComponentString world = new TextComponentString(TextFormatting.BOLD + "-" + TextFormatting.GOLD);
+            if (WebManager.getPlayerProfile().canUseSwitch()) {
```
Incompleteusern: Remove duplicated code
```java
TextComponentString serverLine = new TextComponentString(TextFormatting.BLUE + wynnServer);

if (WebManager.getPlayerProfile().canUseSwitch()) {        
    serverLine.getStyle()
            .setClickEvent(new ClickEvent(ClickEvent.Action.RUN_COMMAND, "/switch " + wynnServer))
            .setHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, new TextComponentString("Switch To " + TextFormatting.BLUE + wynnServer)));
} else {
    serverLine.getStyle()
            .setHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, new TextComponentString(TextFormatting.RED + "HERO or higher rank is required to use /switch")));

}

world.appendSibling(serverLine);

**Review by Incompleteusern — CHANGES_REQUESTED** [review]
Very cool and its great that you did this! However, I ask for some changes.

File: src/main/java/com/wynntils/modules/core/commands/CommandServer.java (hunk 47 above)
Comment (kristofbolyai) [reply src/main/java/com/wynntils/modules/core/commands/CommandServer.java]: Also make sure you are respecting the array size while using indexes, make sure `args.length >= 1`.

File: src/main/java/com/wynntils/modules/core/commands/CommandServer.java (hunk 51 above)
Comment (PhoenixOrigin) [reply src/main/java/com/wynntils/modules/core/commands/CommandServer.java:120]: What is the benefit of this?

File: src/main/java/com/wynntils/modules/core/commands/CommandServer.java [hunk 54]
```diff
@@ -17,11 +18,13 @@
 import net.minecraft.util.text.Style;
 import net.minecraft.util.text.TextComponentString;
 import net.minecraft.util.text.TextFormatting;
+import net.minecraft.util.text.event.ClickEvent;
 import net.minecraft.util.text.event.HoverEvent;
 import net.minecraftforge.client.IClientCommand;
+import org.w3c.dom.Text;
```
kristofbolyai: Unused import?

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +71,89 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
     }
 
+    private void nextSoulPoint(ICommandSender sender, String[] args){
+
+        if (args[1].equalsIgnoreCase("help")) {
+            sender.sendMessage(new TextComponentString("Usage: /s sp \nDefault: Prints 10 worlds with increasing lowest soul point timers"));
+            return;
+        }
+
+        HashMap<String, Integer> nextServers = new HashMap<>();
```
kristofbolyai: Use Map<String, Integer>

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
PhoenixOrigin:
```java
TextComponentString serverLine = new TextComponentString(TextFormatting.BLUE + wynnServer);

if (WebManager.getPlayerProfile().canUseSwitch()) {        
    serverLine.getStyle()
            .setClickEvent(new ClickEvent(ClickEvent.Action.RUN_COMMAND, "/switch " + wynnServer))
            .setHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, new TextComponentString("Switch To " + TextFormatting.BLUE + wynnServer)));
} else {
    serverLine.getStyle()
            .setHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, new TextComponentString(TextFormatting.RED + "HERO or higher rank is required to use /switch")));

}

world.appendSibling(serverLine);
```
This code is needed to be there? It makes it so if you have HERO rank or above, you can run /switch onclick but otherwise it shows you that you can't on hover

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
PhoenixOrigin: Also with this what is the benifit?

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
HighCrit: @Incompleteusern a TreeMap sorts the elements based on their keys not the values.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -73,40 +73,61 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
-        TextComponentString soulPointInfo = new TextComponentString("Approximate soul point times:" + "\n");
+        TextComponentString soulPointInfo = new TextComponentString("Approximate soul point times(" + offset + "min offset, -" + interval + "min interval" + "):" + "\n");
         soulPointInfo.getStyle()
                 .setBold(true)
                 .setColor(TextFormatting.AQUA);
         sender.sendMessage(soulPointInfo);
-        List<String> keys = sortedServers.keySet().stream().limit(10).collect(Collectors.toList());
+        if(sortedServers.keySet().size() < worlds){
+            worlds = sortedServers.keySet().size();
+        }
```
HighCrit: No need to reduce `worlds`'s value `.limit(n)` deals with situations when stream size is less than n

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -73,40 +73,61 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+        List<String> keys = sortedServers.entrySet().stream()
+                .limit(worlds)
+                .map(Map.Entry::getKey)
+                .collect(Collectors.toList());
```
HighCrit: Instead of `.collect` you can immediatly start the loop with `.forEach` and remove the assignment of the variable

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -73,40 +73,61 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
-        int interval = 20;
+            if (args.length >= 2) {
+                interval = args[2] != null ? Integer.parseInt(args[2]) : 0;
+            }
 
-        if(args[2]!=null){
-            interval = Integer.parseInt(args[2]);
+            if (args.length >= 3) {
+                worlds = args[3] != null ? Integer.parseInt(args[2]) : 0;
+            }
```
HighCrit: If length equals three then the minimum possible maximum index is 2. But you try and access index 3 here :?

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -73,40 +73,61 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
-        }
+        try{
+            if (args.length >= 1) {
+                offset = args[1] != null ? Integer.parseInt(args[1]) : 0;
+            }
 
-        int interval = 20;
+            if (args.length >= 2) {
+                interval = args[2] != null ? Integer.parseInt(args[2]) : 0;
+            }
```
HighCrit: If length equals two then the minimum possible maximum index is 1. But you try and access index 2 here :?

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -73,40 +73,61 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+        int interval = 20;
+        int worlds = 10;
 
-        if(args[1]!=null){
-            offset = Integer.parseInt(args[1]);
-        }
+        try{
+            if (args.length >= 1) {
+                offset = args[1] != null ? Integer.parseInt(args[1]) : 0;
+            }
```
HighCrit: If length equals one then the minimum possible maximum index is 0. But you try and access index 1 here :?

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -73,40 +73,61 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
 
-        int interval = 20;
+            if (args.length >= 2) {
+                interval = args[2] != null ? Integer.parseInt(args[2]) : 0;
+            }
 
-        if(args[2]!=null){
-            interval = Integer.parseInt(args[2]);
+            if (args.length >= 3) {
+                worlds = args[3] != null ? Integer.parseInt(args[2]) : 0;
```
HighCrit: You'll want to prevent negative numbers here, because IllegalArgumentExceptions will be thrown by `.limit(n)`

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -130,13 +151,9 @@ private void nextSoulPoint(ICommandSender sender, String[] args){
                         .setHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, new TextComponentString(TextFormatting.RED + "HERO or higher rank is required to use /switch")));
                 world.appendSibling(serverLine);
             }
-            if (uptimeMinutes == 1) {
-                world.appendText(" §b in " + minuteColor + uptimeMinutes + " minute");
-            } else {
-                world.appendText(" §b in " + minuteColor + uptimeMinutes + " minutes");
-            }
+            world.appendText(" §b in " + minuteColor + uptimeMinutes + (uptimeMinutes == 1 ? " minute" : "minutes"));
```
HighCrit: Please be consistent with how you use colour codes. As you've used `TextFormatting` for everything else instead of `§`

HighCrit (changes requested):

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
Incompleteusern: ah

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -90,7 +90,10 @@ private void nextSoulPoint(ICommandSender sender, String[] args){
 
         try{
             if (args.length >= 1) {
-                offset = args[1] != null ? Integer.parseInt(args[1]) : 0;
+                if(!args[1].equalsIgnoreCase("help")){
```
HighCrit: This should always be true since you returned online 82

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
Incompleteusern: It generalizes the assignment and allows for replacement with another subtype (`LinkedHashMap`) later

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
Incompleteusern: What does this mean?

HighCrit: No need to create a list and assign it to a variable. Since the variable is only used once in the loop in the next line. Therefore, we can immediately loop on the stream like:
```java
sortedServers.entrySet().stream()
  .limit(worlds)
  .map(Map.Entry::getKey)
  .forEach(server -> { ... });
```

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -71,15 +72,16 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
-        if (args[1].equalsIgnoreCase("help")) {
-            TextComponentString helpMessage = new TextComponentString("Usage: /s sp \nDefault: Prints 10 worlds with increasing lowest soul point timers");
-            helpMessage.appendText("Args:\n");
-            helpMessage.appendText("1: Offset for timers\n");
-            helpMessage.appendText("2: Interval for which to check\n");
-            helpMessage.appendText("3: How many worlds");
-            sender.sendMessage(helpMessage);
-            return;
+        if(args.length >= 2){
+            if (args[1].equalsIgnoreCase("help")) {
```
HighCrit: Combine these two if-statement

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -71,15 +72,16 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
-
-        if (args[1].equalsIgnoreCase("help")) {
-            TextComponentString helpMessage = new TextComponentString("Usage: /s sp \nDefault: Prints 10 worlds with increasing lowest soul point timers");
-            helpMessage.appendText("Args:\n");
-            helpMessage.appendText("1: Offset for timers\n");
-            helpMessage.appendText("2: Interval for which to check\n");
-            helpMessage.appendText("3: How many worlds");
-            sender.sendMessage(helpMessage);
-            return;
+        if(args.length >= 2){
```
HighCrit: Please change the condition to be `> 1` this makes it more consistent with the line below and is functionally the same.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -89,19 +91,19 @@ private void nextSoulPoint(ICommandSender sender, String[] args){
...
 
         try{
-            if (args.length >= 1) {
-                if(!args[1].equalsIgnoreCase("help")){
-                    offset = args[1] != null ? Integer.parseInt(args[1]) : 0;
-
-                }
-            }
-
             if (args.length >= 2) {
```
HighCrit: Please change the condition to be `> 1` this makes it more consistent with the line below and is functionally the same.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -89,19 +91,19 @@ private void nextSoulPoint(ICommandSender sender, String[] args){
...
+                offset = args[1] != null ? Integer.parseInt(args[1]) : 0;
+            }
+            if (args.length >= 3) {
```
HighCrit: Please change the condition to be `> 2` this makes it more consistent with the line below and is functionally the same.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -89,19 +91,19 @@ private void nextSoulPoint(ICommandSender sender, String[] args){
...
                 interval = args[2] != null ? Integer.parseInt(args[2]) : 0;
             }
 
-            if (args.length >= 3) {
+            if (args.length >= 4) {
```
HighCrit: Please change the condition to be `> 3` this makes it more consistent with the line below and is functionally the same.

HighCrit (changes requested):
Please make the future commit titles more descriptive.  Examples: "Implemented requested changes by HighCrit", "Incorporated feedback" or preferably "Fixed AIOOB, moved main loop"

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +70,98 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                    return;
+                }
+            }
+        } catch (NumberFormatException exception){
+            sender.sendMessage(new TextComponentString(TextFormatting.RED + "Please input valid numbers"));
+            return;
+        }
+
+        for (String availableServer : ServerListManager.getAvailableServers().keySet()) {
+            int time = interval - (ServerListManager.getServer(availableServer).getUptimeMinutes() % interval) - offset;
```
HighCrit: Why not loop over `ServerListManager.getAvailableServers()` such that you don't have to call `ServerListManager.getServer(...)` each iteration?

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
kristofbolyai: Yes, but it's duplicated, so try to extract it into a function or rewrite the code in a way that it is only there once.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -38,7 +41,7 @@ public String getName() {
 
     @Override
     public String getUsage(ICommandSender sender) {
-        return "/s <command> [options]\n\ncommands:\nl,ls,list | list available servers\ni,info | get info about a server\n\nmore detailed help:\n/s <command> help";
+        return "/s <command> [options]\n\ncommands:\nl,ls,list | list available servers\ni,info | get info about a server\nsp,nextsoulpoint | list servers with soul point timers\n\nmore detailed help:\n/s <command> help";
```
kristofbolyai: The help string is outdated, remove `nextsoulpoint`.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +71,87 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+            }
+            nextServers.put(entry.getKey(), time);
+        }
+
+        // Sort servers by time until soul point
+        Map<String, Integer> sortedServers = nextServers.entrySet().stream()
+                .sorted(Map.Entry.comparingByValue())
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (e1, e2) -> e1, LinkedHashMap::new));
+
+        TextComponentString soulPointInfo = args.length > 1 ? new TextComponentString("Approximate soul point times(-" + offset + "min offset, " + interval + "min interval" + "):" + "\n") : new TextComponentString("\"Approximate soul point times:");
```
kristofbolyai:
```suggestion
        TextComponentString soulPointInfo = args.length > 1 ? new TextComponentString("Approximate soul point times(-" + offset + "min offset, " + interval + "min interval" + "):" + "\n") : new TextComponentString("Approximate soul point times:");
```

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +71,87 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                    TextFormatting minuteColor;
+                    if (uptimeMinutes <= 2) {
+                        minuteColor = TextFormatting.GREEN;
+                    } else if (uptimeMinutes <= 5) {
+                        minuteColor = TextFormatting.YELLOW;
+                    } else {
+                        minuteColor = TextFormatting.RED;
+                    }
+                    TextComponentString world = new TextComponentString(TextFormatting.BOLD + "-" + TextFormatting.GOLD);
+                    if (WebManager.getPlayerProfile().canUseSwitch()) {
```
kristofbolyai: Nullcheck WebManager.getPlayerProfile() first.

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +71,87 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                        TextComponentString serverLine = new TextComponentString(TextFormatting.BLUE + server);
+                        serverLine.getStyle()
+                                .setClickEvent(new ClickEvent(ClickEvent.Action.RUN_COMMAND, "/switch " + server))
+                                .setHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, new TextComponentString("Switch To " + TextFormatting.BLUE + server)));
```
kristofbolyai:
```suggestion
                                .setHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, new TextComponentString("Switch to " + TextFormatting.BLUE + server)));
```

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +71,87 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
     }
 
+    private void nextSoulPoint(ICommandSender sender, String[] args){
+        if(args.length > 1 && args[1].equalsIgnoreCase("help")){
+            TextComponentString helpMessage = new TextComponentString("Usage: /s sp \nDefault: Prints 10 worlds with increasing lowest soul point timers");
+            helpMessage.appendText("Args:\n");
+            helpMessage.appendText("1: Offset for timers, default is 0\n");
+            helpMessage.appendText("2: Interval for which to check, default is 20\n");
+            helpMessage.appendText("3: How many worlds, default is 10");
+            sender.sendMessage(helpMessage);
```
kristofbolyai: Use something like:
```java
int messageId = Utils.getRandom().nextInt(Integer.MAX_VALUE);
ChatOverlay.getChat().printUnloggedChatMessage(new TextComponentString("Usage: /s info <serverID>"), messageId);
 ```

File: src/main/java/com/wynntils/modules/core/commands/CommandServer.java [hunk 75]
```diff
@@ -65,6 +71,87 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+            return;
+        }
+
+        Map<String, Integer> nextServers = new HashMap<>();
+
+        int offset = 0;
+        int interval = 20;
+        int worlds = 10;
+
+        try{
```
kristofbolyai:
```suggestion
        try {
```

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +71,87 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+            offset = args.length > 1 && args[1] != null ? Integer.parseInt(args[1]) : 0;
```
kristofbolyai: At this point, that null check is not needed.

```suggestion
            offset = args.length > 1 ? Integer.parseInt(args[1]) : 0;
```

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +71,87 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+            helpMessage.appendText("3: How many worlds, default is 10");
+            sender.sendMessage(helpMessage);
+            return;
+        }
+
+        Map<String, Integer> nextServers = new HashMap<>();
+
+        int offset = 0;
+        int interval = 20;
+        int worlds = 10;
```
kristofbolyai: Remove redundant initializers.

```suggestion
        int offset;
        int interval;
        int worlds;
```

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +71,87 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+
+        try{
+            offset = args.length > 1 && args[1] != null ? Integer.parseInt(args[1]) : 0;
+            interval = args.length > 2 && args[2] != null ? Integer.parseInt(args[2]) : 20;
+            worlds = args.length > 3 && args[3] != null ? Integer.parseInt(args[3]) : 10;
```
kristofbolyai: Also check for negative numbers.

image

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
kristofbolyai: Also extract the statement. Put this before `sortedServers.entrySet().stream()`.

```java
PlayerStatsProfile playerProfile = WebManager.getPlayerProfile();
boolean canUseSwitch = playerProfile != null && playerProfile.canUseSwitch();
 ```

File: src/main/java/com/wynntils/modules/core/commands/CommandServer.java [hunk 79]
```diff
@@ -65,6 +71,87 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                    }
+                    TextComponentString world = new TextComponentString(TextFormatting.BOLD + "-" + TextFormatting.GOLD);
+                    if (WebManager.getPlayerProfile().canUseSwitch()) {
+                        TextComponentString serverLine = new TextComponentString(TextFormatting.BLUE + server);
+                        serverLine.getStyle()
+                                .setClickEvent(new ClickEvent(ClickEvent.Action.RUN_COMMAND, "/switch " + server))
+                                .setHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, new TextComponentString("Switch To " + TextFormatting.BLUE + server)));
+                        world.appendSibling(serverLine);
+                    } else {
+                        TextComponentString serverLine = new TextComponentString(TextFormatting.BLUE + server);
```
kristofbolyai: You can declare `TextComponentString serverLine = new TextComponentString(TextFormatting.BLUE + server);` before the if statements.

kristofbolyai (changes requested):
General thing I found while testing is that this feature does not really work for soul points. Offsets vary be server because of the way Athena and Wynntils logs servers. Some timers were really off (even 5 minutes) and generally, even being 1 minute off can hurt this feature. I know we have optional offsets, they do not work since the server start time is offset by a random amount for each server, so there is not a general offset that could be used for this. I can't think of any solution to this as of right now.

PhoenixOrigin:
> General thing I found while testing is that this feature does not really work for soul points. Offsets vary be server because of the way Athena and Wynntils logs servers. Some timers were really off (even 5 minutes) and generally, even being 1 minute off can hurt this feature. I know we have optional offsets, they do not work since the server start time is offset by a random amount for each server, so there is not a general offset that could be used for this. I can't think of any solution to this as of right now.

Not quite sure why the timers were off, they were fine for me

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +72,91 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+
+        int offset;
+        int interval;
+        int worlds;
+
+        try {
+            offset = args.length > 1 ? Integer.parseInt(args[1]) : 0;
+            interval = args.length > 2 && args[2] != null ? Integer.parseInt(args[2]) : 20;
+            worlds = args.length > 3 && args[3] != null ? Integer.parseInt(args[3]) : 10;
+            if (offset < 0 || interval < 0){
```
kristofbolyai:
```suggestion
            if (offset < 0 || interval < 0) {
```

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +72,91 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                sender.sendMessage(new TextComponentString(TextFormatting.RED + "Please input valid numbers"));
+            }
+        } catch (NumberFormatException exception){
```
kristofbolyai:
```suggestion
        } catch (NumberFormatException exception) {
```

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +72,91 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+        }
+
+        for (Map.Entry<String, ServerProfile> entry : ServerListManager.getAvailableServers().entrySet()) {
+            int time = interval - (entry.getValue().getUptimeMinutes() % interval) - offset;
+            if (time < 0){
+                time = interval - Math.abs(time);
+            }
+            if(time>interval){
+                time = time-interval;
+            }
```
kristofbolyai:
```suggestion
            if (time > interval) {
                time = time-interval;
            }
```

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +72,91 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+        } catch (NumberFormatException exception){
+            sender.sendMessage(new TextComponentString(TextFormatting.RED + "Please input valid numbers"));
+            return;
+        }
+
+        for (Map.Entry<String, ServerProfile> entry : ServerListManager.getAvailableServers().entrySet()) {
+            int time = interval - (entry.getValue().getUptimeMinutes() % interval) - offset;
+            if (time < 0){
+                time = interval - Math.abs(time);
+            }
```
kristofbolyai:
```suggestion
            if (time < 0) {
                time = interval - Math.abs(time);
            }
```

kristofbolyai (changes requested):

src/main/java/com/wynntils/modules/core/commands/CommandServer.java
```diff
@@ -65,6 +72,91 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+            if(time>interval){
+                time = time-interval;
```
DevChromium:
```suggestion
                time = time - interval;
```

DevChromium (changes requested):

magicus:
It's kind a serious if the feature does not work as advertised. :-/ Do we need more testing to confirm if this does or does not work?

PhoenixOrigin:
> It's kind a serious if the feature does not work as advertised. :-/ Do we need more testing to confirm if this does or does not work?

Can you test it a bit? Whenever I have used it, it has been pretty acurate

Incompleteusern:
> > General thing I found while testing is that this feature does not really work for soul points. Offsets vary be server because of the way Athena and Wynntils logs servers. Some timers were really off (even 5 minutes) and generally, even being 1 minute off can hurt this feature. I know we have optional offsets, they do not work since the server start time is offset by a random amount for each server, so there is not a general offset that could be used for this. I can't think of any solution to this as of right now.
> 
> Not quite sure why the timers were off, they were fine for me

I really don't know why the offsets are off. Athena logs the join time by itself and logs it every 30 seconds, so this means that either the code is wrong or Wynn api is behind

PhoenixOrigin:
Closing PR since there are some differences that makes this useless and also avomod has it now

PR #452, opened 2022-02-12
Add the option to filter party finder and event messages

kristofbolyai:
Added the option to filter messages appearing in chat recommending the use of party finder. 
Example: `Party Finder: Hey bol, over here! Join the The Canyon Colossus queue and match up with 3 other players!`

Additionally, the order of conditions in `onChatReceived` was changed as I think it is more logical and faster to check if a feature is enabled before doing string operations.

EDIT: Also added the option to filter event messages from Wynncraft, by default, this feature is turned off.

### DIFF
--- a/src/main/java/com/wynntils/modules/chat/configs/ChatConfig.java
+++ b/src/main/java/com/wynntils/modules/chat/configs/ChatConfig.java
@@ -50,7 +50,7 @@ public class ChatConfig extends SettingsClass {
     public boolean filterTerritoryEnter = true;
 
     @Setting(displayName = "Filter Party Finder Messages", description = "Should Party Finder recommendation messages be hidden from chat?.", order = 12)
-    public boolean filterPartyFinder = true;
+    public boolean filterPartyFinder = false;
 
     @Setting(displayName = "Show Held Item Chat Message", description = "Should details of your compass and soul points be shown in chat while you are holding them?", order = 13)
     public boolean heldItemChat = true;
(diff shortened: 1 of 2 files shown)


src/main/java/com/wynntils/modules/chat/configs/ChatConfig.java
```diff
@@ -40,13 +40,19 @@
...
-    @Setting(displayName = "Filter Territory Enter", description = "Should territory enter messages be hidden from chat?\n\n§8Territory enter messages look like §7[You are now entering Detlas]§8.", order = 10)
+    @Setting(displayName = "Filter Event Messages", description = "Should Wynncraft join messages be hidden from chat?\n\n§8Messages starting with §6[Event] will no longer appear in chat.", order = 10)
+    public boolean filterEventMessages = false;
+
+    @Setting(displayName = "Filter Territory Enter", description = "Should territory enter messages be hidden from chat?\n\n§8Territory enter messages look like §7[You are now entering Detlas]§8.", order = 11)
     public boolean filterTerritoryEnter = true;
 
-    @Setting(displayName = "Show Held Item Chat Message", description = "Should details of your compass and soul points be shown in chat while you are holding them?", order = 12)
+    @Setting(displayName = "Filter Party Finder Messages", description = "Should Party Finder recommendation messages be hidden from chat?.", order = 12)
+    public boolean filterPartyFinder = true;
```
HeyZeer0: Default false, party finder messages are useful for new players and increase quite a lot the use of the party finder, they just need to be disabled if the person is absolutely sure they don't want to see it anymore!
```suggestion
    public boolean filterPartyFinder = false;
```

HeyZeer0 (changes requested):
Awesome! LGTM once the default option is switched 😄

PR #453, opened 2022-02-12
Added the option to highlight emeralds in chests

kristofbolyai:
There is also a setting to change the highlight color. This feature should greatly help lootrunners recognize emeralds.

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/configs/UtilitiesConfig.java
+++ b/src/main/java/com/wynntils/modules/utilities/configs/UtilitiesConfig.java
@@ -355,6 +355,9 @@ public class UtilitiesConfig extends SettingsClass {
         @Setting(displayName = "Highlight Normal Items", description = "Should normal items be highlighted?", order = 28)
         public boolean normalHighlight = false;
 
+        @Setting(displayName = "Highlight Emeralds in chests", description = "Should emeralds be highlighted in chests?", order = 29)
+        public boolean emeraldHighlightInChest = true;
+
         @Setting(displayName = "Highlight Black Market Cosmetics", description = "Should black market cosmetic items be highlighted?", order = 30)
         public boolean blackMarketEffectsHighlight = true;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/wynntils/modules/utilities/configs/UtilitiesConfig.java
```diff
@@ -355,6 +355,9 @@ public String format(double number) {
         @Setting(displayName = "Highlight Normal Items", description = "Should normal items be highlighted?", order = 28)
         public boolean normalHighlight = false;
 
+        @Setting(displayName = "Highlight Emeralds in chests", description = "Should emeralds be highlighted in chests?", order = 29)
+        public boolean emeraldHighlightInChest = true;
```
HeyZeer0: False by default, not useful for a lot of players, and might get a bit too overwhelming! The player can easily opt to toggle it on if needed

```suggestion
        public boolean emeraldHighlightInChest = false;
```

src/main/java/com/wynntils/modules/utilities/overlays/inventories/RarityColorOverlay.java
```diff
@@ -124,6 +124,9 @@ private static CustomColor getHighlightColor(Slot s, ItemStack is, String lore,
             if (UtilitiesConfig.Items.INSTANCE.filterEnabled && !professionFilter.equals("-") && lore.contains(professionFilter)) {
                 return UtilitiesConfig.Items.INSTANCE.professionFilterHighlightColor;
             }
+            if (UtilitiesConfig.Items.INSTANCE.emeraldHighlightInChest && is.getItem() == Items.EMERALD && s.slotNumber <= 35) {
```
HeyZeer0: Checking the slot number might cause some issues at specific inventories, not completely sure!

HeyZeer0 (changes requested):
Simple and convenient addition, nice work! LGTM once the default option is switched!

PR #456, opened 2022-02-14
Fix issues related to zooming on world map introduced by #430

kristofbolyai:
Using `System.nanoTime()` causes issues for some users, reverting the change and let's just use `McIf.getSystemTime()`. The old bug should not come back, as clamping takes care of that.

I don't see any obvious reason why `McIf.getSystemTime()` would work better, but using it solves the issue. Also note that I was not able to reproduce the issue myself (zooming works fine for me), but this revert was tested by someone with the issue and it fixed it for them. 

If there is any obvious thing I missed, please let me know.

EDIT: Link to the PR causing this issue: https://github.com/Wynntils/Wynntils/pull/430


magicus (approved):
LGTM.

PR #463, opened 2022-02-16
feat: add guild territory defense level sorting

ryanzhoudev:
Adds a button to only show a certain difficulty of guild territory defense on the guild map. Currently uses the search icon, though this can be changed to anything else.

This PR also does some refactoring and changes how MapButtons are handled. Tested in game; seems to work fine.

### DIFF
--- a/src/main/java/com/wynntils/modules/map/overlays/objects/MapButton.java
+++ b/src/main/java/com/wynntils/modules/map/overlays/objects/MapButton.java
@@ -13,7 +13,7 @@ import net.minecraft.init.SoundEvents;
 
 import java.util.List;
 import java.util.function.BiConsumer;
-import java.util.function.Function;
+import java.util.function.Supplier;
 
 import static net.minecraft.client.renderer.GlStateManager.*;
 
@@ -25,12 +25,12 @@ public class MapButton {
     MapButtonIcon icon;
     List<String> hoverLore;
     BiConsumer<MapButton, Integer> onClick;
-    Function<Void, Integer> state;
+    Supplier<Integer> state;
     Boolean isBool;
 
     ScreenRenderer renderer = new ScreenRenderer();
 
-    public MapButton(int posX, int posY, MapButtonIcon icon, List<String> hoverLore, Boolean isBool, Function<Void, Integer> state, BiConsumer<MapButton, Integer> onClick) {
+    public MapButton(int posX, int posY, MapButtonIcon icon, List<String> hoverLore, Boolean isBool, Supplier<Integer> state, BiConsumer<MapButton, Integer> onClick) {
         int halfWidth = icon.getWidth() / 2;
         int halfHeight = icon.getHeight() / 2;
 
@@ -53,7 +53,7 @@ public class MapButton {
     public void drawScreen(int mouseX, int mouseY, float partialTicks) {
         pushMatrix();
         {
-            if (this.isBool && !(state.apply(null) == 1)) {
+            if (this.isBool && state.get() != 1) {
                 color(.3f, .3f, .3f, 1f);
             } else {
                 color(1f, 1f, 1f, 1f);
@@ -75,7 +75,7 @@ public class MapButton {
     }
 
     public Boolean isEnabled() {
-        switch (state.apply(null)) {
+        switch (state.get()) {
             case 0:
                 return false;
             case 1:
--- a/src/main/java/com/wynntils/modules/map/overlays/objects/MapButton.java
+++ b/src/main/java/com/wynntils/modules/map/overlays/objects/MapButton.java
@@ -13,6 +13,7 @@ import net.minecraft.init.SoundEvents;
 
 import java.util.List;
 import java.util.function.BiConsumer;
+import java.util.function.Function;
 import java.util.function.Supplier;
 
 import static net.minecraft.client.renderer.GlStateManager.*;
@@ -26,11 +27,11 @@ public class MapButton {
     List<String> hoverLore;
     BiConsumer<MapButton, Integer> onClick;
     Supplier<Integer> state;
-    Boolean isBool;
+    Function<Void, Boolean> isEnabled;
 
     ScreenRenderer renderer = new ScreenRenderer();
 
-    public MapButton(int posX, int posY, MapButtonIcon icon, List<String> hoverLore, Boolean isBool, Supplier<Integer> state, BiConsumer<MapButton, Integer> onClick) {
+    public MapButton(int posX, int posY, MapButtonIcon icon, List<String> hoverLore, Function<Void, Boolean> isEnabled, BiConsumer<MapButton, Integer> onClick) {
         int halfWidth = icon.getWidth() / 2;
         int halfHeight = icon.getHeight() / 2;
 
@@ -41,8 +42,7 @@ public class MapButton {
 
         this.icon = icon;
         this.hoverLore = hoverLore;
-        this.isBool = isBool;
-        this.state = state;
+        this.isEnabled = isEnabled;
         this.onClick = onClick;
     }
 
@@ -53,10 +53,10 @@ public class MapButton {
     public void drawScreen(int mouseX, int mouseY, float partialTicks) {
         pushMatrix();
         {
-            if (this.isBool && state.get() != 1) {
-                color(.3f, .3f, .3f, 1f);
-            } else {
+            if (isEnabled.apply(null)) {
                 color(1f, 1f, 1f, 1f);
+            } else {
+                color(.3f, .3f, .3f, 1f);
             }
 
             // icon itself
@@ -75,14 +75,7 @@ public class MapButton {
     }
 
     public Boolean isEnabled() {
-        switch (state.get()) {
-            case 0:
-                return false;
-            case 1:
-                return true;
-            default:
-                return null; // Generally it shouldn't return null but if it does there's a problem with related code
-        }
+        return isEnabled.apply(null);
     }
 
     public int getStartX() {
--- a/src/main/java/com/wynntils/modules/map/overlays/objects/MapButton.java
+++ b/src/main/java/com/wynntils/modules/map/overlays/objects/MapButton.java
@@ -53,7 +53,7 @@ public class MapButton {
     public void drawScreen(int mouseX, int mouseY, float partialTicks) {
         pushMatrix();
         {
-            if (this.isBool && !Boolean.parseBoolean(String.valueOf(state.apply(null)))) {
+            if (this.isBool && !(state.apply(null) == 1)) {
                 color(.3f, .3f, .3f, 1f);
             } else {
                 color(1f, 1f, 1f, 1f);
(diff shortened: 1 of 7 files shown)


src/main/java/com/wynntils/modules/map/overlays/objects/MapButton.java
```diff
@@ -73,7 +75,7 @@ public void mouseClicked(int mouseX, int mouseY, int mouseButton) {
     }
 
     public Boolean isEnabled() {
-        return isEnabled.apply(null);
+        return Boolean.parseBoolean(String.valueOf(state.apply(null)));
```
kristofbolyai: This looks a bit strange, is there any better way to do this?

src/main/java/com/wynntils/modules/map/overlays/objects/MapTerritory.java
```diff
@@ -106,6 +106,24 @@ public TerritoryProfile getTerritory() {
...
+        } else if (def.contains("Medium")) {
+            return 3;
+        } else if (def.contains("High")) {
+            return 4;
+        } else if (def.contains("Very High")) {
+            return 5;
+        } else {
+            return -1;
+        }
+    }
```
kristofbolyai: Why does equals not work here? Perhaps strip color codes? Or is there an other reason (if so maybe state that in the comment)?

src/main/java/com/wynntils/modules/map/overlays/ui/GuildWorldMapUI.java
```diff
@@ -42,6 +42,9 @@
     private boolean showTradeRoutes = true;
     private boolean territoryManageShortcut = true;
 
+    MapButton cycleButton;
+    private int territoryDefenseFilter = 0; // 0=off; 1=verylow; 2=low; 3=med; 4=high; 5=veryhigh
```
kristofbolyai: Wouldn't an enum be more useful here?

src/main/java/com/wynntils/modules/map/overlays/ui/GuildWorldMapUI.java
```diff
@@ -57,37 +60,72 @@ public void initGui() {
...
+                return "§aLow";
+            case 3:
+                return "§eMedium";
+            case 4:
+                return "§cHigh";
+            case 5:
+                return "§cVery High";
+            default:
+                return "§7Off";
+        }
```
kristofbolyai: I personally prefer to use `TextFormatting` in cases like this, but it's up to you.

src/main/java/com/wynntils/modules/map/overlays/ui/GuildWorldMapUI.java
```diff
@@ -138,8 +176,19 @@ protected void drawIcons(int mouseX, int mouseY, float partialTicks) {
...
         if (showTradeRoutes) generateTradeRoutes();
 
-        territories.values().forEach(c -> c.drawScreen(mouseX, mouseY, partialTicks, showTerritory, resourceColors, !showOwners, showOwners));
-        territories.values().forEach(c -> c.postDraw(mouseX, mouseY, partialTicks, width, height));
+        HashMap<String, MapTerritory> renderableTerritories = new HashMap<>();
+        if (territoryDefenseFilter != 0) { // If active
+            for (Map.Entry<String, MapTerritory> territory : territories.entrySet()) {
+                if (territory.getValue().getDefenses() == territoryDefenseFilter) {
+                    renderableTerritories.put(territory.getKey(), territory.getValue());
+                }
```
kristofbolyai: From what I see here, there is no way to filter like `Medium or higher` territories?

kristofbolyai: Also, maybe try doing something about nesting.

kristofbolyai (commented):
Looks good mostly, noted some things you should consider

kristofbolyai:
Also, just to note, this PR `most likely` won't be merged in this release version. (Not anything related to the code, it looks awesome, we just want to follow our time schedule and don't want to merge anything without a few days of testing in CE)

src/main/java/com/wynntils/modules/map/overlays/objects/MapButton.java
ryanzhoudev: Oh yeah I forgot to change this after I discovered .parseBoolean doesn't actually convert 1s and 0s

src/main/java/com/wynntils/modules/map/overlays/objects/MapTerritory.java
ryanzhoudev: No idea, I even printed both the unformatted and formatted text - they yielded the same results (the strings in the if/else right now), and neither worked with .equals() - I can try again in a bit

src/main/java/com/wynntils/modules/map/overlays/ui/GuildWorldMapUI.java
ryanzhoudev: Yeah I'll change this

src/main/java/com/wynntils/modules/map/overlays/objects/MapTerritory.java
kristofbolyai: Use `StringUtils#stripColorCodes` from Minecraft to actually remove color codes, sometimes formatted and unformatted does not work as expected.

src/main/java/com/wynntils/modules/map/overlays/ui/GuildWorldMapUI.java
ryanzhoudev: No way to filter med+higher, we could try something like shift-click to view that diff and higher?

kristofbolyai: > No way to filter med+higher, we could try something like shift-click to view that diff and higher?

Yeah, maybe that, or some kind of button for that aswell?

Not sure, leaving it up to you, just try to make it as intuitive as possible.

src/main/java/com/wynntils/modules/map/overlays/ui/GuildWorldMapUI.java
ryanzhoudev: I think we should keep this as int; it'll be more consistent with the rest of the buttons
With an enum, it'd probably make more sense to create an enum for all the defense levels in MapTerritory and reference that, but that'd prevent us from having an off state. If we included an off state in the enum, future uses on that enum would be difficult as it'd have to excluded the off state.

src/main/java/com/wynntils/modules/map/overlays/ui/GuildWorldMapUI.java
kristofbolyai: @DonkeyBlaster Are you planning on working on this or are you done with this PR?

ryanzhoudev: Working, though I'm trying to fix my item highlights first

kristofbolyai (changes requested):
Looks good! 

I have found two things that I find a bit strange:

- Should we disable trade routes when using the filter? 
image
Doesn't that image look a bit weird? Maybe disable all trade routes or only have them between drawn territories? I am up for discussion on this one.

- A bug I found while using the filter: Guild map has the ability to run `/gu territory <territory>` when the user Shift + Right clicks on a territory. This currently works with territories that are filtered. I would expect this to only work with territories that are drawn on the screen.

ryanzhoudev:
I'll look into how both of these work - preferably, we'd have trade routes show for the territories that do appear.

kristofbolyai (approved):

src/main/java/com/wynntils/modules/map/overlays/objects/MapButton.java
```diff
@@ -22,25 +22,27 @@
...
-    Function<Void, Boolean> isEnabled;
+    Function<Void, Integer> state;
+    Boolean isBool;
 
     ScreenRenderer renderer = new ScreenRenderer();
 
-    public MapButton(int posX, int posY, MapButtonType type, List<String> hoverLore, Function<Void, Boolean> isEnabled, BiConsumer<MapButton, Integer> onClick) {
-        int halfWidth = type.getWidth() / 2;
-        int halfHeight = type.getHeight() / 2;
+    public MapButton(int posX, int posY, MapButtonIcon icon, List<String> hoverLore, Boolean isBool, Function<Void, Integer> state, BiConsumer<MapButton, Integer> onClick) {
```
Incompleteusern: replace `Function<Void, Integer> state` with `Supplier<Integer>`

src/main/java/com/wynntils/modules/map/overlays/ui/GuildWorldMapUI.java
```diff
@@ -57,37 +68,108 @@ public void initGui() {
...
+                territoryDefenseFilterType = TDFTypes.NORMAL;
+            }
+
+            if (territoryDefenseFilter >= 5 || (territoryDefenseFilterType != TDFTypes.NORMAL && territoryDefenseFilter == 4)) { // Safety check and reset
+                territoryDefenseFilter = 0;
+            } else {
+                ++territoryDefenseFilter;
+            }
+
+            // Remove redundant settings (Very low and lower, very high and lower, very low and higher, very high and higher)
```
Incompleteusern, line 133: Are `very high and lower` and `very high and higher` filtered/

Incompleteusern (changes requested):
a few things

src/main/java/com/wynntils/modules/map/overlays/ui/GuildWorldMapUI.java
ryanzhoudev: What do you mean by filtered? Currently it's just not possible to reach that filter state

Incompleteusern: wdym?

ryanzhoudev: Clicking through the options will skip the settings in that comment

Incompleteusern (approved):

src/main/java/com/wynntils/modules/map/overlays/ui/GuildWorldMapUI.java
```diff
@@ -57,37 +68,108 @@ public void initGui() {
...
+                type = RESET + " and higher";
+                break;
+            case LOWER:
+                type = RESET + " and lower";
+                break;
+            case NORMAL:
+            default:
+                type = "";
+                break;
+        }
```
HighCrit: Personally would've opted for a method in the enum

src/main/java/com/wynntils/modules/map/overlays/ui/WorldMapUI.java
```diff
@@ -123,15 +124,17 @@ public void initGui() {
         Keyboard.enableRepeatEvents(true);
     }
 
-    protected void addButton(MapButtonType type, int offsetX, List<String> hover, Function<Void, Boolean> isEnabled, BiConsumer<MapButton, Integer> onClick) {
+    protected MapButton addButton(MapButtonIcon icon, int offsetX, List<String> hover, Boolean isBool, Supplier<Integer> state, BiConsumer<MapButton, Integer> onClick) {
```
HighCrit: Why did state become an integer if all use cases are in conditions?

ryanzhoudev: I don't actually remember too well - I think I had a reason to do with how I was going to inform the user of which filter state was currently selected but the current implementation doesn't need it. I'll look at reverting it back to what we had before.

PR #466, opened 2022-02-22
fix: add min opacity to highlights

ryanzhoudev:
Anything lower than the limits are invisible.


src/main/java/com/wynntils/modules/utilities/configs/UtilitiesConfig.java
```diff
@@ -439,11 +439,11 @@ public String format(double number) {
         public CustomColor ingredientThreeHighlightColor = new CustomColor(0.9f, .3f, 0);
 
         @Setting(displayName = "Inventory Highlight Opacity", description = "As a percentage, how opaque should highlights in your inventory be?", order = 62)
-        @Setting.Limitations.FloatLimit(min = 0, max = 100, precision = 0.5f)
+        @Setting.Limitations.FloatLimit(min = 29, max = 100, precision = 0.5f)
```
kristofbolyai: Set min to 30 :)

kristofbolyai (approved):

kristofbolyai:
Okay, this commit looks fine but I remember some crashes related to settings min value. I believe if we modify the min value with this PR, everyone having a setting below that value will get stuck in a crash loop. I will try to confirm or deny this, until that let's not merge.

kristofbolyai:
> Okay, this commit looks fine but I remember some crashes related to settings min value. I believe if we modify the min value with this PR, everyone having a setting below that value will get stuck in a crash loop. I will try to confirm or deny this, until that let's not merge.

Actually, this seems to work fine when I tried it. Maybe I am misremembering.

Incompleteusern (approved):

PR #467, opened 2022-02-24
fix: Catch an exception if discord presence cannot load.

kristofbolyai:
Logs from multiple people: 
https://hst.sh/uhujaqecic
https://discord.com/channels/394189072635133952/903833202580406292/904046918731198534


kristofbolyai:
Actually this seems not to work. Let me investigate. Could someone with more knowledge about this code check this out?

PR #470, opened 2022-03-05
fox: lootrun path chest boxes render distance

McPlayHD:
I noticed that chest boxes would also render very far away so I made it so that they no longer render in case the chest is outside of tile entity render distance (64 blocks).
This should improve performance a bit especially for big lootrun files.


Incompleteusern (approved):
It may be better to do some caching, but LGTM!

kristofbolyai (approved):

PR #474, opened 2022-03-10
feat: Change health pot blocking to threshold based

ryanzhoudev:
Blocks health pots when player health higher than % specified. 0 to disable, default 99%.


src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
```diff
@@ -846,44 +846,44 @@ private static void checkLockState(int slot) {
...
             return;
 
         EntityPlayerSP player = McIf.player();
-        if (player.getHealth() != player.getMaxHealth()) return;
+        if (player.getHealth() / player.getMaxHealth() * 100 < UtilitiesConfig.INSTANCE.blockHealingPotThreshold) return;
 
         e.setCanceled(true);
-        McIf.mc().addScheduledTask(() -> GameUpdateOverlay.queueMessage(TextFormatting.DARK_RED + "You are already at full health!"));
+        McIf.mc().addScheduledTask(() -> GameUpdateOverlay.queueMessage(TextFormatting.DARK_RED + "You already have more than " + UtilitiesConfig.INSTANCE.blockHealingPotThreshold + "% health!"));
     }
```
kristofbolyai: Could you extract the duplicated code into a function of its own?

kristofbolyai (approved):

src/main/java/com/wynntils/modules/utilities/configs/UtilitiesConfig.java
```diff
@@ -95,8 +95,9 @@
     @Setting(displayName = "Auto Mount Horse", description = "Should you mount your horse automatically when it is summoned?", order = 2)
     public boolean autoMount = false;
 
-    @Setting(displayName = "Block Health Potions", description = "Should the mod prevent you from using your health potions when you are at full health?", order = 1)
-    public boolean blockHealingPots = true;
+    @Setting.Limitations.IntLimit(min = 0, max = 100)
+    @Setting(displayName = "Health Potion Blocking Threshold", description = "When you have more health than the percentage specified here, Wynntils will block health potions.\n\n§8Set to 0 to disable.", order = 1)
+    public int blockHealingPotThreshold = 99;
```
Incompleteusern, line 103: Why is the default value 99? Also the description seems to imply that it only blocks when you have more than the threshold, while it seems that having exactly the same as the threshold also works.

ryanzhoudev: 99 because that's what I originally put in the #suggestions channel, I just carried it over. Was there another preferred value?

Kepler-17c: Instead of some arbitrary value, wouldn't it be better making it relative to the next consumed potion?

ryanzhoudev: What do you mean relative to the next consumed potion? Like just have it outright block all potion uses unless they're going to be completely effective?

Kepler-17c: Yes, like that: Either configurable or a fixed relative value. Would block if you don't use at least x% of the potion.

ryanzhoudev: Wouldn't this completely break it for those going for eg. no armour runs? Those usually only have like ~500hp even at lv90+, and therefore potions at that level would be unusable for them

Kepler-17c: Mmm, yes, should be at least toggleable then 🤔

ryanzhoudev: Alternatively, we could have an enum switch kind of thing
Off - potions are never blocked
Health% - block potions when health is above certain %
PotionEffective% - block potions unless they're more than % effective
This could even be done with the same % slider

Kepler-17c: Oh, yea, that sounds great 👍

src/main/java/com/wynntils/modules/utilities/configs/UtilitiesConfig.java
```diff
@@ -181,6 +185,13 @@ public enum BankButtonSetting {
         Block
     }
 
+    public enum PotionBlockingType {
+        HealthPercent,
+        EffectivePercent,
+        Never,
+        Always
```
magicus: Why would anyone want "Always"?

ryanzhoudev: Maybe for a no pot challenge or something, where the player wouldn't want to accidentally use pots dropped by mobs

magicus: Is that a thing? Otherwise it seems like a stupid option. We don't need to be have all kinds of values. It is okay to have Never without having Always. But if you say that some players seriously would want that, well, keep it then.

ryanzhoudev: Ok coming back to this it's kind of redundant since you can just set HealthPercent to 0% and it'll always block anyway

magicus (approved):

PR #476, opened 2022-03-14
New Chat Tab Preset and updated Guild Tab Preset

TomR24:
I made some minor changes to the default Guild Preset so that it also redirects messages when "Use Guild Name Roles" is enabled. I also updated  the "Local" Preset found in the Tab Settings GUI to include messages from friends as well (not sure if it was doing that before but it sure does so now).

I also added a new chat tab preset "c". It has the "All" Tab from which all messages can be seen (which is why all tabs are set to Low Priority), as well as:
- Guild and Party Tab (grabbed from the updated DEFAULT_GUILD_REGEX and DEFAULT_PARTY_REGEX respectively)
- Bombs Tab (from #chat-tab-stock from the Wynntils Discord), to which I added the /switch auto-command
- Global Tab (taken from the updated "Local" Preset and changed to include messages from other worlds), which contains all messages sent by players from other worlds as well as your own world
- Local Tab (taken from the updated "Local" Preset), which contains only the messages sent from players of the world you are on
- Private Tab (taken from the "Private" Tab from Config Preset "b")
- Shouts Tab (from #chat-tab-stock from the Wynntils Discord)

I know this preset might seem a bit overkill, but players can remove tabs they don't want/need, and they will still be able to see those messages in the "All" Tab without tweaking the settings. On top of that, players can disable the low priority mode on certain tabs for messages they don't want to see in the "All" Tab. If the /switch auto-command being set by default is not complaint with Wynncraft's TOS it can be removed.


TomR24:
Don't worry about my last pull request, little repository oopsie

src/main/java/com/wynntils/modules/chat/managers/TabManager.java
```diff
@@ -63,6 +63,16 @@ public static void registerPresets() {
                     availableTabs.add(new ChatTab("G/P", DEFAULT_GUILD_REGEX + "|" + DEFAULT_PARTY_REGEX, null, "/g", false, 2));
                     availableTabs.add(new ChatTab("Private", "&7\\[.*\u27A4.*&7\\]", null, "/r", false, 3));
                     break;
+                case c:
+                    availableTabs.add(new ChatTab("All", ".*", null, "", true, 0));
```
kristofbolyai, line 67: This shouldn't be low priority?

kristofbolyai: Actually, nvm, this seems to work fine.

kristofbolyai (approved):

Incompleteusern (approved):

PR #478, opened 2022-03-22
Ingredients sort/filter

chrisstar321:
Add the same sort/filter functionality that the items guide has for the ingredients guide.

Change the 'Favourited' sort method so that it actually filters instead of sorts.


src/main/java/com/wynntils/webapi/profiles/ingredient/enums/ProfessionType.java
```diff
@@ -25,4 +32,35 @@ public String getDisplayName() {
...
+
+    public int getTextureX() {
+        return textureX;
+    }
+
+    public int getTextureY() {
+        return textureY;
+    }
+
+    public static ProfessionType from(String type) {
```
kristofbolyai: Rename this method to something clearer. `fromString` maybe?

kristofbolyai (commented):
Imo, favorited filter should only show favorited items. I know that this is not how it is in the normal items page, perhaps that should be changed as well. I would prefer feedback about this from others on the team though.
image

src/main/java/com/wynntils/webapi/profiles/ingredient/enums/ProfessionType.java
chrisstar321: The same method in `ItemType` is also called `from`. Should i change that one to `fromString` as well?

chrisstar321:
> Imo, favorited filter should only show favorited items. I know that this is not how it is in the normal items page, perhaps that should be changed as well. I would prefer feedback about this from others on the team though. image

Technically, the 'Favorited Items' is a sorting method and not a filter, though

src/main/java/com/wynntils/webapi/profiles/ingredient/enums/ProfessionType.java
kristofbolyai: Sure

kristofbolyai:
> > Imo, favorited filter should only show favorited items. I know that this is not how it is in the normal items page, perhaps that should be changed as well. I would prefer feedback about this from others on the team though. image
> 
> Technically, the 'Favorited Items' is a sorting method and not a filter, though

Yeah, it's kinda a strange one Imo, we will see what others think.

src/main/java/com/wynntils/core/utils/helpers/IngredientFilter.java
```diff
@@ -0,0 +1,498 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+import java.util.*;
+import java.util.function.Predicate;
+import java.util.function.ToIntFunction;
+import java.util.stream.Collectors;
+
+public interface IngredientFilter extends Predicate<IngredientProfile>, Comparator<IngredientProfile> {
```
Incompleteusern, line 28: surely its possible to have a general Filter class which both extend?

Incompleteusern, line 28: it might be nice to have a general interface which both extend

src/main/java/com/wynntils/webapi/profiles/ingredient/enums/IngredientTier.java
```diff
@@ -28,4 +28,31 @@ public int getTierInt() {
     public String getTierString() {
         return tierString;
     }
+
+    public static IngredientTier matchText(String tier) {
+        switch (tier) {
+            case "0":
+            case "common":
```
Incompleteusern, line 35: who does this?

chrisstar321: i used the same aliases as item rarities here, just in case someone wants to use those

src/main/java/com/wynntils/core/utils/helpers/IngredientFilter.java
chrisstar321: i thought so too. I've been working on it a little, but there's a lot of small differences between items and ingredients, and i don't have a lot of free time, so that might take a while

Incompleteusern: ah, its fine both ways

magicus:
The file profession_icons.png seem to have a lot more than the profession icons in it? In fact, it looks like it could be copied from some other texture file. Is that the case? If so, it's better to use the icons directly from that texture.

magicus:
I've never used the ingredient guide, and have no strong opinion on favorite filters. Can't say much more about this PR.

Incompleteusern (approved):

src/main/java/com/wynntils/webapi/profiles/ingredient/enums/ProfessionType.java
```diff
@@ -1,21 +1,28 @@
...
+    COOKING("Cooking", "Ⓐ", 16 * 11, 0),
+    ALCHEMISM("Alchemism", "Ⓛ", 16 * 10, 0),
+    SCRIBING("Scribing", "Ⓔ", 16 * 9, 0);
 
     final String professionName;
     final String professionIconChar;
 
-    ProfessionType(String professionName, String professionIconChar) {
+    int textureX;
+    int textureY;
```
kristofbolyai: These can be made final?

kristofbolyai (changes requested):
Fix this text's position, then we are good to merge. 
image

kristofbolyai (approved):

PR #479, opened 2022-03-26
Fix quest links not opening in quest book for people using old Java

kristofbolyai:
Fixes `[wynntils-web-request-pool-0/INFO] [wynntils]: Error occurred whilst fetching WikiQuestQuery from https://wynncraft.gamepedia.com/index.php?title=Special:CargoExport&format=json&tables=Quests&fields=Quests._pageTitle&where=Quests.name=%27Poisoning+the+Pest%27: sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target` by
registering Lets Encrypt certificate on startup if java version is old.


src/main/java/com/wynntils/webapi/WebManager.java
```diff
@@ -953,4 +963,75 @@ public static String getApiUrl(String key) {
...
+                } catch (NumberFormatException e) {
+                    e.printStackTrace();
+                    Reference.LOGGER.warn("Could not parse java version!");
+                }
+            }
+            return true;
+        }
+
+        public static void registerCerts() {
+            if (needsPatching()) return;
```
DevScyu: Is this correct? Feels like it should read like this:
```java
            if (!needsPatching()) return;
``` 

of course, also update the function above to follow

DevScyu (changes requested):

src/main/java/com/wynntils/webapi/WebManager.java
kristofbolyai: Messed up the name, but I feel like it is fitting, so I changed how the method works.

DevScyu (approved):

src/main/java/com/wynntils/webapi/WebManager.java
```diff
@@ -953,4 +963,78 @@ public static String getApiUrl(String key) {
         return apiUrls.get(key);
     }
 
+    public static class FixSSL {
+
+        private static boolean needsPatching()
+        {
+            String javaVersion = System.getProperty("java.version");
+            Reference.LOGGER.info("Found java version " + javaVersion);
```
Incompleteusern, line 971: should this be merged

kristofbolyai: Are you suggesting that the java version checking could be done better or are you talking about the whole PR?

Incompleteusern: Im talking about the debug

kristofbolyai: I mean there is no harm in logging that if we need to see user logs?

Incompleteusern: sure

Incompleteusern (approved):

magicus:
An alternative approach (which I argue we should use 100% in Artemis) is to let Athena query the wiki, and then publish the data for the mod from Athena instead.

In this case your solution is probably the simplest, but I think this issue really highlights why this is important.

PR #483, opened 2022-04-12
feat: Dungeon key enhancements

ryanzhoudev:
fix: Fix dungeon key specs in TM
feat: add middle click on key to set compass

Adds DungeonKeyManager.java with similar methods to cork amps/pouches.

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
+++ b/src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
@@ -15,10 +15,13 @@ import com.wynntils.core.framework.interfaces.Listener;
 import com.wynntils.core.utils.ItemUtils;
 import com.wynntils.core.utils.StringUtils;
 import com.wynntils.core.utils.Utils;
+import com.wynntils.core.utils.objects.Location;
+import com.wynntils.core.utils.objects.Pair;
 import com.wynntils.core.utils.reference.EmeraldSymbols;
 import com.wynntils.core.utils.reference.RequirementSymbols;
 import com.wynntils.modules.chat.overlays.ChatOverlay;
 import com.wynntils.modules.chat.overlays.gui.ChatGUI;
+import com.wynntils.modules.core.managers.CompassManager;
 import com.wynntils.modules.core.overlays.inventories.ChestReplacer;
 import com.wynntils.modules.core.overlays.inventories.HorseReplacer;
 import com.wynntils.modules.core.overlays.inventories.InventoryReplacer;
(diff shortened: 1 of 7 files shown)


src/main/java/com/wynntils/modules/utilities/managers/DungeonKeyManager.java
```diff
@@ -0,0 +1,77 @@
...
+        public final String acronym;
+        public final String fullname;
+
+        DungeonKey(String acronym, String fullname) {
+            this.acronym = acronym;
+            this.fullname = fullname;
+        }
+    }
+
+    public static boolean isDungeonKey(ItemStack i) {
```
kristofbolyai: Can you rename this to `is` or `stack` or something that is not `i`? I think most people use `i` as a loop variable and this seems strange to me. Same with methods below.

src/main/java/com/wynntils/modules/utilities/managers/DungeonKeyManager.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class DungeonKeyManager {
+
+    private static final Pattern NORMAL_KEY_PATTERN = Pattern.compile("§7Grants access to the§f(.+)§7 once");
+    private static final Pattern BROKEN_KEY_PATTERN = Pattern.compile("§7Use this item at the§fForgery§7 to craft a§4(.+) Key");
+
+    public enum DungeonKey {
```
kristofbolyai, line 24: About this whole enum, perhaps remove the corrupted ones, add a corrupted bool and make a getter for getting the full name (and make fullname private)? Same with acronym.

kristofbolyai, line 24: About this whole enum, perhaps remove the corrupted ones, add a corrupted bool and make a getter for getting the full name (and make fullname private)? Same with acronym.

src/main/java/com/wynntils/modules/utilities/managers/DungeonKeyManager.java
```diff
@@ -0,0 +1,77 @@
...
+        CDecrepitSewers("DS", "Corrupted Decrepit Sewers"),
+        CInfestedPit("IP", "Corrupted Infested Pit"),
+        CLostSanctuary("LS", "Corrupted Lost Sanctuary"),
+        CUnderworldCrypt("UC", "Corrupted Underworld Crypt"),
+        CSandSweptTob("ST", "Corrupted Sand-Swept Tomb"),
+        CIceBarrows("IB", "Corrupted Ice Barrows"),
+        CUndergrowthRuins("UR", "Corrupted Undergrowth Ruins");
+
+
+        public final String acronym;
```
kristofbolyai: Use a public getter and make this field a private. That is how java code standard goes.

```
Very evil: public fields.
Somewhat evil: Getters and setters where they're not required.
Good: Getters and setters only where they're really required - make the type expose "larger" behaviour which happens to use its state, rather than just treating the type as a repository of state to be manipulated by other types.
```

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
```diff
@@ -662,6 +665,18 @@ public void keyPressOnHorse(GuiOverlapEvent.HorseOverlap.KeyTyped e) {
...
+            ItemStack is = e.getGui().getSlotUnderMouse().getStack();
+            if (DungeonKeyManager.isDungeonKey(is)) {
+                Pair<Integer, Integer> coords = DungeonKeyManager.getDungeonCoords(is);
+                CompassManager.setCompassLocation(new Location(coords.a, 0, coords.b));
+                boolean isCorrupted = DungeonKeyManager.isCorrupted(is);
+                String location = (isCorrupted) ? "§4The Forgery" : "§6" + DungeonKeyManager.getDungeonKey(is).getFullName(false);
+                GameUpdateOverlay.queueMessage("§aSet compass to " + location);
+            }
+        }
+
```
kristofbolyai: This code seems duplicated. Perhaps make a method for it and replace this and the latter duplication with the call.

src/main/java/com/wynntils/modules/utilities/managers/DungeonKeyManager.java
```diff
@@ -0,0 +1,107 @@
...
+            Matcher brokenMatcher = BROKEN_KEY_PATTERN.matcher(lore);
+            if (!brokenMatcher.find()) {
+                // For broken keys, eg. from raid rewards
+                result = is.getDisplayName().replace("Broken ", "").replace(" Key", "");
+                if (!result.contains("Corrupted")) {
+                    result = "Corrupted " + result;
+                }
+            } else {
+                result = brokenMatcher.group(1);
+            }
```
kristofbolyai: This could be replaced with a better regex I think. Don't spend much time on it though if you don't know regex that well. I might add a commit in the PR later on to change this, if you don't plan to do so.

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
ryanzhoudev: I remember trying to fix this while writing it - I think I couldn't write a function for it because the two events use different packet types or something. I'll look into it again

kristofbolyai: Then leave the checks and pass itemstack into the method

kristofbolyai (approved):
Looks good now. Nice work.

magicus (approved):

PR #507, opened 2022-05-02
Auto clear lootruns on delete

PhoenixOrigin:
Pretty simple, just a small addition that to my understanding should clear the lootrun shown if you use delete.

### DIFF
--- a/src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
+++ b/src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
@@ -128,7 +128,7 @@ public class CommandLootRun extends CommandBase implements IClientCommand {
                 String message;
                 if (result) {
                     message = GREEN + "Deleted run " + name + " successfully!";
-                    if (LootRunManager.getActivePath() != null && LootRunManager.getActivePathName().equals(name) {
+                    if (LootRunManager.getActivePath() != null && LootRunManager.getActivePathName().equals(name)) {
                         LootRunManager.clear();
                     }
                 } else {
--- a/src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
+++ b/src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
@@ -132,7 +132,7 @@ public class CommandLootRun extends CommandBase implements IClientCommand {
                         LootRunManager.clear();
                     }
                 } else {
-                    message = RED + "The provided lootrun doesn't exists!";
+                    message = RED + "The provided lootrun doesn't exist!";
                 }
 
                 sender.sendMessage(new TextComponentString(message));
--- a/src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
+++ b/src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
@@ -133,7 +133,7 @@ public class CommandLootRun extends CommandBase implements IClientCommand {
                     return;
                 }
                 
-                if (LootRunManager.getActivePath() != null) {
+                if (LootRunManager.getActivePath() != null && LootRunManager.getActivePathName().equals(name) {
                     LootRunManager.clear();
                 }
 
--- a/src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
+++ b/src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
@@ -129,7 +129,12 @@ public class CommandLootRun extends CommandBase implements IClientCommand {
                 if (result) {
                     message = GREEN + "Deleted run " + name + " successfully!";
                 } else {
-                    message = RED + "The provided lootrun doesn't exists!";
+                    sender.sendMessage(new TextComponentString(RED + "The provided lootrun doesn't exists!"));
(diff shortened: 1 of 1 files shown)


DevChromium:
Shouldn't there also be a check if the lootrun you want to delete is the active one?
If I'm doing a lootrun and I decide to delete a random one in my list, this piece of code would unload my current lootrun.

src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
```diff
@@ -129,7 +129,12 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
                 if (result) {
                     message = GREEN + "Deleted run " + name + " successfully!";
                 } else {
-                    message = RED + "The provided lootrun doesn't exists!";
+                    sender.sendMessage(new TextComponentString(RED + "The provided lootrun doesn't exists!"));
```
kristofbolyai: No need to change this line, we print message later on without the return

src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
```diff
@@ -129,7 +129,12 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                    return;
```
kristofbolyai: This line breaks code. Remove it

src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
```diff
@@ -129,7 +129,12 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
...
+                }
+                
+                if (LootRunManager.getActivePath() != null) {
```
kristofbolyai: Check if activepath is the deleted one, only clear if so.

Basically just put this into the if above and it should work.

kristofbolyai (changes requested):

src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
PhoenixOrigin: It was intended to stop the further code from running in the case it doesn't exist and in turn clearing the active lootrun. So what does it break apart from what I made it do?

src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
PhoenixOrigin: Ok 👍

src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
PhoenixOrigin: As shown in the message below

src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
kristofbolyai: Yes I know what return does but if you move the active path if check (the one below) to the correct place this return is not needed and the line above does not need to be changed

PhoenixOrigin: Ok

src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
```diff
@@ -129,7 +129,12 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
                 if (result) {
                     message = GREEN + "Deleted run " + name + " successfully!";
                 } else {
-                    message = RED + "The provided lootrun doesn't exists!";
+                    sender.sendMessage(new TextComponentString(RED + "The provided lootrun doesn't exists!"));
+                    return;
+                }
+                
+                if (LootRunManager.getActivePath() != null && LootRunManager.getActivePathName().equals(name) {
```
kristofbolyai: If missing a )?

kristofbolyai (changes requested):

src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
```diff
@@ -128,6 +128,9 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
                 String message;
                 if (result) {
                     message = GREEN + "Deleted run " + name + " successfully!";
+                    if (LootRunManager.getActivePath() != null && LootRunManager.getActivePathName().equals(name)) {
+                        LootRunManager.clear();
+                    }
                 } else {
                     message = RED + "The provided lootrun doesn't exists!";
```
ryanzhoudev:
```suggestion
                    message = RED + "The provided lootrun doesn't exist!";
```

src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
PhoenixOrigin: done

ryanzhoudev (approved):

Incompleteusern (approved):

kristofbolyai (approved):

PR #508, opened 2022-05-09
feat: Mana timer bar

ryanzhoudev:
Adds a bar similar to health/mana that displays time until MR cycles.
Can be toggled on/off like any standard overlay, and can also switch between seconds/percentage.
Calculates based on soul point ticks.

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/overlays/hud/ManaTimerOverlay.java
+++ b/src/main/java/com/wynntils/modules/utilities/overlays/hud/ManaTimerOverlay.java
@@ -30,6 +30,7 @@ public class ManaTimerOverlay extends Overlay {
     public CustomColor textColor = CommonColors.LIGHT_BLUE;
 
     private static float manaTime = 0.0f;
+    public static boolean isTimeFrozen = false;
 
     @Override
     public void tick(TickEvent.ClientTickEvent event, long ticks) {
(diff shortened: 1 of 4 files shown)


ryanzhoudev:
seconds
percent

src/main/java/com/wynntils/modules/utilities/overlays/hud/ManaTimerOverlay.java
```diff
@@ -0,0 +1,102 @@
...
+    public Pair<Integer, Integer> textPositionOffset = new Pair<>(40, -10);
+
+    @Setting(displayName = "Text Name", description = "The color of the text")
+    public CustomColor textColor = CommonColors.LIGHT_BLUE;
+
+    private static float manaTime = 0.0f;
+
+    @Override
+    public void tick(TickEvent.ClientTickEvent event, long ticks) {
+        if (!(visible = PlayerInfo.get(InventoryData.class).getTicksToNextSoulPoint() != -1 && !Reference.onLobby)) return;
```
kristofbolyai: Simplify this line

src/main/java/com/wynntils/modules/utilities/overlays/hud/ManaTimerOverlay.java
```diff
@@ -0,0 +1,102 @@
...
+                break;
+            case c: drawDefaultBar(-1, 7, 52, 67, textColor);
+                break;
+            case d: drawDefaultBar(-1, 7, 68, 83, textColor);
+                break;
+            case Brune:
+                drawDefaultBar(-1, 8, 83, 100, textColor);
+                break;
+            case Inverse:
+                drawDefaultBar(-1, 7, 100, 115, CommonColors.MAGENTA);
```
kristofbolyai, line 73: Why magenta color here?

kristofbolyai (approved):
Looks nice, good job. Simplify that if statement though.

src/main/java/com/wynntils/modules/utilities/overlays/hud/ManaTimerOverlay.java
ryanzhoudev: I dunno, the normal mana bar has magenta there and the normal health bar has green there
Since this is visually the same as the normal mana one I just left it magenta

kristofbolyai: Got it.

magicus:
I really like the idea! But... I can't make heads or tails from the animated GIFs on how this is supposed to work. What happen when they reach 100%? The mana bar adds up a few mana points. But it also adds upp mana points even before the mana timer reaches 100%. And the mana bar does not go to full mana when the mana timer goes to 100%.

Maybe I'm just lacking knowledge in how Wynncraft calculates and restores mana points. (But I might be a typical sample for a non-expert player then...) So for me, the timers as shown above would not be in any way helpful.

I'd like to see:

1) When am I going to get the next mana point back? 

2) When is the mana bar going to be completely filled up again?

That'd be helpful to me. If this is what the timer is actually saying, then I think it needs a redesign to show this better, or some kind of description on how to interpret it.

ryanzhoudev:
> What happens when they reach 100%?

The mana timer reaching 100% is when your mana regen stat from your gear procs. Eg. if you had total 8/5 MR on your armour/weapons, you'd get 8 mana on that 100% fill.

> But it also adds up mana points even before timer reaches 100%

The points continuously being added are constant and not affected by any gear stats. The timer does not take these into account because it is not significant enough to warrant any gameplay action (see below).

> And the mana bar does not go to full mana when the mana timer goes to 100%

No, the mana timer is only for one cycle of your mana regen stat.

> 1. When am I going to get the next mana point back?

This would be unrealistic and not really useful (imo). The point of this feature is for builds that use high -MR weapons like oblivion or ones that require switching weapons to proc MR. Since normal mana is a constant 1/second, showing this would not be useful since it would be the same between different weapons.

> 2. When is the mana bar going to be completely filled up again?

I believe this would be difficult to implement and a mediocre implementation at best. There is no good way to predict how much mana the player will actually use other than, for example, an average of the player's mana usage in the last few seconds. I also highly doubt anyone is actively trying to achieve full mana during combat, because they will probably be spell cycling.

>So for me, the timers as shown above would not be in any way helpful

I agree the current timer isn't useful for non-endgame/expert players, but this wasn't really implemented with that player group in mind. This feature was originally intended to make predicting -MR procs for things like Lunar Spine, Warp, and Oblivion to allow the player to swap/vanish to cancel the -MR easier. I am not sure how a non-expert/endgame player would benefit from a mana timer; Personally I don't see any value in the timer unless your weapon has a significant mana regen stat, which is rare for lower level weapons.


Please let me know if you need more clarification; I can make a video of the proposed use case if needed

magicus:
@DonkeyBlaster Thank you for your explanation. I understand that this goes way above my pay grade... er, I mean level, build and skills. :-) 

I'll get some inspiration from this though, and think about if I can make a noob version of this, that would be beneficial to players at my knowledge and skill level. I'm constantly confused by my mana levels and never understand how much I drain them or when they will be restored. I think I need to do some reading on the theory to be able to think of a way to communicate it better. But that really has nothing to do with this PR.

kristofbolyai (approved):

HighCrit (approved):

PR #511, opened 2022-05-17
feat: Add better ingredient pouch sorting

ryanzhoudev:
Allows sorting by quantity and rarity (both descending), as well as alphabetically. Tested with normal ingredients, ~~will test with Ironman in a few minutes.~~ also works fine with ironman.

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
+++ b/src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
@@ -1214,18 +1214,20 @@ public class ClientEvents implements Listener {
             case Quantity:
                 sortedIngredients = new LinkedHashMap<>();
                 List<Map.Entry<String, Pair<Integer, Integer>>> sortedQtyList = new LinkedList<>(items.entrySet());
-                sortedQtyList.sort(Comparator.comparing(o -> (o.getValue().a))); // sort based on the pair's first value (qty)
-                for (int i = sortedQtyList.size() - 1; i >= 0; i--) { // reversed for most -> least
-                    sortedIngredients.put(sortedQtyList.get(i).getKey(), sortedQtyList.get(i).getValue());
+                // sort based on the pair's first value (qty), reversed for descending
+                sortedQtyList.sort(Comparator.comparing((Map.Entry<String, Pair<Integer, Integer>> o) -> (o.getValue().a)).reversed());
+                for (Map.Entry<String, Pair<Integer, Integer>> ingredient : sortedQtyList) {
+                    sortedIngredients.put(ingredient.getKey(), ingredient.getValue());
                 }
                 break;
             case Rarity:
             default:
                 sortedIngredients = new LinkedHashMap<>();
                 List<Map.Entry<String, Pair<Integer, Integer>>> sortedRarityList = new LinkedList<>(items.entrySet());
-                sortedRarityList.sort(Comparator.comparing(o -> (o.getValue().b))); // sort based on rarity
-                for (int i = sortedRarityList.size() - 1; i >= 0; i--) { // reversed for rare -> common
-                    sortedIngredients.put(sortedRarityList.get(i).getKey(), sortedRarityList.get(i).getValue());
+                // sort based on rarity, reversed for descending (3* -> 0*)
+                sortedRarityList.sort(Comparator.comparing((Map.Entry<String, Pair<Integer, Integer>> o) -> (o.getValue().b)).reversed());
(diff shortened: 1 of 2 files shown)


src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
```diff
@@ -1187,18 +1184,52 @@ else if (!matcher.matches())
...
             foundFirstItem = true;
 
+            int rarity = 0;
+            if (line.contains("§6")) {
+                rarity = 1;
+            } else if (line.contains("§5")) {
+                rarity = 2;
+            } else if (line.contains("§3")) {
+                rarity = 3;
+            }
```
kristofbolyai: is this good enough for a check?

ryanzhoudev: I feel like it is, that colour code is for the first ['s colour here:
image
i.e. the dark aqua, dark purple, and orange

We could also check for the star color codes if necessary but I don't think any ingredients would reuse these colours elsewhere?

kristofbolyai: If we check for color code + stars it is more obvious

magicus: I agree with kristofbolyai. Checking for stars is clearer, and feels more robust for potential changes from changes in Wynncraft code. The stars are the significant information, the color is incidental.

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
```diff
@@ -1187,18 +1184,52 @@ else if (!matcher.matches())
...
+        Map<String, Pair<Integer, Integer>> sortedIngredients;
+        switch (UtilitiesConfig.INSTANCE.sortIngredientPouch) {
+            case Alphabetical:
+                sortedIngredients = new TreeMap<>(items); // TreeMap will sort it for us
+                break;
+            case Quantity:
+                sortedIngredients = new LinkedHashMap<>();
+                List<Map.Entry<String, Pair<Integer, Integer>>> sortedQtyList = new LinkedList<>(items.entrySet());
+                sortedQtyList.sort(Comparator.comparing(o -> (o.getValue().a))); // sort based on the pair's first value (qty)
+                for (int i = sortedQtyList.size() - 1; i >= 0; i--) { // reversed for most -> least
```
magicus: Why are you reversing the sort instead of inverting the comparator?

ryanzhoudev: I feel like this is easier to read, no? Most sorts will be in this direction and inverting it isn't very computationally expensive, it's only O(n) for reversing a list. LMK if this isn't preferred though

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
ryanzhoudev: Yeah we can check for stars - I believe the only way to do this is to split that section of the string since they use color codes to denote how many stars are actually present.

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
magicus: You just need to chain on `reversed()` on the `Comparator`. I think that's not only more efficient (which doesn't matter here), but definitely cleaner and more obvious.

ryanzhoudev: Ah okay thanks

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
```diff
@@ -1186,19 +1183,56 @@ else if (!matcher.matches())
...
         }
 
+        Map<String, Pair<Integer, Integer>> sortedIngredients;
+        switch (UtilitiesConfig.INSTANCE.sortIngredientPouch) {
+            case Alphabetical:
+                sortedIngredients = new TreeMap<>(items); // TreeMap will sort it for us
+                break;
+            case Quantity:
+                sortedIngredients = new LinkedHashMap<>();
+                List<Map.Entry<String, Pair<Integer, Integer>>> sortedQtyList = new LinkedList<>(items.entrySet());
```
magicus: Also, LinkedList sucks in performance. As a rule of thumb, always use ArrayList instead.

magicus (approved):

kristofbolyai (approved):

PR #512, opened 2022-05-20
Misc patches.

ByThePowerOfScience:
Fixes crash on https://wynncraft.fandom.com/wiki/Crop_Failure when harvesting "rotten oats".


ByThePowerOfScience:
I'm fixing these things via Mixin for a Wynncraft modpack, but I figure I should submit a PR first so they can be fixed internally.

src/main/java/com/wynntils/core/framework/enums/professions/GatheringMaterial.java
```diff
@@ -13,7 +13,7 @@ public enum GatheringMaterial {
     COPPER, GRANITE, GOLD, SANDSTONE, IRON, SILVER, COBALT, KANDERSTONE, DIAMOND, MOLTEN, VOIDSTONE,
 
     // Farming
-    WHEAT, BARLEY, OATS, MALT, HOPS, RYE, MILLET, DECAY_ROOTS, RICE, SORGHUM, HEMP,
+    WHEAT, BARLEY, OATS, MALT, HOPS, RYE, MILLET, DECAY_ROOTS, RICE, SORGHUM, HEMP, ROTTEN,
```
ryanzhoudev: I feel like this could be in its own section at the bottom of the enum with a comment on why it exists; ROTTEN itself isn't super descriptive

ByThePowerOfScience: Fair. Maybe a "quest-only" category or something.

ByThePowerOfScience: Edited.

ryanzhoudev (approved):

PR #515, opened 2022-05-21
FIX: Mounting horse when sprinting uses stamina

LapinoLapidus:
Currently, whenever you sprint and mount your horse, your stamina bar will keep running out until you remount it while not sprinting. This one-liner fixes that by stopping the sprinting of the player. I've tried restoring the sprint status after mounting but then the stamina bar runs out again, maybe there needs to be a little wait time? Not sure how you'd like me to implement that though as that may be confusing for the end user, so I'll just leave it like this for now

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/managers/MountHorseManager.java
+++ b/src/main/java/com/wynntils/modules/utilities/managers/MountHorseManager.java
@@ -139,6 +139,7 @@ public class MountHorseManager {
         }
 
         player.inventory.currentItem = 8; // swap to soul points to avoid any right-click conflicts
+        McIf.player().setSprinting(false);
         playerController.interactWithEntity(player, playersHorse, EnumHand.MAIN_HAND);
         player.inventory.currentItem = prev;
         return MountHorseStatus.SUCCESS;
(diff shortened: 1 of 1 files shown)


McPlayHD:
What happens if you sprint, mount your horse, and then dismount again all while keep sprinting? Will the server then not think that you are not sprinting but since you are sprinting locally it will lag you back or something?

src/main/java/com/wynntils/modules/utilities/managers/MountHorseManager.java
```diff
@@ -122,7 +122,8 @@ public static MountHorseStatus mountHorse(boolean allowRetry) {
                 return MountHorseStatus.NO_HORSE;
             }
 
-            player.inventory.currentItem = horse.getInventorySlot();
+            
+            .inventory.currentItem = horse.getInventorySlot();
```
kristofbolyai, line 126: I assume this is not intentional?

LapinoLapidus: Yup it's something weird that happened while I was editing it on the github desktop editor, never doing that again. I'll revert it in a bit

PR #518, opened 2022-05-22
fix: Quest completed dialogue chat fixes

ryanzhoudev:
Fixes the quest completed dialogue looking like this: https://cdn.discordapp.com/attachments/709413517907722300/977665607896031304/unknown.png

Works by doubling the number of new lines processed when that quest completed bit comes up. Also adds ignores to empty and \n lines. Tested on the starting quest and the lava springs quest.


kristofbolyai (approved):

magicus (approved):
I think it looks good.

ByThePowerOfScience:
Awesome! Apparently the uncaught exception that used to be in that block stopped this from being an issue, meaning that my fix for that exception brought the bug back to light. Thanks for fixing this problem!

PR #520, opened 2022-05-23
feat: Mana timer enhancements

ryanzhoudev:
Hides by default, moves positioning more to the right to not be on top of the normal mana bar. Solves most of the complaints about it in the general channel

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/configs/OverlayConfig.java
+++ b/src/main/java/com/wynntils/modules/utilities/configs/OverlayConfig.java
@@ -139,6 +139,9 @@ public class OverlayConfig extends SettingsClass {
     public static class ManaTimer extends SettingsClass {
         public static ManaTimer INSTANCE;
 
+        @Setting(displayName = "Force Hide Mana Timer", description = "Enabling this will hide the mana timer regardless of overlay config status.")
+        public boolean hideManaTimer = false;
+
         @Setting(displayName = "Mana Timer Percentage", description = "Should the mana timer use percentages instead of seconds?")
         public boolean manaTimerUsePercentage = false;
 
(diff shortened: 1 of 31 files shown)


P0keDev (changes requested):
not a fan of the "disable by default" implementation, and having to change the whole overlay framework for it. would be cleaner and easier for the user if it was a normal setting in the config next to the other mana bar settings, imo

ryanzhoudev:
Yeah adding a setting to the config book works. Though this makes all the other overlay settings inconsistent - did you want settings for all of them?

P0keDev:
no, i think it's fine like this - this is much more of an "additional feature" than any other overlay, and it won't be the only overlay to have config book settings (action bar, off the top of my head)

src/main/java/com/wynntils/modules/utilities/configs/OverlayConfig.java
```diff
@@ -139,6 +139,9 @@ public enum ManaTextures {
     public static class ManaTimer extends SettingsClass {
         public static ManaTimer INSTANCE;
 
+        @Setting(displayName = "Force Hide Mana Timer", description = "Enabling this will hide the mana timer regardless of overlay config status.")
```
P0keDev: why isn't this just a normal toggle instead of being inverted (enabled = no timer)? and it wouldn't be disabled by default now, which was important

ryanzhoudev: I assumed the default hide isn't important anymore now that there are multiple ways to disable it

P0keDev: i'd expect the majority of users to not really benefit from this feature, so it's best to have it disabled by default regardless

P0keDev (approved):

kristofbolyai (approved):

PR #524, opened 2022-05-26
Move SSL registration to top of pre-init, on windows, use windows-root cert store

kristofbolyai:


HeyZeer0:
Can you unnest the FixSSL class and name it more properly? Something like `SSLCertificateManager` would work I guess.
Also instead of calling the class twice just create a method that calls both things or simply merge the method together to make it more clear!

magicus (approved):

HeyZeer0 (approved):

PR #528, opened 2022-06-05
PlayerClassProfile: Remove fields from deserializer that is not in the Wynn API anymore

kristofbolyai:


ryanzhoudev (approved):

magicus:
Do we think it's coming back, or should we search and remove all usages of chestsFound? (Or is it not worth it to keep Legacy clean?)

kristofbolyai:
> Do we think it's coming back, or should we search and remove all usages of chestsFound? (Or is it not worth it to keep Legacy clean?)

I think most of this class is not used, it's just deserialized along with the useful stuff. (iirc I checked and chestcount was not used anywhere)

magicus:
Then maybe just remove it instead of commenting it out..?

kristofbolyai:
> Then maybe just remove it instead of commenting it out..?

Every other unused field is left there, not sure what the original intention was, but I don't think it's worth the time messing with it.

magicus:
Ah, ok then.

magicus (approved):

PR #536, opened 2022-06-20
Do not lock slots in ability tree page

ryanzhoudev:


src/main/java/com/wynntils/core/utils/Utils.java
```diff
@@ -147,6 +147,12 @@ public static boolean isCharacterInfoPage(GuiScreen gui) {
 
     }
 
+    public static boolean isAbilityTreePage(GuiScreen gui) {
+        if (!(gui instanceof GuiContainer)) return false;
+        Matcher m = ABILITY_TREE_PATTERN.matcher(((GuiContainer) gui).inventorySlots.getSlot(0).inventory.getName());
```
ryanzhoudev, line 152: Not super fond of the regex here and was looking into just doing a .startsWith and a .endsWith. Supposedly regexes are significantly slower than those two

kristofbolyai (approved):

PR #547, opened 2022-07-15
Added horse overlays

playfuldoggo:
Created an overlay that shows the current xp, level/maxlevel and an approximate time in minutes of riding until the next level  (based off of [this](https://wynncraft.fandom.com/wiki/Horses)) if any horse is in your hotbar (utilizes 
the existing 'com/wynntils/core/framework/instances/data/HorseData.java'). By default it is on, it can be toggled on or off in the configuration menu.

Very new to all this, please let me know if there is anything that needs changing, anything you would do differently, or anything I would messed up. I decided not to add comments as I didn't think I could add anything meaningful in comment, if you would still like some comments then I will go ahead and add them.


ryanzhoudev:
How does this overlay improve upon the already existing info variables? Could the new timer be added as an info variables instead?

Also, the approximate time function is inaccurate at some points. 43 seconds passed for the timer to go down by 1 minute on a 23/40 white horse. 1m17s passed for a 1.4min timer on a 2/10 brown horse. 1m49s passed for a 1.8min timer on a 3/10 brown horse, which is pretty good. All these tests were done riding nonstop in a (somewhat) straight line between Detlas/Ragni through Nivla.

playfuldoggo:
I guess it doesn't really improve on anything. I've now moved the approximate time function to an info variable and deleted the standalone overlay and related lines. Also sorry about all the unnecessary commits.

src/main/java/com/wynntils/modules/utilities/instances/InfoFormatter.java
```diff
@@ -202,6 +203,20 @@ public InfoFormatter() {
             return cache.get("horsetier");
         }, "horse_tier", "h_tier");
 
+        // Approximate time (in minutes) until next horse level
+        registerFormatter((input) -> {
+            double mlvl = 3.0 * PlayerInfo.get(HorseData.class).getLevel();
+            double plvl = mlvl + 2;
```
ryanzhoudev: It would be nice to have a comment explaining this process or more descriptive variables. Here it is just a mess of letters without much meaning.

src/main/java/com/wynntils/modules/utilities/UtilitiesModule.java
```diff
@@ -65,6 +65,7 @@ public void onEnable() {
         registerOverlay(new ManaTimerOverlay(), Priority.LOW);
         registerOverlay(new ExpBarOverlay(), Priority.NORMAL);
         registerOverlay(new LevelingOverlay(), Priority.LOW);
+        //registerOverlay(new HorseOverlay(), Priority.LOW);
```
ryanzhoudev: Remove unnecessary comment please.

ryanzhoudev:
> I guess it doesn't really improve on anything. I've now moved the approximate time function to an info variable and deleted the standalone overlay and related lines. Also sorry about all the unnecessary commits.

Don't worry about the commits, they're squashed when merged anyway

playfuldoggo:
I made the calculations a lot more compact. I am unsure on what I would put in a comment explaining it, as it is just following a formula.

ryanzhoudev:
> I made the calculations a lot more compact. I am unsure on what I would put in a comment explaining it, as it is just following a formula.

I was thinking the comment would include something like how the formula works, or where the formula is from

playfuldoggo:
Hopefully this is alright, I actually made the formula for the wiki, it's based off what I assume is [someone's research](https://forums.wynncraft.com/threads/calculate-how-much-time-until-a-horse-reaches-a-target-level.278754/) from December 2020, however they have been long inactive and have never discussed how it was made.

ryanzhoudev:
> Hopefully this is alright, I actually made the formula for the wiki, it's based off what I assume is [someone's research](https://forums.wynncraft.com/threads/calculate-how-much-time-until-a-horse-reaches-a-target-level.278754/) from December 2020, however they have been long inactive and have never discussed how it was made.

Looks good. Last thing that I believe should be considered is that the timer still displays a time on a maxed horse. (ie. 40/40, 30/30]. This should be checked for and a different value should be returned. Maybe `-`, `0`, or even an empty string?

playfuldoggo:
Should be fixed, "-" will be returned in case the horse is not in the hot bar/inventory or if it is already at max level.

ryanzhoudev (approved):

magicus:
The addition of a new info variable as such looks good to me. However, I'd like to get some more clarity regarding the formula.

If I interpret the formula on https://www.desmos.com/calculator/mvedit4swq correctly, it is slightly different than the one on the Wiki. I'm not 100% sure I understand what the different parts mean, but they seem to be using "3o/6 + 5/6" instead or "(3o+2)/6".

Also, it sounds like @DonkeyBlaster is getting results that indicate that the formula is incorrect. If I understand him correctly, the timer said it should take 60 seconds to increase one level, but it only took 43 seconds.

And finally, there is a dissenting comment on the wiki, questioning the formula: "Note: (Someone suspected that a horse takes slightly shorter to level up riding non-stop. For example, a Chestnut horse going from level 22 to 23 takes roughly 9 minutes, whereas this article implies it would take about 11m20s. Please recheck.) "

This comment seems in line with Donkey's result, in that the formula seem to give larger values than what is actually experienced.

I think it is a good thing to add a timer such as this, but it needs to be (at least roughly) correct. and I'm not convinced the formula from the wiki is correct. Can you see if you can understand how to extract the spreadsheet formula, and see if that gives results that are more in line with reality?

magicus:
Also note that the wiki was very recently edited, 15 July 2022, a change to precisely this formula. The old version of the wiki had this formula:

https://wynncraft.fandom.com/wiki/Horses?oldid=149009

The change had this comment: "Changed formula for approximating time using the original source, old formula would go negative when using high 90s xp values. Tested using current formula and it does not seem to be too far off (5 - 10 seconds off going from level 8 to level 9"

ryanzhoudev:
I am not sure the formula estimating slightly longer than the actual time is such a huge issue, especially considering all my tests were done in nonstop straight lines. On one hand, most of the playerbase does not spend a lot of time riding horses in straight lines. On the other hand, people that level horses nonstop to sell may find this timer particularly useful. Either way, since the timer updates in real-time, I don't think errors up to a minute are really worth nitpicking over. The 9 minutes vs 11m20s may be an issue though; I'll try and buy a horse in that level range sometime soon and check it out myself.

magicus:
It might be that an incorrect timer is better than no timer, yes. (It's just me who am so finicky about getting formulas 100% correct. I think I spent a day trying to get the mana-cost-per-skill-point formula right down to the fraction :-)).

magicus:
@HeyZeer0 Maybe you know about horse leveling? I get the impression people have tried reverse-engineering the formula used by Wynncraft. But if it is not a secret, maybe we can get the proper formula directly from the source?

playfuldoggo:
> Also note that the wiki was very recently edited, 15 July 2022, a change to precisely this formula. The old version of the wiki had this formula:
> 
> https://wynncraft.fandom.com/wiki/Horses?oldid=149009
> 
> The change had this comment: "Changed formula for approximating time using the original source, old formula would go negative when using high 90s xp values. Tested using current formula and it does not seem to be too far off (5 - 10 seconds off going from level 8 to level 9"

I edited the wiki. As I was creating the variable I had realised it had gone negative when my horse reached high xp values, I recalculated it using a real calculator to check whether it was the formula or me that had an error, the formula itself was wrong and using the [source](https://www.desmos.com/calculator/rrckinnsjo) for the old formula, made a new one that matched the source exactly instead of whoever created the formula that was put on the wiki. The source is from a [forum post](https://forums.wynncraft.com/threads/calculate-how-much-time-until-a-horse-reaches-a-target-level.278754/) from nearly 2 years ago, they have since been inactive on the forum.

https://user-images.githubusercontent.com/83927732/181294178-34e7cc35-48d5-4df4-8846-828ef904568a.mp4

Here is a test I just performed, it was only a second off, which was likely the rounding that is performed on the number that gets put into the variable or my slow stopping of the timer. Hopefully this answers some questions.

magicus:
Ah, I see. I should have noticed the user name on the wiki. :)

Then I think it is all good. Thanks for your contribution to Wynntils!

magicus (approved):

PR #549, opened 2022-07-17
Beta fixes 6

ryanzhoudev:
- Fixes archer speed timer
- Fixes ingredient filter button position in character info page
- Fixes custom character selector appearing incorrectly
- Add specification nums for remaining SP
- Add specification nums for remaining AP (works in both ability tree page and char info page)
- Add specification nums for # of archetype upgrades selected
- Remove mentions of spell numbers (eg. 1st spell, 2nd spell). Spell numbers are no longer used in-game. (They are still used internally)
- Fix assassin vanish timer(s)
- Fix middle click to open items on wynnbuilder
- Fix skill point overlay not showing up

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/managers/KeyManager.java
+++ b/src/main/java/com/wynntils/modules/utilities/managers/KeyManager.java
@@ -64,10 +64,10 @@ public class KeyManager {
 
         zoomOutKey = CoreModule.getModule().registerKeyBinding("Zoom Out", Keyboard.KEY_MINUS, "Wynntils", KeyConflictContext.IN_GAME, false, () -> MiniMapOverlay.zoomBy(-1));
 
-        CoreModule.getModule().registerKeyBinding("Cast 1st Spell", Keyboard.KEY_Z, "Wynntils", KeyConflictContext.IN_GAME, true, QuickCastManager::castFirstSpell);
-        CoreModule.getModule().registerKeyBinding("Cast 2nd Spell", Keyboard.KEY_X, "Wynntils", KeyConflictContext.IN_GAME, true, QuickCastManager::castSecondSpell);
-        CoreModule.getModule().registerKeyBinding("Cast 3rd Spell", Keyboard.KEY_C, "Wynntils", KeyConflictContext.IN_GAME, true, QuickCastManager::castThirdSpell);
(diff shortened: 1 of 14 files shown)


ryanzhoudev:
A few more things left to do in ticket 48, though I feel like this is enough for one commit/update

And yeah, commits are kind of messy
If you don't feel like looking through them we could just gamble and merge this, then deal with all the beta fixes at the end once merged to development

src/main/java/com/wynntils/modules/utilities/managers/KeyManager.java
```diff
@@ -64,10 +64,10 @@ public static void registerKeys() {
 
         zoomOutKey = CoreModule.getModule().registerKeyBinding("Zoom Out", Keyboard.KEY_MINUS, "Wynntils", KeyConflictContext.IN_GAME, false, () -> MiniMapOverlay.zoomBy(-1));
 
-        CoreModule.getModule().registerKeyBinding("Cast 1st Spell", Keyboard.KEY_Z, "Wynntils", KeyConflictContext.IN_GAME, true, QuickCastManager::castFirstSpell);
-        CoreModule.getModule().registerKeyBinding("Cast 2nd Spell", Keyboard.KEY_X, "Wynntils", KeyConflictContext.IN_GAME, true, QuickCastManager::castSecondSpell);
-        CoreModule.getModule().registerKeyBinding("Cast 3rd Spell", Keyboard.KEY_C, "Wynntils", KeyConflictContext.IN_GAME, true, QuickCastManager::castThirdSpell);
-        CoreModule.getModule().registerKeyBinding("Cast 4th Spell", Keyboard.KEY_V, "Wynntils", KeyConflictContext.IN_GAME, true, QuickCastManager::castFourthSpell);
+        CoreModule.getModule().registerKeyBinding("Cast R-L-R Spell", Keyboard.KEY_Z, "Wynntils", KeyConflictContext.IN_GAME, true, QuickCastManager::castFirstSpell);
```
magicus, line 67: Only for Archer it's L-R-L, etc..?

ryanzhoudev: Yeah but you can't have dynamic keybind text, and with the removal of user facing spell numbers, I feel like this compromise is the best option

If you thought of a better label, we can include it in the next set of beta fixes

Issue #545, opened 2022-07-07
ChatOverlay causes crash on multiple occasions.


RainVaporeon:
### Version

1.11.4_663

### Brief Description

Crash occurs upon several cases:

- On Item pick up (For instance, grinding squids)
- On switching servers (For instance, /switch)

Might be an incompatibility with other mods that sends messages but it's annoying regardless

<details>
 <summary> Stacktrace </summary>
[09:42:04] [Netty Client IO #1/ERROR] [FML]: NetworkDispatcher exception 
java.lang.NullPointerException: null
	at com.wynntils.modules.chat.overlays.ChatOverlay.lambda$null$3(ChatOverlay.java:481) ~[ChatOverlay.class:?]
	at com.wynntils.modules.chat.overlays.ChatOverlay$$Lambda$580/1771157745.test(Unknown Source) ~[?:?]
	at java.util.ArrayList.removeIf(ArrayList.java:1405) ~[?:1.8.0_51]
	at com.wynntils.modules.chat.overlays.ChatOverlay.lambda$deleteChatLine$4(ChatOverlay.java:481) ~[ChatOverlay.class:?]
	at com.wynntils.modules.chat.overlays.ChatOverlay$$Lambda$579/814043951.accept(Unknown Source) ~[?:?]
	at java.util.ArrayList.forEach(ArrayList.java:1249) ~[?:1.8.0_51]
	at com.wynntils.modules.chat.overlays.ChatOverlay.deleteChatLine(ChatOverlay.java:479) ~[ChatOverlay.class:?]
	at com.wynntils.modules.chat.managers.ChatManager.onLeave(ChatManager.java:1026) ~[ChatManager.class:?]
	at com.wynntils.modules.chat.events.ClientEvents.onLeaveWorld(ClientEvents.java:154) ~[ClientEvents.class:?]
	at net.minecraftforge.fml.common.eventhandler.ASMEventHandler_105_ClientEvents_onLeaveWorld_Leave.invoke(.dynamic) ~[?:?]
	at net.minecraftforge.fml.common.eventhandler.ASMEventHandler.invoke(ASMEventHandler.java:90) ~[ASMEventHandler.class:?]
	at net.minecraftforge.fml.common.eventhandler.EventBus.post(EventBus.java:182) ~[EventBus.class:?]
	at com.wynntils.core.events.ClientEvents.onTabListChange(ClientEvents.java:200) ~[ClientEvents.class:?]
[...]

PR #551, opened 2022-07-23
Fix NPE crash when chatLine is null

RainVaporeon:
Fix issue #545 after testing for days, implemented a null check in case of NullPointerExceptions

### DIFF
--- a/src/main/java/com/wynntils/modules/chat/overlays/ChatOverlay.java
+++ b/src/main/java/com/wynntils/modules/chat/overlays/ChatOverlay.java
@@ -478,6 +478,7 @@ public class ChatOverlay extends GuiNewChat {
 
         TabManager.getAvailableTabs().forEach(tab -> {
             if (tab == currentTab) return;
+            tab.getCurrentMessages().removeIf(chatline -> chatline == null);
             tab.getCurrentMessages().removeIf(chatline -> chatline.getChatLineID() == id);
         });
 


kristofbolyai:
I dont think this fixes the root cause, chatline get's to be null in some other place.

ryanzhoudev:
I have never actually experienced issue 545. You claim it is an incompatibility with other mods that send messages. What mods are these? Can you rewrite steps to reproduce with more detail so we can determine if this is an issue that needs to be fixed, and if this fix actually works? As far as I know, "[picking up] a large quantity of items constantly" doesn't actually send any messages in chat, so I'm not sure how a chat crash is related to that.

RainVaporeon:
> I have never actually experienced issue 545. You claim it is an incompatibility with other mods that send messages. What mods are these? Can you rewrite steps to reproduce with more detail so we can determine if this is an issue that needs to be fixed, and if this fix actually works? As far as I know, "[picking up] a large quantity of items constantly" doesn't actually send any messages in chat, so I'm not sure how a chat crash is related to that.

Not sure, even without the mods I still crash while grinding mobs (specifically when I was doing squids), another person also told me that they were getting kicked/crashing while just doing /switch to servers, and this patched it apparently

(Not sure if this matters but when picking up ings there are pouch messages? Might be related i dunno)

ryanzhoudev:
The pouch messages when picking up ings are titles/subtitles and those are redirected to the update ticker, so chat shouldn't be involved at all. I have experienced the kick/crash when /switching, but that was only this morning when dxp was announced, so I thought it was just due to server load. I'll pay attention to if it happens during normal server loads

kristofbolyai (approved):
I guess this is okay as a temporary fix to patch the crashes.

ryanzhoudev (approved):

RainVaporeon:
Yeah, when I got time I'll look deeper into this, for now this is just a
bandage. Feel free to reject though

2022年7月23日(土) 21:11 Kristof Kovacs ***@***.***>:

PR #552, opened 2022-07-27
Changed the character the down arrow uses in the 'Waypoint overview menu' from 'V' to 'ᵛ'. Added a new config option to change the amount of space between the data in the 'Waypoint overview menu'.

playfuldoggo:
Me and my friends have found the 'Waypoint overview menu' hard to navigate and hard to delete and look at many waypoints. I made this to try and add an option that would help with this issue. There are three options 'DEFAULT' (what it is currently), 'COMPACT' (the same sizes except with no space between the 'Edit...' and 'Delete' buttons) and 'DECREASED_SIZE' (smaller 'Edit...' and 'Delete' buttons, and smaller up arrow and down arrow buttons).

Hopefully this is all alright, please let me know if there is anything that needs changing or comments that need to be added.

I am also not sure whether or not the commits form my previous pull request matter or not as I have not re-cloned the repository since Jul 15.


src/main/java/com/wynntils/modules/map/overlays/ui/WaypointOverviewUI.java
```diff
@@ -51,13 +51,17 @@ public class WaypointOverviewUI extends GuiScreen {
     private int group = ungroupedIndex;
     private int groupWidth;
     private int groupScroll = 0;
+    private int[] spacingArr = {25, 20, 14};
```
magicus: Instead of having a separate array with constants and an index into this array, can't you just set the spacing multiplier in the WaypointSpacing enum constructor directly? And have a `getSpacingMultiplier` method instead of getting the index.

src/main/java/com/wynntils/modules/utilities/configs/OverlayConfig.java
```diff
@@ -703,6 +703,7 @@ public enum Variables {
             HORSE_LEVEL_MAX("Max horse level"),
             HORSE_XP("Current horse xp"),
             HORSE_TIER("Current horse tier"),
+            HORSE_TIME_ESTIMATE("Estimated time (in minutes) from current horse level to next horse level"),
```
magicus: Somehow this PR seems to contain the latest updates to the mainline. I don't know what happened, but try merging in the latest main.

src/main/java/com/wynntils/modules/map/overlays/ui/WaypointOverviewUI.java
```diff
@@ -389,10 +393,10 @@ private void setEditButtons() {
...
-            GuiButton down = new GuiButton(7 + 10 * i, this.width/2 + 172 + groupShift, 65 + 25 * i, 9, 9, "v");
+            editButtons.add(new GuiButton(3 + 10 * i, this.width/2 + 85 + groupShift, 54 + spacingArr[sizeArrayIndex] * i, (int)Math.round(40.0 * (decreasedSize ? 0.6 : 1.0)), (int)Math.round(20.0 * (decreasedSize ? 0.6 : 1.0)),"Edit..."));
+            editButtons.add(new GuiButton(5 + 10 * i, this.width/2 + 130 + groupShift, 54 + spacingArr[sizeArrayIndex] * i, (int)Math.round(40.0 * (decreasedSize ? 0.8 : 1.0)), (int)Math.round(20.0 * (decreasedSize ? 0.6 : 1.0)), "Delete"));
+            GuiButton up = new GuiButton(6 + 10 * i, this.width/2 + 172 + groupShift, 54 + spacingArr[sizeArrayIndex] * i, (int)Math.round(9 * (decreasedSize ? 0.75 : 1.0)), (int)Math.round(9 * (decreasedSize ? 0.75 : 1.0)), "\u028C");
+            GuiButton down = new GuiButton(7 + 10 * i, this.width/2 + 172 + groupShift, 54 + spacingArr[sizeArrayIndex] * i + (int)Math.round(9 * (decreasedSize ? 0.75 : 1.0)), (int)Math.round(9 * (decreasedSize ? 0.75 : 1.0)), (int)Math.round(9 * (decreasedSize ? 0.75 : 1.0)), "\u1D5B");
```
magicus: This was originally based on 65, not 54 as the others. Is this change to 54 intentional?

playfuldoggo: Yes as there is 11 added to 54 for that one as to add space between the two arrow buttons, however the space between the arrow buttons needs to be changed by the settings so I separated it out, the default is now 63 * i instead of 65 * i , this was done on complete accident because I did 65 - 54 wrong, however I think it actually looks better than 65 * i




Smaller gap is the 63 * i, bigger gap is the 65 * i. I can change it so that the gap is the same as it was before on default if you would like.

src/main/java/com/wynntils/modules/map/overlays/ui/WaypointOverviewUI.java
playfuldoggo: I think this is what you wanted (latest commit). If you want me to just use the `getSpacingMultiplier` method everywhere instead of just putting it into an into and using that, I can.

src/main/java/com/wynntils/modules/map/overlays/ui/WaypointOverviewUI.java
```diff
@@ -51,13 +51,17 @@ public class WaypointOverviewUI extends GuiScreen {
     private int group = ungroupedIndex;
     private int groupWidth;
     private int groupScroll = 0;
+    private int spacingMultiplier = MapConfig.Waypoints.INSTANCE.waypointSpacing.getSpacingMultiplier();
+    private boolean decreasedSize = (spacingMultiplier == 14);
+
```
magicus, line 56: Delete extra empty line

src/main/java/com/wynntils/modules/map/overlays/ui/WaypointOverviewUI.java
magicus: Yes, this look good.

magicus (dismissed):
This looks okay for me now.

src/main/java/com/wynntils/modules/map/configs/MapConfig.java
```diff
@@ -229,6 +229,22 @@ public boolean isTierAboveThis(String testTier) {
         @Setting(displayName = "Compass Marker", description = "Should a marker appear on the map where the compass location is set to?")
         public boolean compassMarker = true;
 
+        @Setting(displayName = "Waypoint Menu Spacing", description = "What amount of space should be between each set of values in the waypoint menu")
+        public WaypointSpacing waypointSpacing = WaypointSpacing.DEFAULT;
+
+        public enum WaypointSpacing {
+            DECREASED_SIZE(2),
```
ryanzhoudev: image

The DECREASED_SIZE option makes this menu completely broken - Text is overflowing out of the buttons and the icons do not line up at all.

src/main/java/com/wynntils/modules/map/configs/MapConfig.java
```diff
@@ -229,6 +229,22 @@ public boolean isTierAboveThis(String testTier) {
...
+            COMPACT(1),
```
ryanzhoudev: Same goes for COMPACT - the text is fine, but the icons do not line up at all.

src/main/java/com/wynntils/modules/map/configs/MapConfig.java
```diff
@@ -229,6 +229,22 @@ public boolean isTierAboveThis(String testTier) {
...
+            DEFAULT(0);
+
+            private int[] spacingArr = {25, 20, 14};
```
ryanzhoudev: What is the point of this int array? Could we not just do this?
```
DECREASED_SIZE(14),
COMPACT(20),
DEFAULT(25)
```

ryanzhoudev: Also if it is necessary, make it final please

src/main/java/com/wynntils/modules/map/overlays/ui/WaypointOverviewUI.java
```diff
@@ -389,10 +392,10 @@ private void setEditButtons() {
...
-            GuiButton down = new GuiButton(7 + 10 * i, this.width/2 + 172 + groupShift, 65 + 25 * i, 9, 9, "v");
+            editButtons.add(new GuiButton(3 + 10 * i, this.width/2 + 85 + groupShift, 54 + spacingMultiplier * i, (int)Math.round(40.0 * (decreasedSize ? 0.6 : 1.0)), (int)Math.round(20.0 * (decreasedSize ? 0.6 : 1.0)),"Edit..."));
+            editButtons.add(new GuiButton(5 + 10 * i, this.width/2 + 130 + groupShift, 54 + spacingMultiplier * i, (int)Math.round(40.0 * (decreasedSize ? 0.8 : 1.0)), (int)Math.round(20.0 * (decreasedSize ? 0.6 : 1.0)), "Delete"));
+            GuiButton up = new GuiButton(6 + 10 * i, this.width/2 + 172 + groupShift, 54 + spacingMultiplier * i, (int)Math.round(9 * (decreasedSize ? 0.75 : 1.0)), (int)Math.round(9 * (decreasedSize ? 0.75 : 1.0)), "\u028C");
+            GuiButton down = new GuiButton(7 + 10 * i, this.width/2 + 172 + groupShift, 54 + spacingMultiplier * i + (int)Math.round(9 * (decreasedSize ? 0.75 : 1.0)), (int)Math.round(9 * (decreasedSize ? 0.75 : 1.0)), (int)Math.round(9 * (decreasedSize ? 0.75 : 1.0)), "\u1D5B");
```
ryanzhoudev: image

Since you changed the character into the different `v`, if possible, could you line it up so it's consistent (vertically) with the `^`? It's fine if you can't, this isn't a huge issue imo

ryanzhoudev (changes requested):

src/main/java/com/wynntils/modules/map/configs/MapConfig.java
playfuldoggo: Forgot to write anything about this in the most recent commit message, however the icons and the text should both be fixed in the most recent commit.

src/main/java/com/wynntils/modules/map/overlays/ui/WaypointOverviewUI.java
playfuldoggo: I don't seem to be able to as just trying to use a newline doesn't work with 'GuiButton'. There may be another way to line it up, if you find one then let me know.

src/main/java/com/wynntils/modules/map/configs/MapConfig.java
playfuldoggo: Removed the array and changed the values in the enum constructor to what was in the array.

magicus (dismissed):
The code looks okay to me

src/main/java/com/wynntils/modules/map/overlays/ui/WaypointOverviewUI.java
```diff
@@ -389,10 +392,10 @@ private void setEditButtons() {
...
-            GuiButton down = new GuiButton(7 + 10 * i, this.width/2 + 172 + groupShift, 65 + 25 * i, 9, 9, "v");
+            editButtons.add(new GuiButton(3 + 10 * i, this.width/2 + 85 + groupShift, 54 + spacingMultiplier * i, (int)Math.round(40.0 * (decreasedSize ? 0.7 : 1.0)), (int)Math.round(20.0 * (decreasedSize ? 0.6 : 1.0)),"Edit..."));
+            editButtons.add(new GuiButton(5 + 10 * i, this.width/2 + 130 + groupShift, 54 + spacingMultiplier * i, (int)Math.round(40.0 * (decreasedSize ? 0.9 : 1.0)), (int)Math.round(20.0 * (decreasedSize ? 0.6 : 1.0)), "Delete"));
+            GuiButton up = new GuiButton(6 + 10 * i, this.width/2 + 172 + groupShift, 54 + spacingMultiplier * i, (int)Math.round(9 * (decreasedSize ? 0.75 : 1.0)), (int)Math.round(9 * (decreasedSize ? 0.75 : 1.0)), "\u028C");
+            GuiButton down = new GuiButton(7 + 10 * i, this.width/2 + 172 + groupShift, 54 + spacingMultiplier * i + (int)Math.round(9 * (decreasedSize ? 0.75 : 1.0)), (int)Math.round(9 * (decreasedSize ? 0.75 : 1.0)), (int)Math.round(9 * (decreasedSize ? 0.75 : 1.0)), "\u1D5B");
```
ryanzhoudev:
```suggestion
            GuiButton down = new GuiButton(7 + 10 * i, this.width/2 + 172 + groupShift, 54 + spacingMultiplier * i + (int)Math.round(9 * (decreasedSize ? 0.75 : 1.0)), (int)Math.round(9 * (decreasedSize ? 0.75 : 1.0)), (int)Math.round(9 * (decreasedSize ? 0.75 : 1.0)), "\u1D5B"); // This is the ᵛ character
```

src/main/java/com/wynntils/modules/map/configs/MapConfig.java
```diff
@@ -229,6 +229,21 @@ public boolean isTierAboveThis(String testTier) {
         @Setting(displayName = "Compass Marker", description = "Should a marker appear on the map where the compass location is set to?")
         public boolean compassMarker = true;
 
+        @Setting(displayName = "Waypoint Menu Spacing", description = "What amount of space should be between each set of values in the waypoint menu")
+        public WaypointSpacing waypointSpacing = WaypointSpacing.DEFAULT;
+
+        public enum WaypointSpacing {
+            DECREASED_SIZE(14),
+            COMPACT(20),
+            DEFAULT(25);
```
ryanzhoudev: Code looks fine now. I'm not sure we should name these the way they are right now. DECREASED_SIZE and COMPACT mean pretty much the same thing in my book, and there is no way for the user to determine which of the two are smaller, unless they go to the menu and check. I think something like LARGE, MEDIUM, and SMALL works just fine. No need to mention which is the default, because the "reset" button exists. Also, if you want to, you can make MEDIUM(20) the default.

src/main/java/com/wynntils/modules/map/overlays/ui/WaypointOverviewUI.java
```diff
@@ -389,10 +392,10 @@ private void setEditButtons() {
         editButtons.clear();
         int groupShift = group == ungroupedIndex ? 20 : 0;
         for (int i = 0, lim = Math.min(pageHeight, getWaypoints().size() - pageHeight * page); i < lim; i++) {
-            editButtons.add(new GuiButton(3 + 10 * i, this.width/2 + 85 + groupShift, 54 + 25 * i, 40, 20,"Edit..."));
-            editButtons.add(new GuiButton(5 + 10 * i, this.width/2 + 130 + groupShift, 54 + 25 * i, 40, 20, "Delete"));
-            GuiButton up = new GuiButton(6 + 10 * i, this.width/2 + 172 + groupShift, 54 + 25 * i, 9, 9, "\u028C");
-            GuiButton down = new GuiButton(7 + 10 * i, this.width/2 + 172 + groupShift, 65 + 25 * i, 9, 9, "v");
+            editButtons.add(new GuiButton(3 + 10 * i, this.width/2 + 85 + groupShift, 54 + spacingMultiplier * i, (int)Math.round(40.0 * (decreasedSize ? 0.7 : 1.0)), (int)Math.round(20.0 * (decreasedSize ? 0.6 : 1.0)),"Edit..."));
+            editButtons.add(new GuiButton(5 + 10 * i, this.width/2 + 130 + groupShift, 54 + spacingMultiplier * i, (int)Math.round(40.0 * (decreasedSize ? 0.9 : 1.0)), (int)Math.round(20.0 * (decreasedSize ? 0.6 : 1.0)), "Delete"));
```
ryanzhoudev, line 396: I think the text for decreasedSize could be even smaller, maybe by another 0.1 or 0.2. Also while we're here, I don't think the `...` is needed at the end of "Edit...".

playfuldoggo: Looking at the javadocs for forge 1.12, the constructor for 'GuiButton' doesn't seem to have any variable that changes the size of the text itself. I have removed the '...' at the end of 'Edit'.

src/main/java/com/wynntils/modules/map/configs/MapConfig.java
playfuldoggo: Changed the names of the variables and set the default to 'MEDIUM'.

ryanzhoudev (approved):
LGTM

magicus (approved):

PR #554, opened 2022-07-30
feat: Party management GUI enhancements

ryanzhoudev:
- PartyManager now correctly resets the party when retrieving the new party with /party list, instead of just appending new members to an invalid party
- Party management UI refreshes party status when opening with keybind
- New buttons!
    - "Refresh Party" force refreshes the party, in case anything goes wrong. Shouldn't really be necessary, but a nice to have
    - "Kick Offline", well, kicks offline players in the party. Only accessible when there are offline players & you are party leader
- Invite players has been moved, max characters extended, and a note added for bulk inviting
- Offline players now show up as ~Strikethrough~ on the player list
- Nearby players are now added to a suggestions list on the right. Nearby being a 30 block radius. (Pretty similar to Wynncraft's friend/party highlight range)
- Self now shows up as **Bold**
- Friends and nearby now have colours
- See image
- Fixes cascading lag (move static btns to init)
- Fixes static button sound being absurdly loud
image

Everything should be tested pretty well since I did the testing in a dxp weekend grind party, so tons of players in various states.
Also centered the entire UI properly. Was off by 5 pixels last time.

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/overlays/ui/PartyManagementUI.java
+++ b/src/main/java/com/wynntils/modules/utilities/overlays/ui/PartyManagementUI.java
@@ -10,6 +10,8 @@ import com.wynntils.core.framework.instances.data.SocialData;
 import com.wynntils.core.framework.rendering.ScreenRenderer;
 import com.wynntils.core.utils.Utils;
 import com.wynntils.core.utils.objects.Pair;
+import com.wynntils.modules.core.managers.PartyManager;
+import com.wynntils.modules.utilities.configs.OverlayConfig;
 import net.minecraft.client.gui.GuiButton;
 import net.minecraft.client.gui.GuiLabel;
 import net.minecraft.client.gui.GuiScreen;
@@ -19,6 +21,7 @@ import net.minecraft.client.network.NetworkPlayerInfo;
 import net.minecraft.client.renderer.GlStateManager;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.entity.player.EnumPlayerModelParts;
+import net.minecraft.scoreboard.Scoreboard;
 import net.minecraft.util.ResourceLocation;
 import net.minecraft.util.text.TextFormatting;
 import org.lwjgl.input.Keyboard;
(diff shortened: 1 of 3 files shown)


magicus:
Not related to this PR, but a general wish since you're working on the party management gui: I typically want to team up with my friends. It would be nice if there were some way to have a list of common people to invite to a party. Either by parsing the friends list and showing all your friends (or, possibly, allow you to "star" a subset of your friends to have easy access to them if you have a lot of friends), or to just have a separate list of user names in the party management gui that you could add.

ryanzhoudev:
Agreed, I'll also look in to a way to detect nearby players and add them to the suggestion list. Probably useful for grind parties

src/main/java/com/wynntils/modules/utilities/overlays/ui/PartyManagementUI.java
```diff
@@ -233,7 +287,8 @@ private void updatePartyMemberList() {
         temporaryMembers.sort(sortPartyMembers);
         partyMembers.clear();
         partyMembers.addAll(temporaryMembers);
-        // partyMembers list is now sorted in [self, owner, party alphabetically]
+        partyMembers.addAll(offlineMembers);
```
magicus, line 356: Should offline members not be sorted alphabetically as well?

magicus:
Yes, that'd be really nice. Not so simple to implement, though, I guess. Or maybe there is a "find entities within a BB" which you can use. (I have a vague memory about this). Also, I think we collect a list of your friends at login time. I'm not sure I remember correctly, nor if we actually use this for anything real, so if it is not being used it might have bit-rotted.

ryanzhoudev:
Reviews appreciated, don't merge though, keeping the gui open slowly increases lag in-game until you close the gui so I'm trying to fix that

magicus (commented):
I think this looks good.

If lag time is increasing, check that you are not accumulating stuff in some list.

kristofbolyai:
Seeing the image, does this fit on different resolutions?

ryanzhoudev:
> Seeing the image, does this fit on different resolutions?

Fits on 1020px width or larger. Imo this is fine as even 720p is 1280p wide. There's maybe another 70px maximum we could move the suggestions to the left. Legend fits great all the way down to 820px wide.

magicus (approved):
Nice! I've really been enjoying the party mgmt gui when playing on legacy. Keep up the good work.

kristofbolyai (approved):

PR #558, opened 2022-08-09
Use a proper UA instead of Mozilla's

HighCrit:


magicus (approved):
Just to be clear, I have not reviewed data flow in detail for the legacy mod, but it sounds reasonable that the same should apply here, i.e. that the VERSION should be set before WebManager gets to clinit.

(And once again, clinit blocks make no difference)

HighCrit:
> Just to be clear, I have not reviewed data flow in detail for the legacy mod, but it sounds reasonable that the same should apply here, i.e. that the VERSION should be set before WebManager gets to clinit.
> 
> (And once again, clinit blocks make no difference)

The debugger reports the expected UA

ryanzhoudev (approved):

PR #560, opened 2022-08-13
Beta improvements #7

ryanzhoudev:
- Add custom blood pool bar (in style of health/mana)
    - Also allows cancelling default boss bar
- Fix spell keybind checks (level checks)
- Fixes regex for ActionBar mana parsing
- Adds %es% or %elemental_special% for the elemental special charge <- **Remember to update FAQ on website once merged to main**
- Fixes totem tracker parsing

### DIFF
--- a/src/main/java/com/wynntils/modules/core/instances/TotemTracker.java
+++ b/src/main/java/com/wynntils/modules/core/instances/TotemTracker.java
@@ -151,11 +151,11 @@ public class TotemTracker {
     }
 
     public void onTotemSpellCast(SpellEvent.Cast e) {
-        if (e.getSpell().equals("Totem") || e.getSpell().equals("Sky Emblem")) {
+        if (e.getSpell().equals("Totem")) {
             totemCastTimestamp = System.currentTimeMillis();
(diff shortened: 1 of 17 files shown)


src/main/java/com/wynntils/modules/core/instances/TotemTracker.java
```diff
@@ -151,11 +151,11 @@ public void onTotemSpawn(PacketEvent<SPacketSpawnObject> e) {
     }
 
     public void onTotemSpellCast(SpellEvent.Cast e) {
-        if (e.getSpell().equals("Totem") || e.getSpell().equals("Sky Emblem")) {
```
magicus, line 154: Just curious, is "reskinned" spells being removed in 2.0?

ryanzhoudev: Yes, as far as I can tell. My reskinned classes now have the normal spell names, and the ability tree does not have an option to use reskinned names.

magicus: That's kind of a blessing. It was just annoying with those names...

PR #561, opened 2022-08-17
Add ability to load multiple lootruns at once

byBackfish:
Implement [Ability to load more then 1 lootrun at once.](https://feedback.wynntils.com/posts/34/ability-to-load-more-then-1-lootrun-at-once)

### DIFF
--- a/src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
+++ b/src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
@@ -53,7 +53,7 @@ public class CommandLootRun extends CommandBase implements IClientCommand {
 
     @Override
     public String getUsage(ICommandSender sender) {
-        return "lootrun <load/save/delete/rename/record/undo/note/list/folder/clear/resetdry/help>";
+        return "lootrun <current/load/unload/save/delete/rename/record/undo/note/list/folder/clear/resetdry/help>";
     }
 
     @Override
--- a/src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
+++ b/src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
@@ -9,6 +9,7 @@ import com.wynntils.core.utils.Utils;
 import com.wynntils.core.utils.helpers.Delay;
 import com.wynntils.core.utils.objects.Location;
 import com.wynntils.modules.map.instances.LootRunNote;
+import com.wynntils.modules.map.instances.LootRunPath;
 import com.wynntils.modules.map.managers.LootRunManager;
 import com.wynntils.modules.questbook.enums.QuestBookPages;
 import com.wynntils.modules.utilities.UtilitiesModule;
@@ -29,6 +30,7 @@ import net.minecraftforge.client.IClientCommand;
 
 import java.net.URI;
 import java.util.*;
+import java.util.stream.Collectors;
 
 import static net.minecraft.util.text.TextFormatting.*;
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
```diff
@@ -66,24 +68,33 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
         }
 
         switch (args[0].toLowerCase(Locale.ROOT)) {
+            case "current": {
+                if (LootRunManager.getActivePath().isEmpty()) {
```
ryanzhoudev: Refactor getActivePath() to getActivePaths() or something of the sort

src/main/java/com/wynntils/modules/map/commands/CommandLootRun.java
```diff
@@ -66,24 +68,33 @@ public void execute(MinecraftServer server, ICommandSender sender, String[] args
         }
 
         switch (args[0].toLowerCase(Locale.ROOT)) {
+            case "current": {
```
ryanzhoudev, line 71: This needs to be added to the `/lr help` command.

src/main/java/com/wynntils/modules/map/managers/LootRunManager.java
```diff
@@ -143,37 +146,31 @@ public static boolean rename(String oldName, String newName) {
...
 
     public static void clear() {
-        activePath = null;
-        activePathName = null;
+        activePaths.clear();
         recordingPath = null;
         updateMapPath();
     }
 
     public static void clearOnlyLoaded() {
```
ryanzhoudev: Remove unused method

src/main/java/com/wynntils/modules/map/managers/LootRunManager.java
```diff
@@ -311,12 +325,25 @@ public static void renderActivePaths() {
...
         }
     }
 
+    public static void clearLootrun(String name) {
+        activePaths.remove(name);
+        recordingPath = null;
+        updateMapPath();
+    }
+
+    public static void unloadLootrun(String name) {
```
ryanzhoudev: Since this already exists, would be nice to see `/lr unload`, similar to `/lr load`

ryanzhoudev (changes requested):

ryanzhoudev:
Would also be nice to see lootruns being colored differently to each other. All of them being the same shade of blue is messy

byBackfish:
> Would also be nice to see lootruns being colored differently to each other. All of them being the same shade of blue is messy

Yea, that would be a nice idea. I'll look into it and then make it a setting I guess. I think having no way to turn that off wouldn't be that great in some cases

src/main/java/com/wynntils/modules/map/configs/MapConfig.java
```diff
@@ -277,6 +278,9 @@ public static class LootRun extends SettingsClass {
         @Setting(displayName = "Show Paths as Rainbow", description = "Should paths be shown in the colours of a rainbow?", order = 4)
         public boolean rainbowLootRun = false;
 
+        @Setting(displayName = "Show different colors if multiple", description = "Should paths be colored differently if multiple lootruns are active?\nDoes not work with rainbow paths enabled", order = 4)
```
ryanzhoudev:
```suggestion
        @Setting(displayName = "Differentiate Path Colors", description = "Should paths be colored differently if multiple lootruns are active?\nDoes not work with rainbow paths enabled", order = 4)
```

src/main/java/com/wynntils/modules/map/configs/MapConfig.java
```diff
@@ -277,6 +278,9 @@ public static class LootRun extends SettingsClass {
...
+        public boolean differentColorsMultipleLootruns = false;
```
ryanzhoudev: I think having this on by default is better. As far as I can tell there is no drawback to doing so.

src/main/java/com/wynntils/modules/map/managers/LootRunManager.java
```diff
@@ -310,13 +312,38 @@ public static void renderActivePaths() {
         }
     }
 
+    private static CustomColor getPathColor(LootRunPath path){
+        CustomColor defaultColor = MapConfig.LootRun.INSTANCE.rainbowLootRun ? CommonColors.RAINBOW : MapConfig.LootRun.INSTANCE.activePathColour;
+        if (!MapConfig.LootRun.INSTANCE.differentColorsMultipleLootruns || getActivePaths().size() == 1 || MapConfig.LootRun.INSTANCE.rainbowLootRun) return defaultColor;
+
+        int index = ArrayUtils.indexOf(getActivePaths().values().toArray(), path) % CommonColors.colors.length;
+        return CommonColors.colors[index];
```
ryanzhoudev: This needs to be changed. CommonColors.colors[0] is black, which can be difficult to see in dark areas. I think making your own list here with suitable colors (brighter ones) would be better. This also allows CommonColors.colors to stay private.

ryanzhoudev (changes requested):
Also, `/lr` (the usage exception) does not list the new commands.

ryanzhoudev (approved):
This looks good in-game. However, lootruns that mark the same chests will result in the chest only being marked one color. I personally think this is a nonissue, but not sure about everyone else.

magicus (approved):

PR #563, opened 2022-08-18
Add trade market sell amount display/overlay

byBackfish:
Pretty self explanatory
image

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/overlays/inventories/TradeMarketOverlay.java
+++ b/src/main/java/com/wynntils/modules/utilities/overlays/inventories/TradeMarketOverlay.java
@@ -28,6 +28,7 @@ public class TradeMarketOverlay implements Listener {
             return;
         }
         if (inventory.getStackInSlot(20) != ItemStack.EMPTY && !inventory.getStackInSlot(20).getDisplayName().isEmpty()) return;
+        if (getItemName(inventory) == null || getAmount(getItemName(inventory)) <= 1) return;
 
         ItemStack itemStack = new ItemStack(buildNBT(inventory));
 
@@ -62,7 +63,7 @@ public class TradeMarketOverlay implements Listener {
         shouldSend = false;
     }
 
-    public NBTTagCompound buildNBT(InventoryBasic inventory) {
+    private NBTTagCompound buildNBT(InventoryBasic inventory) {
         String itemName = getItemName(inventory);
         int amount = itemName != null ? getAmount(itemName) : 1;
 
@@ -94,7 +95,7 @@ public class TradeMarketOverlay implements Listener {
         return nbt;
     }
 
-    public int getAmount(String name) {
+    private int getAmount(String name) {
         int amount = 0;
 
         for (ItemStack item : McIf.player().inventory.mainInventory) {
@@ -107,7 +108,7 @@ public class TradeMarketOverlay implements Listener {
         return amount;
     }
 
-    public String getItemName(InventoryBasic inventory) {
+    private String getItemName(InventoryBasic inventory) {
         ItemStack itemStack = inventory.getStackInSlot(10);
         if (itemStack == ItemStack.EMPTY) return null;
         if (itemStack.getDisplayName().contains("Click an Item to sell")) return null;
--- a/src/main/java/com/wynntils/modules/utilities/overlays/inventories/TradeMarketOverlay.java
+++ b/src/main/java/com/wynntils/modules/utilities/overlays/inventories/TradeMarketOverlay.java
@@ -29,8 +29,8 @@ public class TradeMarketOverlay implements Listener {
         }
         if (inventory.getStackInSlot(20) != ItemStack.EMPTY && !inventory.getStackInSlot(20).getDisplayName().isEmpty()) return;
 
-        ItemStack itemStack = new ItemStack(buildNBT());
-        itemStack.setTagCompound(buildNBT());
+        ItemStack itemStack = new ItemStack(buildNBT(inventory));
+        itemStack.setTagCompound(buildNBT(inventory));
 
         inventory.setInventorySlotContents(20, itemStack);
         shouldSend = true;
@@ -42,28 +42,13 @@ public class TradeMarketOverlay implements Listener {
         if (!inventory.getName().contains("What would you like to sell?")) return;
         if (e.getSlotId() != 20) return;
 
-        ItemStack itemStack = inventory.getStackInSlot(10);
-
-        if (itemStack == ItemStack.EMPTY) return;
-        if (itemStack.getDisplayName().contains("Click an Item to sell")) return;
-
-        Matcher matcher = ITEM_NAME_PATTERN.matcher(itemStack.getDisplayName());
-
-        if (!matcher.matches()) return;
-
-        String itemName = matcher.group(1);
-        int amount = 0;
-
-        for (ItemStack item : McIf.player().inventory.mainInventory) {
-            if (item.getDisplayName().toLowerCase().trim().equals(itemName.toLowerCase().trim())) {
-                int itemAmount = item.getCount();
-                amount += itemAmount;
-            }
-        }
+        String itemName = getItemName(inventory);
+        if(itemName == null) return;
 
         int slotId = 11;
         shouldSend = true;
-        amountToSend = amount;
+        amountToSend = getAmount(itemName);
+
         e.getGui().handleMouseClick(e.getGui().inventorySlots.getSlot(slotId), slotId, 2, ClickType.PICKUP);
     }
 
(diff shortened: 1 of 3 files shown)


magicus:
Is the idea to help the user set the right amount?

Can't you take it a step further, so if you e.g. right click on the amount button, the determined amount is sent as a chat message?

byBackfish:
> Is the idea to help the user set the right amount?
> 
> Can't you take it a step further, so if you e.g. right click on the amount button, the determined amount is sent as a chat message?

Good idea. I do not know why i did not think of this earlier. Ty for the suggestion

ryanzhoudev:
https://feedback.wynntils.com/posts/21/a-sell-all-button

src/main/java/com/wynntils/modules/utilities/overlays/inventories/TradeMarketOverlay.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class TradeMarketOverlay implements Listener {
+
+    private final Pattern ITEM_NAME_PATTERN = Pattern.compile(".*Selling §f\\d+ (.*) for .* Each");
+
+    @SubscribeEvent
+    public void onRender(GuiOverlapEvent.ChestOverlap.HandleMouseClick e) {
```
kristofbolyai: Can we get some new lines? This code feels so crammed.

src/main/java/com/wynntils/modules/utilities/overlays/inventories/TradeMarketOverlay.java
```diff
@@ -0,0 +1,44 @@
...
+        if (McIf.player() == null) return;
+        if (McIf.player().openContainer == null) return;
+        InventoryBasic inventory = (InventoryBasic) e.getGui().getLowerInv();
+        if (!inventory.getName().contains("What would you like to sell?")) return;
+        if(e.getSlotId() != 11) return;
+        ItemStack itemStack = inventory.getStackInSlot(10);
+        if (itemStack == ItemStack.EMPTY) return;
+        if (itemStack.getDisplayName().contains("Click an Item to Sell")) return;
+        System.out.println(itemStack.getDisplayName());
```
kristofbolyai: Debug line?

kristofbolyai (changes requested):

magicus:
> https://feedback.wynntils.com/posts/21/a-sell-all-button

@DonkeyBlaster Thanks for doing the work with confirming approval from the Wynncraft team, for this and other potential additions!

magicus:
@byBackfish You misunderstood me. I meant that Wynntils should take care of sending the reply in chat for the user. That is, if the user right clicks on the plus, Wynntils will do the following:
 * Fake a click on the "+" icon.
 * Send the amount of items as a chat message sent *from* the user.

I realize right-clicking might be both harder to do, and less visible, so maybe a better way to trigger this would be to add an additional "+" icon below (or above, whichever looks best), give it the tooltip "Add all items of this kind", and when that is clicked, do the "dance" above.

If this needs to much knowledge of Wynntils inner workings for you to pull off, you can go with the current solution. But if possible, I think it would be much nicer to have.

magicus:
(Also, that new "+" item could have the number you are going to add as index/"amount") so you can directly see the result of your action before doing it.

byBackfish:
@magicus oh alright. I tried my best, hope this is fine
image

byBackfish:
Actually, I thought an Emerald would look nicer than the "+", but how about making it a green checkmark?

magicus:
Yeah, the "+" was just a suggestion. If there were a kind of "++" icon, I think that'd be the best, but whatever. I think the check mark should be reserved for executing the sale. 

The drawback with the emerald is that it is already used to signal setting the price. 

Maybe using a full pouch icon?

src/main/java/com/wynntils/modules/utilities/overlays/inventories/TradeMarketOverlay.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    public NBTTagCompound buildNBT(InventoryBasic inventory) {
+        String itemName = getItemName(inventory);
+
+        NBTTagCompound nbt = new NBTTagCompound();
+        NBTTagCompound tag = new NBTTagCompound();
+
+        nbt.setString("id", "minecraft:emerald");
+        nbt.setInteger("Count", 1);
```
magicus: Try setting this to `getAmount(itemName)`. It should work up to 64 at least, I believe. I don't know if our code to display higher values is general enough to pick this up, but maybe. It's better than nothing, I think.

src/main/java/com/wynntils/modules/utilities/overlays/inventories/TradeMarketOverlay.java
```diff
@@ -0,0 +1,120 @@
...
+
+        NBTTagCompound display = new NBTTagCompound();
+        display.setString("Name", "§aSell All");
+
+        NBTTagList lore = new NBTTagList();
+        lore.appendTag(new NBTTagString("§7Click to sell all items in your inventory"));
```
magicus, line 79:
```suggestion
        lore.appendTag(new NBTTagString("§7Click to sell all " + getAmount(itemName) + " items in your inventory"));
```

magicus:
If you want to test the ingredient pouch idea, set `id` to `minecraft:diamond_axe` and `Damage` to `93`. Unfortunately, I did not have a character with a full ingredient pouch (and no time to fill one up right now), so I can't give you a damage value for the full pouch icon.

byBackfish:
Yea, I have those damage values but thanks! Wouldn't it be better to make it a full emerald pouch instead of a full ingredient pouch?

magicus:
Maaaaaybe...? But I think my point here is that the first two slots is what you sell (item + amount), and the third is the price. And that "price" is associated with emeralds, but the other two are not. So I think the association with emeralds is a bit odd; it feels more like "set a maximum price" or so, rather than "sell all that I got". 

Ideally, I think we should have some kind of stacking +:es, something along these lines: http://svn.slimdevices.com/repos/jive/7.7/trunk/squeezeplay/assets/MasterIcons/icon_add_all.png. But creating new icons is tricky and kind of out of the scope for this PR.

byBackfish:
Once the count is above 64, the item just simply does not display. It now uses 64 if the amount would be above 64.

image

magicus:
I think this looks great! :-)

magicus (dismissed):

magicus:
(An alternative would be to set count to 1 if above 64, to have no number at all instead of an incorrect. But I'm not even sure that's better.)

ryanzhoudev (dismissed):
Very nice

Actually, you could make a few of the methods in TradeMarketOverlay private. (getAmount, getItemName, buildNBT)

src/main/java/com/wynntils/modules/utilities/overlays/inventories/TradeMarketOverlay.java
```diff
@@ -0,0 +1,121 @@
...
+
+    @SubscribeEvent
+    public void onGuiUpdate(GuiOverlapEvent.ChestOverlap.DrawScreen e) {
+        InventoryBasic inventory = (InventoryBasic) e.getGui().getLowerInv();
+        if (!inventory.getName().startsWith("What would you like to sell?")) {
+            shouldSend = false;
+            return;
+        }
+        if (inventory.getStackInSlot(20) != ItemStack.EMPTY && !inventory.getStackInSlot(20).getDisplayName().isEmpty()) return;
+
```
magicus, line 32: Thinking about it a bit more... Maybe it only makes sense to display this icon if you have more than one item to sell? So if `getAmount(itemName) <= 1` you'd just return here. What do you think? Or do we always want this icon, even if you have only a single object of that kind?

ryanzhoudev (dismissed):

magicus (dismissed):

src/main/java/com/wynntils/modules/utilities/overlays/inventories/TradeMarketOverlay.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @SubscribeEvent
+    public void onClick(GuiOverlapEvent.ChestOverlap.HandleMouseClick e) {
+        InventoryBasic inventory = (InventoryBasic) e.getGui().getLowerInv();
+        if (!inventory.getName().contains("What would you like to sell?")) return;
+        if (e.getSlotId() != 20) return;
+
+        String itemName = getItemName(inventory);
+        if(itemName == null) return;
```
kristofbolyai:
```suggestion
        if (itemName == null) return;
```

kristofbolyai (approved):

magicus (approved):

PR #564, opened 2022-08-19
feat: Add glowing totems

ryanzhoudev:
This is in OverlayEvents so it can piggyback off the consumable timer setting for the totem, though it can be (easily) moved to TotemTracker if needed.

Also no, we can't use the existing totem "entity" in TotemTracker. That thing is just the nametag, not the actual totem.

For future reference, legality of this is confirmed in https://discord.com/channels/394189072635133952/405350459126317065/1009605568618582076.

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
+++ b/src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
@@ -20,10 +20,12 @@ import com.wynntils.modules.utilities.managers.MountHorseManager;
 import com.wynntils.modules.utilities.overlays.hud.*;
 import com.wynntils.webapi.WebManager;
 import com.wynntils.webapi.profiles.item.enums.ItemTier;
+import net.minecraft.entity.item.EntityArmorStand;
 import net.minecraft.init.MobEffects;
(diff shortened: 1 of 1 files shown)


kristofbolyai:
Does this also make other players totems glow?

byBackfish:
> Does this also make other players totems glow?

Maybe (if possible), make it a setting whether it should make other people's totems glow

ryanzhoudev:
> Does this also make other players totems glow?

No, this is your own totem only. Same check as the totem timer.

ryanzhoudev:
> > Does this also make other players totems glow?
> 
> Maybe (if possible), make it a setting whether it should make other people's totems glow

That is very illegal

byBackfish:
> > > Does this also make other players totems glow?
> 
> > 
> 
> > Maybe (if possible), make it a setting whether it should make other people's totems glow
> 
> 
> 
> That is very illegal

Okay yea, I didn't think of the ways on how to abuse this. Would be very op if you could actually see the totems of hunters etc through walls

kristofbolyai (approved):
Seems fine then.

HighCrit (approved):

PR #565, opened 2022-08-20
feat: click to copy message

byBackfish:
Adds a component to every message, that on click copies that message content to the clipboard.
Pressing `L_CONTROL` whilst clicking on the component copies the raw message, including the color codes


src/main/java/com/wynntils/modules/chat/events/ClientEvents.java
```diff
@@ -4,13 +4,15 @@
 
 package com.wynntils.modules.chat.events;
 
+import com.mojang.realmsclient.gui.ChatFormatting;
```
kristofbolyai: Is this used?

byBackfish: yea, it is used for `ChatFormatting#stripFormatting`, I know that `TextFormatting#getTextWithoutFormattingCodes` also exists, but I have always used `ChatFormatting` for that. If you want, then i can change it to use `TextFormatting` instead

kristofbolyai: We use ChatFormatting in 1.18.2 because it is renamed there.

byBackfish: Yea, I've alway just used `ChatFormatting` in my previous mods, and never thought about switching to `TextFormatting`. I'll change it

src/main/java/com/wynntils/modules/chat/events/ClientEvents.java
```diff
@@ -183,4 +191,28 @@ public void onClickNPC(PacketEvent<CPacketUseEntity> e) {
         ChatManager.progressDialogue();
     }
 
+    @SubscribeEvent
+    public void onMessageReceived(ClientChatReceivedEvent e) {
```
kristofbolyai, line 195: I think this does make every message with a hover or click effect sent from Wynncraft lose their events. Did you test that?

byBackfish: Pretty sure it does not, testing once more just to make sure. It only adds a component, and does not change the existing ones

byBackfish: Yep, seems to work just fine

kristofbolyai: Ah, I miss read the code. But. Make this toggle able and off by default

HighCrit (approved):

kristofbolyai (approved):

PR #567, opened 2022-09-06
Fixes a StringIndexOutOfBounds exception

byBackfish:
When putting a crafted item into the skill point gui, and that item has the same name as one of the checked items (like `Daily Rewards`), and then does not have the lore that Wynntils is parsing for, it will crash


kristofbolyai:
Why not check if one of the indexes are incorrect later?

```java
            } else if (name.contains("Daily Rewards")) { //Daily Reward Multiplier
                int start = lore.indexOf("Streak Multiplier: ") + 19;
                int end = lore.indexOf("Log in everyday to");
                value = Integer.parseInt(lore.substring(start, end));
```

kristofbolyai:
start (and end) will be -1 there

byBackfish:
> start (and end) will be -1 there

well, crafted items could still be used for other items like `Damage Info`, this makes it so every check skips crafted items

kristofbolyai:
So we just skip crafted items and ignore that our parsing logic is pretty much broken?
I would rather put everything in a try-catch then (even though this is still bad)..

byBackfish:
Well yea, it's not the best to just ignore it, but checking if the index is -1 could also have problems then as you can change the lore of a crafted item. You could still give it to someone and then crash them as a "troll". And all those issues are caused by crafted items. Best would probably be to check for -1 and then check make a new method that safely parses the int and returns 0 if its not actually containing an int

kristofbolyai (approved):
Fine by me.

ryanzhoudev (approved):

PR #571, opened 2022-09-08
Auto changelog + Curseforge deployment

DevScyu:


.pre-commit.js
```diff
@@ -0,0 +1,34 @@
...
+exports.preCommit = (props) => {
+    const replace = (path, searchValue, replaceValue) => {
+        let content = fs.readFileSync(path, "utf-8");
+        if (content.match(searchValue)) {
+            fs.writeFileSync(path, content.replace(searchValue, replaceValue));
+            console.log(`"${path}" changed`);
+        }
+    };
+
+    const replaceAll = (startPath, filter, searchValue, replaceValue) => {
```
magicus: What is the purpose of this?

.github/workflows/release.yml
```diff
@@ -0,0 +1,75 @@
...
+      - name: Upload to CurseForge
+        if: ${{ steps.changelog.outputs.skipped == 'false' }}
+        uses: itsmeow/curseforge-upload@v3
+        with:
+          file_path: build/Wynntils-v${{ steps.changelog.outputs.version }}.jar
+          game_endpoint: minecraft
+          # Minecraft 1.12.2: 6756
+          # Forge 14.23.5.2859: 8867
+          # Java 8: 4458
+          game_versions: "Minecraft 1.12:1.12.2,java:Java 8,Forge"
```
HighCrit: The docs advise to use the numerical ids https://github.com/itsmeow/curseforge-upload#api-method just like the curseforge docs

.github/workflows/release.yml
```diff
@@ -0,0 +1,75 @@
...
+          version-file: ./version.json
+
+      - name: Set up JDK 8
+        uses: actions/setup-java@v2
+        with:
+          distribution: "temurin"
+          java-version: 8
+          cache: "gradle"
+
+      - name: Build
```
HighCrit: It also seems a bit waste full to do a second build of the project. Is there no way to reuse it?

HighCrit (changes requested):

.pre-commit.js
```diff
@@ -0,0 +1,34 @@
+const fs = require("fs");
```
HighCrit, line 1: Could we not use sngrep to get the version number from build.gradle? Or set the version in build.gradle using environment variables?

.github/workflows/release.yml
HighCrit: Example: https://github.com/null511/MCRP-Textureless/blob/501dcc639dce1e39247dc51c95443808e11dbeac/.github/workflows/Release.yml#L67

PR #583, opened 2022-09-16
fix: old waypoints being invalid when trying to import

byBackfish:
Fixes a bug with importing old waypoints

(wynntils thought the `showBeaconBeam` field was in there, even tho it wasn't so always showed as invalid)


byBackfish:
I am doing some more tests on this right now, but it seems to work fine

src/main/java/com/wynntils/modules/map/instances/WaypointProfile.java
```diff
@@ -274,7 +280,10 @@ public void decode(byte format, ByteBuffer buf) throws IllegalArgumentException,
         } else {
             this.group = WaypointType.values()[group];
         }
-        this.showBeaconBeam = buf.get() == 1;
+
+        if(format == 2) {
```
kristofbolyai:
```suggestion
        if (format == 2) {
```

src/main/java/com/wynntils/modules/map/instances/WaypointProfile.java
```diff
@@ -89,13 +89,13 @@ public void setGroup(@Nullable WaypointType group) {
...
 
-    public static final byte currentFormat = 1;
+    public static final byte currentFormat = 2;
 
     /**
      * Returns an upper bound of the length (in bytes) encoding this waypoint will be (with the given format)
      */
     public int encodeLength(byte format) {
-        assert 0 <= format && format <= 1;
+        assert 0 <= format && format <= 2;
```
kristofbolyai: It is a bit weird all these asserts to not use `currentFormat` as the upper limit. We can perhaps fix that now?

kristofbolyai (approved):

HighCrit (approved):

PR #584, opened 2022-09-16
feat: custom trade market sell amount buttons

byBackfish:
Implements [add-more-options-in-the-selling-menu-of-tm](https://feedback.wynntils.com/posts/103/add-more-options-in-the-selling-menu-of-tm)

Adds 3 buttons to the trade market gui, pressing them will automatically put the set amount (in the config) into the chat. 
Fixes various bugs
TradeMarketOverlay now uses the nbt tag to determine the amount of the item to sell


byBackfish:
image

src/main/java/com/wynntils/modules/utilities/configs/UtilitiesConfig.java
```diff
@@ -495,6 +495,16 @@ public static class Market extends SettingsClass {
...
         public boolean showSellAllButton = true;
+
+        @Setting(displayName = "Custom Sell button 1 amount", description = "What amount should be selected when the custom sell button 1 is clicked?\n\n§8Set the value to 0 to disable this")
+        public String customSellButton1 = "64";
+
+        @Setting(displayName = "Custom Sell button 2 amount", description = "What amount should be selected when the custom sell button 2 is clicked?\n\n§8Set the value to 0 to disable this")
+        public String customSellButton2 = "1728";
+
+        @Setting(displayName = "Custom Sell button 3 amount", description = "What amount should be selected when the custom sell button 3 is clicked?\n\n§8Set the value to 0 to disable this")
+        public String customSellButton3 = "6399";
```
kristofbolyai, line 506: String for an integer setting?

kristofbolyai: I assume the problem is that integer configs don't really render properly without a limit, due to how the config GUI is written?

kristofbolyai (commented):
Also, was the bug fixed where the sell all button inputted numbers even when the user did not press the button?

src/main/java/com/wynntils/modules/utilities/configs/UtilitiesConfig.java
byBackfish: yea, and if i set a limit (with the max bound of 6399, which is the trade market sell limit), it goes up in steps of 30+

byBackfish:
yea, that bug was fixed already

src/main/java/com/wynntils/modules/utilities/configs/UtilitiesConfig.java
kristofbolyai: aah. this is sad

kristofbolyai: We can limit the string field to numbers only though by checking config updates and reverting if a non number char was added.

byBackfish: i assume with `onSettingChanged`? How would i revert/cancel it tho? Does it save the previous value anywhere?

kristofbolyai: I think it does in arguments? If it does not, let's leave it like this.

byBackfish: the only argument is `name`, so sadly not

src/main/java/com/wynntils/modules/utilities/overlays/inventories/TradeMarketOverlay.java
```diff
@@ -117,7 +138,24 @@ private String getItemName(InventoryBasic inventory) {
         Matcher matcher = ITEM_NAME_PATTERN.matcher(itemStack.getDisplayName());
 
         if (!matcher.matches()) return null;
-        return matcher.group(1);
+        return matcher.group(2);
+    }
+
+    private void addCustomSellItems(String itemName, InventoryBasic inventory) {
+        if (StringUtils.parseIntOr(UtilitiesConfig.Market.INSTANCE.customSellButton1, 0) > 0) {
```
kristofbolyai: Don't parse everything twice?

kristofbolyai (dismissed):

src/main/java/com/wynntils/modules/utilities/configs/UtilitiesConfig.java
HighCrit, line 506: Is there demand for this?

byBackfish: Yes, it's very convenient, especially for people that grind professions. They often sell the max amount possible on the trade market (6399) and this makes it a bit easier

ryanzhoudev:
I think the button and text for this is rather large, is the settings text necessary?
image

ryanzhoudev:
> I think the button and text for this is rather large, is the settings text necessary?

Also, I think the "Click to sell all items in your inventory" shouldn't be there

ryanzhoudev (dismissed):

magicus (dismissed):

magicus (approved):

kristofbolyai (approved):

PR #593, opened 2022-09-21
[RELEASE] v1.13.0 

DevScyu:


ryanzhoudev:
How do we feel about the default bar placement for awakened, focus, and the shaman mask? Apparently the awakened and focus ones are exactly on top of each other and can be difficult to find for people

kristofbolyai:
> How do we feel about the default bar placement for awakened, focus, and the shaman mask? Apparently the awakened and focus ones are exactly on top of each other and can be difficult to find for people

They are on top of each other because of the different classes. This works well in Artemis's overlay management, but is a bit weird here.

ryanzhoudev:
It's not possible to have focus and awakened at the same time, yea? Since one is archer and the other is shaman? We could combine the bars into one single bar and have it dynamically do stuff based on what class is selected

byBackfish:
> It's not possible to have focus and awakened at the same time, yea? Since one is archer and the other is shaman? We could combine the bars into one single bar and have it dynamically do stuff based on what class is selected

Could also merge the mana bank bar into that then, same with the corrupted bar (once its added, might add it tomorrow if no one else has started it)

P0keDev (approved):

ryanzhoudev (approved):

kristofbolyai (approved):

PR #594, opened 2022-09-22
fix: mythic detection & /guild list spam

byBackfish:
Fixes mythics in chests not being detected by Wynntils
Removes the `/guild list` spam 

The problem with the mythic's not being detected, was that Wynn sends two item change packets. One with empty items, one with the actual items. This PR makes Wynntils ignore the first, empty, packet and only handle the second, actual, packet. Before, Wynntils would only check the first packet, and then ignore the second, as the `lastProcessedChestID` was the same as the ID of the second packet

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
+++ b/src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
@@ -4,7 +4,6 @@
 
 package com.wynntils.modules.utilities.events;
 
-import com.google.common.eventbus.Subscribe;
 import com.wynntils.McIf;
 import com.wynntils.Reference;
 import com.wynntils.core.events.custom.*;
@@ -87,7 +86,8 @@ import java.time.format.DateTimeFormatter;
 import java.time.format.FormatStyle;
 import java.util.*;
 import java.util.regex.Matcher;
-import java.util.regex.Pattern;import java.util.stream.Collectors;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import static com.wynntils.core.framework.instances.PlayerInfo.get;
 import static net.minecraft.util.text.TextFormatting.getTextWithoutFormattingCodes;
--- a/src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
+++ b/src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
@@ -1146,6 +1146,8 @@ public class ClientEvents implements Listener {
         // Only run at first time we get items, don't care about updating
         if (e.getPacket().getWindowId() == lastProcessedOpenedChest) return;
 
+        // Wynncraft sends you two packets when opening the chest. The first is empty, the second one has the actual items.
+        // By returning here, we only process the second packet, whilst still only handling each chest once.
         List<ItemStack> actualItems = e.getPacket().getItemStacks().subList(0, 27).stream().filter(item -> !item.isEmpty() && item.hasDisplayName() && item.getItem() != Items.AIR).collect(Collectors.toList());
         if (actualItems.size() == 0) return;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
```diff
@@ -1146,6 +1146,9 @@ public void onMythicFound(PacketEvent<SPacketWindowItems> e) {
         // Only run at first time we get items, don't care about updating
         if (e.getPacket().getWindowId() == lastProcessedOpenedChest) return;
 
+        List<ItemStack> actualItems = e.getPacket().getItemStacks().subList(0, 27).stream().filter(item -> !item.isEmpty() && item.hasDisplayName() && !item.getDisplayName().equalsIgnoreCase("AIR")).collect(Collectors.toList());
```
kristofbolyai: You can do smth like `item.getItem() != Items.AIR`

ryanzhoudev (dismissed):
Checked in-game, works fine.

src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java
```diff
@@ -1191,28 +1193,6 @@ public void onMythicFound(PacketEvent<SPacketWindowItems> e) {
...
-        // Size should be at least 27, checked for it earlier
-        int size = 27;
-        for (int i = 0; i < size; i++) {
-            ItemStack stack = e.getPacket().getItemStacks().get(i);
-            if (stack.isEmpty() || !stack.hasDisplayName()) continue;
-            if (!stack.getDisplayName().contains(TextFormatting.DARK_PURPLE.toString())) continue;
-            if (!stack.getDisplayName().contains("Unidentified")) continue;
-
-            try {
-                SoundTrackManager.findTrack(WebManager.getMusicLocations().getEntryTrack("mythicFound"),
```
magicus, line 1209: Is this not needed? How is the sound played then?

byBackfish: The sound was played here https://github.com/Wynntils/Wynntils/blob/development/src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java#L1167 

And here https://github.com/Wynntils/Wynntils/blob/development/src/main/java/com/wynntils/modules/utilities/events/ClientEvents.java#L1209

byBackfish: And it seems like the second one is not needed

magicus: I see. Thanks!

magicus:
I looked through the code with more context. I just can't understand how this patch is supposed to fix the stated problem "mythics in chests not being detected by Wynntils". All you are doing is adding a quick way out if there are no "actual" items in the container. I was thinking I'd see another use of actualItems where it was iterated over, looking for a mythic.

byBackfish:
Wynncraft sends the chest items twice. The first time, everything is empty. Shortly after, the actual items "generate". So that the chest doesn't get handled twice, there is a check "if last Processed Chest == this Chest, then return". This would only take the first, empty, chest contents, and ignore the actual items. By adding this check, if the chest contains items or is empty, and then returning early. By doing this, the lastProcessedChestId is only set, if the chest contains items. This way, the first packet with the empty items is ignored, and only the second packet with the actual items is handled

magicus:
Aha, I see. I think it would be great if you could include this description as a comment (possibly shorted down a bit).

byBackfish:
👍

magicus (commented):
Please include a comment describing the purpose of the code you added, like you wrote in the PR

magicus (approved):

ryanzhoudev (approved):

PR #597, opened 2022-09-28
fix: improve skill point loadouts

ryanzhoudev:
So, the old method for getting skill point data was to have Wynntils scroll through the player info (head itemstack) menu by sending clicks and parsing. This was kind of bad and messy. It also broke if you loaded a new build, reset your sp, then loaded another build.

This pr changes the method for getting skill point data. Now, we will save the skill points as soon as the player opens the compass menu, then get all skill points from gear (armour + accessories) by going through the itemstack lores. With this information, it is possible to calculate what tomes the player has on. Now, we have four sets of information: `gearSP`, `tomeSP`, `currentSPActual`, `currentSPWithoutGearTome`.

gearSP will only contain the modifiers that exist because of equipped gear.
tomeSP will only contain the modifiers that exist because of equipped tomes.
currentSPActual is the SP show in the menu (what the players sees)
currentSPWithoutGearTome is what we use to calculate how many more SP to allocate to any given element.

Tested with tomes and accessories that give positive SP. Didn't have time to find gear that does negative, maybe I'll do it when I have time again


P0keDev:
what happens if the player has a set bonus that gives skill points (ie morph)? i'm guessing it would just be calculated as extra tome sp which shouldn't be a problem - although it might be good to rename the variable to something more general - but that's something to check alongside negative sp

ryanzhoudev:
> what happens if the player has a set bonus that gives skill points (ie morph)? i'm guessing it would just be calculated as extra tome sp which shouldn't be a problem - although it might be good to rename the variable to something more general - but that's something to check alongside negative sp

I think the regex on the armour would pick up the set bonus, if it was enabled. Might end up with us having like 21*5*4 extra skill points. I'll take a look, but yeah if that doesn't happen I'll rename the variable

src/main/java/com/wynntils/modules/utilities/overlays/inventories/SkillPointOverlay.java
```diff
@@ -74,69 +73,78 @@ public class SkillPointOverlay implements Listener {
...
+        if (shouldUpdateSkillPoints) {
+            // Reset all skill point data
+            currentSPActual = getSkillPoints(e.getGui()).getAsArray();
+            gearSP = new int[5];
+            tomeAndSetBonusSP = new int[5];
+
+            // Set gearSkillPoints to the number of skill points that our armour and accessories give us
+            // We will do this by going through all our armour and accessory slots and adding the skill point modifiers to gearSkillPoints
+            // 36-39 inclusive is armour, 9-12 inclusive is accessories
+            int[] gearSlotsToCheck = {36, 37, 38, 39, 9, 10, 11, 12};
```
kristofbolyai: make this a static field?

src/main/java/com/wynntils/modules/utilities/overlays/inventories/SkillPointOverlay.java
```diff
@@ -196,63 +204,22 @@ public void onChestForeground(GuiOverlapEvent.ChestOverlap.DrawGuiContainerForeg
...
+
         if (e.getSlotId() == SAVE_SLOT) {
             e.setCanceled(true);
-            if (waitingForSkillPointData) return;
             nameField = new GuiTextFieldWynn(200, McIf.mc().fontRenderer, 8, 5, 130, 10);
             nameField.setFocused(true);
             nameField.setText("Enter build name");
             Keyboard.enableRepeatEvents(true);
-            skipResettingSkillPointData = true;
+
```
kristofbolyai: random newline

src/main/java/com/wynntils/modules/utilities/overlays/inventories/SkillPointOverlay.java
```diff
@@ -322,22 +293,37 @@ private SkillPointAllocation getSkillPoints(ChestReplacer gui) {
...
         }
 
         return new SkillPointAllocation(sp[0], sp[1], sp[2], sp[3], sp[4]);
     }
 
     public void loadBuild(SkillPointAllocation build) {
-        if (build.getTotalSkillPoints() > skillPointsRemaining) {
-            TextComponentString text = new TextComponentString("Not enough free skill points!");
+        String errorMsg = "";
+        System.out.println(currentSPActual[0] + " " + currentSPActual[1] + " " + currentSPActual[2] + " " + currentSPActual[3] + " " + currentSPActual[4]);
```
kristofbolyai: debug

src/main/java/com/wynntils/modules/utilities/overlays/inventories/SkillPointOverlay.java
```diff
@@ -322,22 +293,37 @@ private SkillPointAllocation getSkillPoints(ChestReplacer gui) {
...
+            errorMsg = "You must reset your skill points before loading a build.";
+        } else if (build.getStrength() > skillPointsRemaining + currentSPWithoutGearTome[0] ||
+                build.getDexterity() > skillPointsRemaining + currentSPWithoutGearTome[1] ||
+                build.getIntelligence() > skillPointsRemaining + currentSPWithoutGearTome[2] ||
+                build.getDefence() > skillPointsRemaining + currentSPWithoutGearTome[3] ||
+                build.getAgility() > skillPointsRemaining + currentSPWithoutGearTome[4]) {
+            errorMsg = "You don't have enough skill points to load this build.";
+        }
+
+        if (errorMsg.length() > 0) {
```
kristofbolyai: ! .isEmpty

kristofbolyai (commented):
Looks fine otherwise

P0keDev (approved):

kristofbolyai (approved):

Incompleteusern (approved):

PR #598, opened 2022-09-29
feat: multiple totem tracking

ryanzhoudev:
Adds multiple totem tracking.
Allows totems to update their locations after being casted (one of the abilities on shaman allows this)
Re-adds working totem highlights: Adds colored totem highlights and corresponding color in timer list.

Current flaw is that it takes one second after the totem lands to highlight the totem. This happens because we need to make sure we are highlighting the correct entity, and the check is done every second (when the timer updates). Can probably be fixed soon


ryanzhoudev:
Didn't wanna deal with merge conflicts so I rewrote it better #651

PR #599, opened 2022-09-30
refactor: applying minor caching of variables in a few files.

TheAvonian:
Very minor changes to assure no null accesses (even though it shouldn't be possible in these places) and caching `SocialData` in several methods where there is heavy usage of `PlayerInfo.get(SocialData.class)`.

### DIFF
--- a/src/main/java/com/wynntils/modules/core/events/ServerEvents.java
+++ b/src/main/java/com/wynntils/modules/core/events/ServerEvents.java
@@ -98,14 +98,14 @@ public class ServerEvents implements Listener {
      */
     @SubscribeEvent(priority = EventPriority.LOWEST)
     public void joinWorldEvent(WynnWorldEvent.Join e) {
-        if (PlayerInfo.get(CharacterData.class).getClassId() == -1 || CoreDBConfig.INSTANCE.lastClass == ClassType.NONE)
+        CharacterData data = PlayerInfo.get(CharacterData.class);
+        if ((data != null && data.getClassId() == -1) || CoreDBConfig.INSTANCE.lastClass == ClassType.NONE)
             McIf.player().sendChatMessage("/class");
 
         // This codeblock will only be executed if the Wynncraft AUTOJOIN setting is enabled
         // Reason: When you join a world with autojoin enabled, your current class is NONE,
         // while joining without autojoin will make your current class into the selected over the character selection.
-        CharacterData data = PlayerInfo.get(CharacterData.class);
-        if (data.getCurrentClass() == ClassType.NONE && CoreDBConfig.INSTANCE.lastClass != ClassType.NONE) {
+        if (data != null && data.getCurrentClass() == ClassType.NONE && CoreDBConfig.INSTANCE.lastClass != ClassType.NONE) {
             data.updatePlayerClass(CoreDBConfig.INSTANCE.lastClass, CoreDBConfig.INSTANCE.lastClassIsReskinned);
         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/wynntils/modules/core/events/ServerEvents.java
```diff
@@ -142,15 +142,20 @@ public void chatMessage(ClientChatReceivedEvent e) {
         }
         PartyManager.handleMessages(e.getMessage());  // party messages here
 
+        SocialData socialData = PlayerInfo.get(SocialData.class);
+        if(socialData == null) {
```
ryanzhoudev:
```suggestion
        if (socialData == null) {
```

src/main/java/com/wynntils/modules/core/events/ServerEvents.java
```diff
@@ -98,14 +98,14 @@ public void joinServer(FMLNetworkEvent.ClientConnectedToServerEvent e) {
      */
     @SubscribeEvent(priority = EventPriority.LOWEST)
     public void joinWorldEvent(WynnWorldEvent.Join e) {
-        if (PlayerInfo.get(CharacterData.class).getClassId() == -1 || CoreDBConfig.INSTANCE.lastClass == ClassType.NONE)
+        CharacterData data = PlayerInfo.get(CharacterData.class);
```
ryanzhoudev: I think this would benefit from being called `characterData` instead of `data`

src/main/java/com/wynntils/modules/core/events/ServerEvents.java
```diff
@@ -210,14 +215,19 @@ public void chatMessage(ClientChatReceivedEvent e) {
      */
     @SubscribeEvent
     public void addFriend(ClientChatEvent e) {
+        SocialData socialData = PlayerInfo.get(SocialData.class);
+        if(socialData == null) {
```
ryanzhoudev:
```suggestion
        if (socialData == null) {
```

src/main/java/com/wynntils/modules/core/events/ServerEvents.java
```diff
@@ -333,6 +343,10 @@ private static void startUpdateRegionName() {
             FrameworkManager.getEventBus().post(new SchedulerEvent.RegionUpdate());
 
             LocationData data = PlayerInfo.get(LocationData.class);
+            if(data == null) {
```
ryanzhoudev: Fix spacing, also maybe call this `locationData`

src/main/java/com/wynntils/modules/core/managers/PartyManager.java
```diff
@@ -20,15 +22,20 @@ public class PartyManager {
     private static final Pattern PROMOTE_PATTERN = Pattern.compile("Successfully promoted (.+) to party leader!");
 
     private static final CommandResponse listExecutor = new CommandResponse("/party list", (matcher, text) -> {
-        PlayerInfo.get(SocialData.class).resetPlayerParty();
+
+        SocialData socialData = PlayerInfo.get(SocialData.class);
+        if(socialData == null) {
```
ryanzhoudev:
```suggestion
        if (socialData == null) {
```

src/main/java/com/wynntils/modules/core/managers/PartyManager.java
```diff
@@ -47,36 +54,46 @@ public static void handlePartyList() {
...
+        String formattedText = McIf.getFormattedText(component);
+
         if (unformattedText.startsWith("You have successfully joined the party.")) {
             handlePartyList();
             return;
         }
-        PartyContainer partyContainer = PlayerInfo.get(SocialData.class).getPlayerParty();
+
+        SocialData socialData = PlayerInfo.get(SocialData.class);
+        if(socialData == null) {
```
ryanzhoudev:
```suggestion
        if (socialData == null) {
```

ryanzhoudev (commented):
A few things I noticed while skimming through this quickly. I will take another look in the morning or some other time

TheAvonian:
I fixed the spacing, I forgot to reformat before I committed at the time. Also changed out the generic data names into the more specific ones suggested.

kristofbolyai (approved):
This seems okay. Thanks.

kristofbolyai:
I am wondering though.. Was there a crash related to these being null? Where did the idea of making these specific methods null-safe come from?

TheAvonian:
The idea more so came from me working on another feature and realizing in a lot of files there is heavy use of `PlayerStats.get(T)` within the same method bodies so I went through those ones I was using and just made them try to call it once/twice at most in the body as every call to it is a foreach loop that is unneeded. 
Honestly the null checks are unneeded additions based on how everything is setup as it seems like they could never be null at those points but I added it for the sake of assurance. 

Ideally I would do this to every file but I only made these changes on my open files at the time.

kristofbolyai:
Yeah, this feels a bit paranoid. But then, the design of those "getters" are kinda bad imo.. So it should be fine to keep it.

ryanzhoudev (approved):

PR #621, opened 2022-10-28
fix: Devauth does not load JVM args

ryanzhoudev:
This line in build.gradle overrides any JVM args, such as `-Ddevauth.account=alt`. 
Also, I feel like since Minecraft accounts can be used across multiple projects, storing them in the Wynntils directory isn't very useful.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,8 +22,6 @@ minecraft {
         client {
             workingDirectory project.file('run')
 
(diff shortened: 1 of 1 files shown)


DevScyu:
You can add an override option to your local run config

DevScyu:
As devauth doesn't load unless a config is present, it makes sense for when people are cloning the project to have it already setup inside the project.

PR #627, opened 2022-10-29
feat: Add Lootrun Variables

Zffu:
**Info Variables Changes**
- Added the %slc% and %session_lootrun_chests% variables (The amount of opened chests in a lootrun in the playing session)
- Added the %lname% and %lootrun_name% variables (The name of the current lootrun)
- Added the %sl% and %session_lootruns% variables (The amount of lootruns done in the playing session)
- Added the %lc% and %lootrun_chests% variables (The amount of chests loaded in the current lootrun)
- Added the %lp% and %lootrun_points% variables (The amount of points loaded in the current lootrun)
- Added the %loc% and %lootrun_opened_chests% variables (The amount of opened chest during the current lootrun)

### DIFF
--- a/src/main/java/com/wynntils/modules/map/events/ClientEvents.java
+++ b/src/main/java/com/wynntils/modules/map/events/ClientEvents.java
@@ -123,7 +123,7 @@ public class ClientEvents implements Listener {
         if (LootRunManager.isRecording())
             LootRunManager.addChest(lastLocation); // add chest to the current lootrun recording
 
-        if(LootRunManager.isLootrunLoaded()){
+        if (LootRunManager.isLootrunLoaded()) {
             LootRunManager.addOpenedChestToSession();
             LootRunManager.addOpenedChestToLatestLootrun();
         }
--- a/src/main/java/com/wynntils/modules/map/events/ClientEvents.java
+++ b/src/main/java/com/wynntils/modules/map/events/ClientEvents.java
@@ -126,7 +126,6 @@ public class ClientEvents implements Listener {
         if(LootRunManager.isLootrunLoaded()){
             if(LootRunManager.isCheckALootrunChest(lastLocation)){
                 LootRunManager.addOpenedChestToSession();
-                LootRunManager.addOpenedChestToLatestLootrun();
             }
         }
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/wynntils/modules/map/events/ClientEvents.java
```diff
@@ -123,6 +123,11 @@ public void guiOpen(GuiOverlapEvent.ChestOverlap.InitGui e) {
         if (LootRunManager.isRecording())
             LootRunManager.addChest(lastLocation); // add chest to the current lootrun recording
 
+        if(LootRunManager.isLootrunLoaded()){
```
ryanzhoudev, line 126:
```suggestion
        if (LootRunManager.isLootrunLoaded()) {
```
Please be consistent with your formatting.

Zffu:
is there anything i need to modify ?

ryanzhoudev:
image

After unloading a lootrun, the %lname% variables does not reset. Additionally, before loading a lootrun, the %lname% variable displays as `%lname`. These should both be fixed; the lname variable should reset to `N/A` or `-` when nothing is loaded.

ryanzhoudev:
Additionally, the %lname% variable only displays the most recently loaded lootrun, even though it is possible to load multiple runs. Maybe something like `name1, name2`?

Also, %SLC% tracks all chests opened when a lootrun is loaded, not just chests in the lootrun. Is this intentional?

Zffu:
yes %slc% variable is the number of lootrun chests opened during the playing session

Zffu:
and for the %lname% variable i can make a list of every lootrun loaded

Zffu:
imma fix the %slc% chest bug

Zffu:
imma fix the bugs first and then add multi lootruns comptability

Zffu:
It will display that when no lootrun is loaded 
image

Zffu:
Display when there is multiple lootruns

image

Zffu:
I am adding a hashmap for every opened chest on specific lootruns so when you disable a lootrun it removes only the chest you have opened on this lootrun

Zffu:
i think im gonna remove the %loc% variable because its not very useful and it can be confusing with player location

ryanzhoudev (changes requested):
You are changing code that is not within the scope of this PR. (OverlayEvents, ConsumableTimerOverlay, DynamicTimerContainer, TabListData). Looks like you reverted a recent PR in your own branch. Please fix.

Zffu:
how to do that ?

Zffu:
im gonna reopen a pull request when i finished reverting

PR #630, opened 2022-11-02
fix: Skillpoints redirect regex not finding every message

Zffu:
The previous regex was not finding skill points messages such as:
- "You have 1 unused Skill Point and 1 unused Ability Point! Right-Click while holding your compass to use them"
- "You have 1 unused Skill Point!  Right-Click while holding your compass to use them"
- "You have 10 unused Skill Points! Right-Click while holding your compass to use them"

### DIFF
--- a/src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
+++ b/src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
@@ -335,7 +335,7 @@ public class OverlayEvents implements Listener {
         }
 
         if (OverlayConfig.GameUpdate.RedirectSystemMessages.INSTANCE.redirectOther) {
-            if (messageText.matches("You have \\d+ unused Skill Points and \\d+ unused Ability Points! Right-Click while holding your compass to use them")) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/wynntils/modules/utilities/overlays/OverlayEvents.java
```diff
@@ -335,7 +335,7 @@ else if (messageText.matches("[a-zA-Z0-9_ ]{1,19} has given you 20% resistance\\
         }
 
         if (OverlayConfig.GameUpdate.RedirectSystemMessages.INSTANCE.redirectOther) {
-            if (messageText.matches("You have \\d+ unused Skill Points and \\d+ unused Ability Points! Right-Click while holding your compass to use them")) {
+            if (messageText.matches("You have \\d+ unused Skill Point(s\\b|\\b)( and \\d+ unused Ability Point(s\\b|\\b))?! Right-Click while holding your compass to use them")) {
```
kristofbolyai: Why is a simple s? not enough?

kristofbolyai:
```suggestion
            if (messageText.matches("You have \\d+ unused Skill Points?( and \\d+ unused Ability Points?)?! Right-Click while holding your compass to use them")) {
```

PR #649, opened 2022-12-05
fix: Fix various ability bar bugs

### LINKED ISSUE #596: Custom Bars don't reset
### Version

1.12.1 (basically every other version as well)

### Brief Description

As the bossbar data (like focus, blood pool etc) are only updated when the bossbar is updated/sent, it does not reset to 0 when it should. Wynncraft fully removes your bossbar for Focus (and Awakened etc), but Wynntils doesn't recognize that it got removed, and the old data is kept

### Steps to Reproduce

Be on archer

Get the focus ability
Hit a shot with a bow

Miss a shot

### Expected Behaviour

Focus bar should be 0

### Actual Behaviour

Focus bar stays at 1

### Additional Comments

Im currently on a study trip. If anyone else wants to do this, feel free to. Otherwise, I would "fix" it once I'm back/have more time (at max in 3days)

### Installed Mods

_No response_

ryanzhoudev:
Fixes #596:
- Corrupted bar previously did not render properly
- Bars did not disappear when the vanilla Wynn one disappeared
- Bars did not differentiate between packet types

### DIFF
--- a/src/main/java/com/wynntils/core/framework/instances/data/BossBarData.java
+++ b/src/main/java/com/wynntils/core/framework/instances/data/BossBarData.java
@@ -142,7 +142,12 @@ public class BossBarData extends PlayerData {
         if (packet.getOperation() != SPacketUpdateBossInfo.Operation.UPDATE_PCT) return;
 
         int bloodPool = get(CharacterData.class).getCurrentBloodPool();
-        if (packet.getPercent() != 0) { // This is only sent initially, don't attempt to div by zero
+        // Only update the max blood pool if the blood pool is 1
+        // The max blood pool being 1 means the ADD packet was just received
+        // This max should only be set one time
+        // If done multiple times, the bar's maximum fluctuates between for example, 28-32, where it should be 30
+
+        if (packet.getPercent() != 0 && get(CharacterData.class).getMaxBloodPool() == 1) {
             int maxBloodPool = Integer.parseInt(maxBloodPoolFormat.format(bloodPool / packet.getPercent()));
             get(CharacterData.class).setMaxBloodPool(maxBloodPool);
         }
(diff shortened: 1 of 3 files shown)


ryanzhoudev:
Would be good if someone with sharpshooter could test in-game, I could only test for Blood Pool and Corrupted

src/main/java/com/wynntils/core/framework/instances/data/BossBarData.java
```diff
@@ -14,67 +15,152 @@ public class BossBarData extends PlayerData {
...
         maxBloodPoolFormat.setRoundingMode(RoundingMode.DOWN);
     }
 
-    public void updateBossbarStats(SPacketUpdateBossInfo packet) {
-        // (!) Do not remove .getName() check, Intellij is wrong about it
-        if (packet == null || packet.getName() == null) return;
+    public void processAddPacket(SPacketUpdateBossInfo packet) {
+        String title = packet.getName().getFormattedText();
+        // This is a new bar being added, set the UUID field
+        // .contains checks are okay since the title is not a user input
```
Incompleteusern: What if a guild contains one of the strings, unlikely as that is?

ryanzhoudev: Oh that is true

Incompleteusern (dismissed):
Looks good to me for the most part

Incompleteusern (approved):

src/main/java/com/wynntils/core/framework/instances/data/BossBarData.java
```diff
@@ -14,67 +15,171 @@ public class BossBarData extends PlayerData {
...
+     * Clears the given bossbar's stats back to -1
+     * Also removes any related visible bars from the user
+     * @param packet the packet of the bossbar to clear
+     */
+    public void processRemovePacket(SPacketUpdateBossInfo packet) {
+        UUID uuid = packet.getUniqueId();
+
+        if (uuid.equals(BLOOD_POOL_UUID)) {
+            BLOOD_POOL_UUID = null;
+            get(CharacterData.class).setBloodPool(-1);
```
kristofbolyai, line 111: In Artemis, I would say this will crash someone sometime, as we set an int which we essentially assume to be >=0 to be -1. But in legacy, this can work..

kristofbolyai (approved):
Looks okay

PR #651, opened 2022-12-16
feat: multi totem tracking

ryanzhoudev:
- Same thing as the previous draft but uses the new TimerContainer framework.

Also completely reworks the Cast SpellEvent. Previous implementation does not work anymore because it is possible to cast a spell with the same cost twice in a row, which skips posting the second SpellEvent. Actionbar version guarantees the post.
With this, I have also cleaned up a bunch of related files.

Also adds triple totem tracking instead of dual.


kristofbolyai (approved):

ryanzhoudev:
suffers from some bugs that we fixed in artemis that i do not want to backport, probably better off never being merged

PR #667, opened 2023-02-20
feat: Remove support for stable update stream

P0keDev:
Intended to prepare for the next (and final) stable push, since legacy receives minimal dev support at the moment.


Incompleteusern (approved):

ryanzhoudev (approved):

HighCrit:
Shouldn't this commit also remove/change either the `release` or `pre-release` action? Depending on if we always want to go remain being called beta

P0keDev:
> Shouldn't this commit also remove/change either the `release` or `pre-release` action? Depending on if we always want to go remain being called beta

We can change those as necessary down the road, but we need to push to stable at least once after merging this PR for it to work.

PR #675, opened 2023-04-16
fix: The mod can now load Artemis (1.19.3) lootruns

kristofbolyai:


src/main/java/com/wynntils/modules/map/managers/LootRunManager.java
```diff
@@ -362,13 +363,11 @@ private static class LootRunPathIntermediary {
         public List<Location> points;
         public List<BlockPos> chests;
         public List<LootRunNote> notes;
-        public Date date;
```
kristofbolyai, line 365: This could not be parsed easily. But we never used it, so I removed it instead of bothering with a fix.

magicus (approved):

ryanzhoudev (approved):

PR #676, opened 2023-05-06
fix: Race condition NPE for timer overlay

ryanzhoudev:


src/main/java/com/wynntils/modules/utilities/overlays/hud/ConsumableTimerOverlay.java
```diff
@@ -83,13 +83,15 @@ public static void addDynamicTimer(String name, int initialTimeSeconds, boolean
...
+        removeTimer(name);
 
         activeTimers.add(new StaticTimerContainer(prefix, name, displayedTime, persistent));
     }
 
     public static void removeTimer(String name) {
-        activeTimers.remove(findTimerContainer(name));
+        if (findTimerContainer(name) != null) {
+            activeTimers.remove(findTimerContainer(name));
+        }
```
kristofbolyai: Too bad this does not really fix the race, it just moves it. Extract `findTimerContainer(name)` to a variable, otherwise it can change until the next call (I am not sure how async this whole code is, but it really should be like that)

PR #686, opened 2023-08-12
fix: Fixed TAB_EFFECT_PATTERN regex

ENORMOUZ:
Fixed TAB_EFFECT_PATTERN regex so that Saviour's Sacrifice shows up in status effects list


magicus (approved):
We're not actively maintaining legacy, but I guess accepting trivial PR contributions like this is okay.

ryanzhoudev (approved):

PR #696, opened 2023-11-02
feat: Export favorites and waypoints from main menu

ShadowCat117:

### DIFF
--- a/src/main/java/com/wynntils/modules/core/config/CoreDBConfig.java
+++ b/src/main/java/com/wynntils/modules/core/config/CoreDBConfig.java
@@ -47,6 +47,9 @@ public class CoreDBConfig extends SettingsClass {
     @Setting(upload = false)
     public String lastVersion = "0.0.0";
 
+    @Setting(upload = false)
+    public boolean shownExportScreen = false;
+
     @Setting(displayName = "Main Menu Wynncraft Button", description = "Should a button be added to the main menu that allows you to connect to Wynncraft directly?")
     public boolean addMainMenuButton = true;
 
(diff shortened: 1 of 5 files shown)


magicus (dismissed):
Good thinking!

ShadowCat117:
> Good thinking!

I will say before this gets merged, not sure if you've seen the discussion on discord but some think this might be too intrusive as this screen always appears. I do agree but it is meant to be a constant reminder that 1.12 will be unsupported soon and it crashes a few seconds into logging in to the server anyway due to the territory changes

magicus:
We said that we would try to keep Legacy crash-free even if it means removing functionality. Can we chop off the territory things that are crashing?

(And yes, I think it can be intrusive)

ShadowCat117:
Tried fixing the territory crash but something else seems to still be casing a crash, sometimes before I can even load into the server so if we want to solve that might have to be done by someone who's more familiar with legacy..

Rafii2198:
I just realized two small issues that originated in my idea, so it is not anything with your implementation. 
* Saying "Artemis (1.20.2)" is not future-proof, if we update to for example 1.20.3 or 1.21 that statement will be false as well as sounds like Artemis is something else completely to Wynntils while it is just a code name. 
   * I suggest changing it to something like just `Wynntils Artemis` or `Wynntils for modern versions of Minecraft`
* Linking to whatever is the latest file on Modrinth could make the user think it is only for one specific loader, the one that happens to be uploaded the latest.
   * Simply changing the link to https://modrinth.com/mod/wynntils for the description or https://modrinth.com/mod/wynntils/versions to all versions should not cause any confusions

ryanzhoudev (approved):

kristofbolyai (approved):

